PR #4, opened 2013-07-05
Encapsulate BSSID, SSID, and location sanity checks in blocklists

cpeterso:
These classes do not implement lists, but conceptually they act like block lists.

If we want to get fancy, we could block BSSIDs that don't have valid manufacturer OUIs. Wireshark maintains a list of known OUIs: http://anonsvn.wireshark.org/wireshark/trunk/manuf . However, I don't think we should care as long as the BSSIDs are still unique (unlike `00:00:00:00:00:00`, which I have seen multiple times in my own stumbling).

The SSID blocklist checks for `_nomap` and a list of some common SSIDs for mobile devices (`AndroidTether` and `iPhone`), hotspots (`MiFi` and `MacBook`), and shuttles (`GBusWifi` and `ac_transit_wifi_bus`) we could block. If our server algorithms uses trilateration, then we probably want to block mobile APs. If our server algorithms uses the grid method instead of trilateration, we probably _want to unblock_ mobile APs and ad-hoc BSSIDs because duplicate measurements of the same device in multiple locations doesn't hurt. Unfortunately, our server can't make these distinctions if it only knows about SHA1 fingerprints instead of actual SSIDs. I think our early stumbling filters should be conservative because it's easier to remove these restrictions later than to clean up a polluted database later.

The values I chose for the location sanity checks are pretty arbitrary. I figured we could tighten the limits later. I have seen some Android devices, like the Galaxy Nexus, that report valid latitude and longitude with bogus negative altitudes. So we might want to be relax altitude sanity checks. The iPhone GPS seems to clamp negative altitudes at 0.

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -15,7 +15,7 @@
     <uses-permission android:name="android.permission.INTERNET"/>
     <uses-feature android:name="android.hardware.location"/>
     <uses-feature android:name="android.hardware.location.gps"/>
-    
+
     <application
         android:allowBackup="true"
         android:icon="@drawable/ic_launcher"
--- a/src/com/mozilla/mozstumbler/BSSIDBlockList.java
+++ b/src/com/mozilla/mozstumbler/BSSIDBlockList.java
@@ -1,0 +1,40 @@
+package com.mozilla.mozstumbler;
+
+import android.net.wifi.ScanResult;
+
+import java.util.Arrays;
+import java.util.HashSet;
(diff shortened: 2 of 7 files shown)


cpeterso:
After talking with @hannosch on IRC, I think the SSID blocklist is a bad idea. I will submit a revised PR later.

The ad-hoc network filter should catch most of the same networks and the list of SSIDs would miss many networks, so the server would need to detect moving networks anyways.

PR #8, opened 2013-07-09
Dougt changes

dougt:


dougt:
> What if the device is indoors and the GPS provider is unavailable?

I am not sure we'd want to do that.

cpeterso:
@dougt, why don't you want to fall back to the cell network location provider if GPS is unavailable indoors? The stumbler will see APs, but won't report them without a location.

Google's new FusedLocationProvider uses the network location provider if its reading is at least 11 seconds more recent than the GPS reading: https://github.com/android/platform_frameworks_base/blob/0fdfa6b49e7642a90a1b98881e4e2801cbcd3c77/packages/FusedLocation/src/com/android/location/fused/FusionEngine.java#L196

dougt:
I want to attempt to only collect gps derived data -- not wifi derived data.

PR #25, opened 2013-07-20
Generate random user token for database submission stats

### LINKED ISSUE #16: Generate random user token for database submission stats
Client generates random user token to be included in submissions. There is no opt-out yet; the user implicitly opts-in by running the stumbler.

cpeterso:
This PR fixes #16.

If a user token (UUID) does not exist, a random one is generated. The `Prefs` class includes a `deleteToken()` method so we can add UI to allow users to delete their token to opt-out.

I moved the report's JSON encoding to the background thread to avoid StrictMode warnings because `getToken()` caused SharedPreferences file I/O on the main thread.

### DIFF
--- a/src/org/mozilla/mozstumbler/MainActivity.java
+++ b/src/org/mozilla/mozstumbler/MainActivity.java
@@ -18,139 +18,137 @@ import android.widget.Button;
 
 public class MainActivity extends Activity {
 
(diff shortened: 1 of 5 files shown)


cpeterso:
I rebased this PR onto the new ScannerService. I also fixed the tabs/spaces whitespace again.

PR #26, opened 2013-07-22
Make service

dougt:
Chris, this is a pretty big change.  It attempts to do the following:

1) factors out stuff from the main activity into a new Service.
2) implements a basic notification item
2a) this seems to be required and what we're suppose to do if we have a background service running for any long period of time
3) creates a alarm that wakes us up if the service somehow dies (hard to test -- needs more testing)
4) creates a runnable looper thread in the service so that we can do things that normally require a looper (like geo/cell/wifi APIs.)
5) I've moved the service into it's own process.  So, I needed to add a bit of IPC via aidl.  (Totally amazing eclipse makes this.  Makes me want to cry when I think at mozilla/chrome style ipdl.)

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -25,12 +25,20 @@
         android:theme="@style/AppTheme" >
         <activity
             android:name="org.mozilla.mozstumbler.MainActivity"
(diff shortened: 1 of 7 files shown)


dougt:
Both the activity and the service must be exported.

PR #30, opened 2013-07-23
Add broadcast receiver

dougt:
This adds a way for the service code to send messages to the activity.  This allows us to inform the UI if there is anything interesting happening.  For example, the service now can do something like:

```
                    Intent i = new Intent(ScannerService.MESSAGE_TOPIC);
                    i.putExtra(Intent.EXTRA_SUBJECT, "Notification");
                    i.putExtra(Intent.EXTRA_TEXT, "Yippie!");
        sendBroadcast(i);
```

And the Activity, if running, will display a toast with the text "Yippie!"

I think we can use this to inform the UI how many AC's have been uploaded for gaming.


cpeterso:
OK

PR #35, opened 2013-07-25
Add Gradle command-line build system and ProGuard optimizations.

cpeterso:
I got fed up with Eclipse, so I investigate command-line build systems for Java and Android. Gradle is Android's new official build system. People can still use Eclipse if they like.

http://www.gradleware.com/resources/tech/android

http://tools.android.com/tech-docs/new-build-system/user-guide

I also enabled ProGuard optimizations for release builds. ProGuard shrinks our MozStumbler.apk from 258 KB to 70 KB. I didn't enable ProGuard obfuscation because we don't care about hiding our open source code and obfuscation makes exception stack traces more difficult to debug.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,14 @@ jdk: oraclejdk7
 before_script:
     - sudo apt-get update -qq
     - if [ `uname -m` = x86_64 ]; then sudo apt-get install -qq --force-yes libgd2-xpm ia32-libs ia32-libs-multiarch; fi
+
+    # Install Gradle
+    - wget http://services.gradle.org/distributions/gradle-1.6-bin.zip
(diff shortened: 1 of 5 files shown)


dougt:
This is good.  Could you fix up the travis c-i?

https://travis-ci.org/dougt/MozStumbler/builds/9461475

cpeterso:
The travis log looks like it's skipping all the .travis.yml before_script steps before running `gradle assemble`. I don't know why.

The travis build is failing because "SDK location not found. Define location with sdk.dir in the local.properties file or with an ANDROID_HOME environment variable." But ANDROID_HOME should be getting set in .travis.yml's before_script section.

Issue #18, opened 2013-07-18
Allow user to enter a nickname


cpeterso:
Client should allow a user to enter a nickname. Client should only include the user token in submissions iff nickname is defined. This makes the leaderboard tracking opt-in because the user must manually enter a nickname to be listed on the leaderboard.

hannosch:
Corresponding server-side issue is: https://github.com/mozilla/ichnaea/issues/9

hannosch:
The nickname should be at least 3 characters long and at most 128 - at least that's the validation I currently put on the server side. This struck me as a reasonable length range.

PR #36, opened 2013-07-26
Add X-Token and X-Nickname request headers

cpeterso:
@hannosch recommends (in mozilla/ichnaea#9) replacing our `token` JSON property with `X-Token` and `X-Nickname` request headers.
1. This PR adds an `X-Token` header without removing the `token` JSON property. We can remove the JSON property after the server understands `X-Token` headers. This should allow users to upgrade to new client versions without losing their token (and score). The server may want to continue looking for both to ensure for a short period to avoid losing tokens for laggards who have not upgraded their clients.
2. This PR adds an `X-Nickname` header but does not add any UI to read or write the nickname! That is issue #18. As a temporary hack, I generate a semi-unique but personally recognizable nickname by concatenating the user's device model and Android OS version. For example, my generated nickname is "Nexus 4 (Android 4.2.2)".

### DIFF
--- a/src/org/mozilla/mozstumbler/Prefs.java
+++ b/src/org/mozilla/mozstumbler/Prefs.java
@@ -2,6 +2,8 @@ package org.mozilla.mozstumbler;
 
 import android.content.Context;
 import android.content.SharedPreferences;
+import android.os.Build;
+import android.os.Build.VERSION;
 import android.util.Log;
 
 import java.util.UUID;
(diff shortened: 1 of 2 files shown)


dougt:
I am happy to make this change, but I'd wonder why having it as a http header is better in the json?

cpeterso:
@hannosch suggested we move token and nickname from JSON to HTTP headers. The token and nickname are not really part of the location protocol; we just use them for gamifying the stumbler. Also, the headers are a more usual mechanism for user identification or sessions.

hannosch:
The header support is done and live on the server.

Issue #33, opened 2013-07-24
Scanning toast messages are blank


cpeterso:
Toast messages are blank and I see the following exception stack trace in logcat because `R.string.start_scanning` and `stop_scanning` are integers, not strings.

```
src/org/mozilla/mozstumbler/ScannerService.java
133:                        i.putExtra(Intent.EXTRA_TEXT, R.string.start_scanning);
158:                        i.putExtra(Intent.EXTRA_TEXT, R.string.stop_scanning);
```

```
W/Bundle  (31242): Key android.intent.extra.TEXT expected String but value was a java.lang.Integer.  The default value <null> was returned.
W/Bundle  (31242): Attempt to cast generated internal exception:
W/Bundle  (31242): java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.String
W/Bundle  (31242):  at android.os.Bundle.getString(Bundle.java:1069)
W/Bundle  (31242):  at android.content.Intent.getStringExtra(Intent.java:4350)
W/Bundle  (31242):  at org.mozilla.mozstumbler.c.onReceive(Unknown Source)
W/Bundle  (31242):  at android.app.LoadedApk$ReceiverDispatcher$Args.run(LoadedApk.java:758)
W/Bundle  (31242):  at android.os.Handler.handleCallback(Handler.java:725)
W/Bundle  (31242):  at android.os.Handler.dispatchMessage(Handler.java:92)
W/Bundle  (31242):  at android.os.Looper.loop(Looper.java:137)
W/Bundle  (31242):  at android.app.ActivityThread.main(ActivityThread.java:5227)
W/Bundle  (31242):  at java.lang.reflect.Method.invokeNative(Native Method)
W/Bundle  (31242):  at java.lang.reflect.Method.invoke(Method.java:511)
W/Bundle  (31242):  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:795)
W/Bundle  (31242):  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:562)
W/Bundle  (31242):  at dalvik.system.NativeStart.main(Native Method)
D/org.mozilla.mozstumbler.MainActivity(31242): Received a notification intent and showing: null
```

cpeterso:
Fixed by #37.

PR #37, opened 2013-07-26
Properly load string to be passed to the toast.

eggpi:
This fixes issue #33.


cpeterso:
Thanks! That bug was really annoying. :)

PR #42, opened 2013-07-27
Add a couple of permissions I think I forgot initially.  Also, add some ...

dougt:
...really bad UI so that we can see how many uploads we've done.  (Turns out we need to do a bit of local filtering... i think)


cpeterso:
I think the above issues should be fixed, but I'll merge this because I want the UI. :)

PR #55, opened 2013-08-03
Add nickname editor

### LINKED ISSUE #18: Allow user to enter a nickname

cpeterso:
This PR fixes #18.
1. Add nickname EditText
2. Make "Start Scanning" button _HUGE_ (Pixels are cheap, so why not?)
3. Consolidate some UI strings
4. Remove "View Scan Map" button (because most users will only care about the leaderboard and the world map is not very usable in a mobile browser)

NB: The nickname editor adds some StrictMode violations, but someone can fix them later. <:)

### DIFF
--- a/res/layout/activity_main.xml
+++ b/res/layout/activity_main.xml
@@ -17,6 +17,15 @@
            android:text="@string/gps_fixes"
            android:textAppearance="?android:attr/textAppearanceMedium" />
 
+       <TextView
+           android:id="@+id/reportedTextView"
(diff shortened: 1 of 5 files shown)


cpeterso:
The server limits nicknames to at least 3 characters long and at most 128. The client's nickname editor should probably enforce those limits.

PR #57, opened 2013-08-05
Display Froyo-compatible notification when scanning starts

### LINKED ISSUE #27: ScannerService breaks Froyo and Gingerbread compatibility by using Notification.Builder
From https://github.com/dougt/MozStumbler/pull/26#commitcomment-3689527

Notification.Builder was added in Android API Level 11 (Honeycomb). We should maintain Gingerbread support because it's still 34% of devices. I don't care too much about Froyo because it's only 3%, but it's a nice-to-have if it's not much work.

We will need to configure Notifications manually like Fennec does here: https://hg.mozilla.org/mozilla-central/file/bda9723bdccc/mobile/android/base/UpdateService.java#l367

cpeterso:
This PR fixes #27.
1. The `Notification.Builder` code is only supported in API Level 11+. This PR creates the notification using a (deprecated) `Notification` constructor instead of `Notification.Builder`.
2. This PR also displays the notification icon when scanning starts, not when `MainActivity.onStop()` is called. Displaying the notification immediately visually ties the notification's lifetime with the scanning period.
3. The new notification makes the toasts redundant, so I removed the calls (but left the toast code so we can use it for other service alerts).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,5 +44,5 @@ dependencies {
 }
 
 tasks.withType(JavaCompile) {
-    options.compilerArgs = ['-Xlint:all,-deprecation', '-Werror']
+    options.compilerArgs = ['-Xlint:all', '-Werror']
 }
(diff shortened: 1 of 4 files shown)


dougt:
> This PR also displays the notification icon when scanning starts, not when MainActivity.onStop() is called.

i guess, i don't care that much, but I thought that putting a notification up while the application was already in front of the user was kind of jarring.  We'll give this a go for a while. ;)

PR #59, opened 2013-08-05
Batch report support

dougt:


dougt:
> It would be nice to check for null storedReports rather than handling the NPE. getReports() will return null in non-error cases, which is much different than getReports() returning a bogus JSON string.

You need the try in any case:

/Users/dougt/builds/MozStumbler/src/org/mozilla/mozstumbler/Reporter.java:44: unreported exception org.json.JSONException; must be caught or declared to be thrown

dougt:
> You never call sendReports(true) to force sending. Maybe shutdown() could be renamed flushReports() and call sendReports(true)? If the user never runs the stumbler again, we could be missing up to 99 unreported measurements!

We can do that...

My worry is was about the lifetime of a service.  if we do a network request during onDestroy of the Service, that might slow other things down, right?  Also.  I am not sure if onDestroy will allow us to complete the network request.  If you think we can improve here, file follow up?  I don't really know what we can do more than this.

> If you're going to remove the code to sending the toast intents from the ScannerService, you should probably remove the toast handling code from MainActivity, too.

Someone might use it! :)

cpeterso:
LGTM

Issue #41, opened 2013-07-27
Pause scanning on ACTION_BATTERY_LOW intent


cpeterso:
We should politely stop scanning when we receive an `ACTION_BATTERY_LOW` intent or `ComponentCallbacks.onLowMemory()` callback. We will definitely need to implement this if we add the always-on background mode described in issue #21.

Should we resume scanning if we later receive an `ACTION_BATTERY_OKAY` intent? Or wait for the user to manually resume scanning?

dougt:
pause IFF you're on battery.  I think it's okay to ignore this if you're powered.

eggpi:
Tried to get started on this today at guilherme-pg/MozStumbler@540d312eac8de475bd4d8da99a4e8daaf13c2ff4 , but for some reason I couldn't get the app to receive the broadcasts. I have only tried it on the simulator so far, by using the power command (https://developer.android.com/tools/devices/emulator.html#power).

I tried both methods of receiving broadcasts (receiveBroadcast and AndroidManifest.xml) to no avail. In that specific commit I'm sticking closely to the manifest method described in the documentation (https://developer.android.com/training/monitoring-device-state/battery-monitoring.html#MonitorLevel), but still no luck.

cpeterso:
@guilherme-pg: you can't register for the ACTION_BATTERY_LOW intent in AndroidManifest.xml. You must call registerReceiver(). If your battery is low, starting a bunch of apps that are not already running to tell them is probably a bad idea. :)

I don't know if you can register for ACTION_BATTERY_OKAY in AndroidManifest.xml or not. For more details, see http://stackoverflow.com/a/12348778/61352

cpeterso:
btw, I disagree with @dougt's recommendation that we ignore ACTION_BATTERY_LOW when we're on AC power. If your battery is low, then your battery is low. If your device is running a heavyweight app (such as one scanning Wi-Fi networks and GPS), it can drain the battery faster than AC power can recharge the battery. I see logcat warnings about this from Fennec.

eggpi:
Ok, I got this to work somewhat.

It seems the BATTERY_OKAY and BATTERY_LOW broadcasts are only triggered when not on AC power -- I had to connect to the emulator and do `power ac off` before I could get my receivers to be called, even after registering at runtime with registerReceiver().

[...]

PR #67, opened 2013-08-14
Stop scanning when we're on low battery.

eggpi:
This addresses issue #41.

This leaves some room for refactoring: we no longer have the main activity as a the service's sole controller that asks it to start and stop; we now also have the service stopping itself sometimes and we need that information to flow back to the UI. My solution here was straightforward (issue a "Scanner" intent to force an UI update), but we might want to re-think the information flow between the main activity and the server in the future.


cpeterso:
Looks good! just a couple nits to fix. :)

eggpi:
Just updated this, the nits should be fixed now.

cpeterso:
LGTM. The Travis build failed because github was down.

Issue #86, opened 2013-08-25
Avoid sending "empty" locations


hannosch:
I've seen a couple dozen validation errors on the server, with this message:

_error_handler[{'location': 'body', 'name': 'body', 'description': 'You need to provide a mapping with least one cell or wifi entry.'}]_

This indicates a missing client side check. We should make sure we only include batch items, which have either cell or wifi entries or both. Probably some simple length check in the right place.

PR #87, opened 2013-08-28
Dodge possible race condition when uploading reports

cpeterso:
In issue #86, @hannosch says some clients are uploading empty reports. I suspect they are older pre-batch-upload clients because the current code checks for `mReports.length() > 0`, but there is a (theoretical) race condition where a Reporter background thread could reset `mReports = new JSONArray()` after a second Reporter background thread is spawned but before the second thread uploads its (now empty) `mReports`.

### DIFF
--- a/src/org/mozilla/mozstumbler/Reporter.java
+++ b/src/org/mozilla/mozstumbler/Reporter.java
@@ -2,6 +2,9 @@ package org.mozilla.mozstumbler;
 
 import android.content.Context;
 import android.content.Intent;
+import android.content.pm.PackageInfo;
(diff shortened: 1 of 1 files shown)


hannosch:
This isn't really addressing #86. You are preventing totally empty batches to be sent to the server. The error message indicates that there at least one batch item, which has the lat/lon/time data, but no cell or wifi record to go along with it. Think of it as an incomplete batch item. So this problem is likely to occur in the data gathering part.

cpeterso:
I updated this PR to include a `User-Agent: MozStumbler/X.Y.Z` HTTP header to fix #88.

PR #89, opened 2013-08-28
Don't upload report if we have no WiFi scan results or cell info

### LINKED ISSUE #86: Avoid sending "empty" locations

cpeterso:
This PR should fix #86.

Also cache radio type. I am assuming the radio type does not change at run time. :)

### DIFF
--- a/src/org/mozilla/mozstumbler/Scanner.java
+++ b/src/org/mozilla/mozstumbler/Scanner.java
@@ -221,15 +221,18 @@ class Scanner implements LocationListener {
         } else {
             Log.d(LOGTAG, "New location: " + location);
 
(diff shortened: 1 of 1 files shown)


hannosch:
Looks good to me.

dougt:
not sure why CI failed.  Change looks good, merging.

PR #90, opened 2013-08-28
Reduce wifi polling time to 1s

dougt:


cpeterso:
LGTM, but we are throwing away all these scan results if the device is not moving. If we get different scan results when we're not moving, we should consider reporting the new scan results, too.

PR #113, opened 2013-09-15
Resurrect SSID block list

cpeterso:
@hannosch had convinced me that client-side filtering of ad hoc networks and mobile device SSIDs was an unnecessary optimization because the server must detect moving APs anyways. But I've changed my mind on the mobile device SSIDs because the server-side detection will not be perfect and the client can easily detect (some) mobile devices.

I still agree that we don't need to filter ad hoc networks because there are quite a few stationary ad hoc networks (and the mobile ones should get filtered by the SSID block list or server-side moving AP detection).


cpeterso:
Maybe later..

cpeterso:
I resurrected the SSID block list, but made the name filters much more focused to reduce false positives.

dougt:
do we want/need this still?  It isn't clear.

cpeterso:
As I wrote an argument in favor of client-side SSID filtering, I convinced myself it really is not desirable. <:)

The server will need to detect moving APs, though its algorithm won't be perfect. Client-side filtering can help by detecting likely moving APs (by mobile SSIDs). However, if a "possibly mobile" AP (like a laptop) that doesn't actually move would still be useful. Also, if we replace our simple trilateration location algorithm with the grid-based "fingerprinting" approach, APs reported in multiple locations are not a problem (and are still somewhat useful).

PR #119, opened 2013-10-03
Consolidate NoticeDialog's negative and cancel code paths

cpeterso:
If the user tapped the grayed-out app view behind the NoticeDialog, the dialog would be dismissed but neither the positive ("OK") or negative ("Cancel") button code paths would be called.

Also move the Prefs.hasSeenNotice() check outside of NoticeDialog to avoid unnecessary allocation of new NoticeDialog.


cpeterso:
I need this PR for the updater.

Issue #112, opened 2013-09-12
nickname not saved unless you hit "Done"


dholbert:
tl;dr: You have to hit "Done" to actually save the nickname edits, which is non-obvious, since you can hide the keyboard and un-focus the field, and the edits look like they persist, until you quit & re-launch the app.

STR:
1. Open app
2. Tap nickname field & modify the content. (do _NOT_ hit "Done")
3. Hit the "Hide Keyboard" downarrow button (at bottom-left of keyboard on my Nexus4)
4. Optional: Start scanning, stop scanning, whatever.
5. quit app by triggering app switcher & swipe app offscreen
6. Re-launch app

EXPECTED RESULTS: My nickname should look the way it was when I quit the app.
ACTUAL RESULTS: My nickname edits weren't saved.

flaki:
As I see things now there is no `Done` button any more, so this issue seems obsolete to me.

cpeterso:
The `Done` button is on the virtual keyboard and the key layouts might be different for different keyboards and locales.

flaki:
Yeah, actually (I think) I have experienced the issue myself later on today. Will try to investigate this, too.

cpeterso:
Fixed by #135.

PR #135, opened 2013-10-22
Use text change events to save nickname.

eggpi:
This is a possible workaround for issue #112.


cpeterso:
LGTM. Thanks, @guilherme-pg!

PR #137, opened 2013-10-24
Rename "About Privacy" menu item to "About MozStumbler"

cpeterso:
The FAQ is about more than just privacy. So saith Jishnu.

Also remove the first-run legal notice dialog because the FAQ supersedes it.


cpeterso:
@dougt: r?

cpeterso:
Yes, we probably should move this FAQ to location.services.mozilla.com. I filed ichnaea issue 37 to track the site change.

Issue #129, opened 2013-10-10
Map crashes when Google Play Services is not installed


eggpi:
I'm running CyanogenMod 10.1.3 without Google's apps. Clicking the Locate Me button seems to cause a crash ("Unfortunately MozStumbler has stopped") on top of an error message telling my I don't have Google Play Services.

The stumbler does seem to start scanning, though, so it may be something crashing in the background. Logcat tells me that the exception on MapActivity.java:193 is caught.

cpeterso:
oops! Thanks for the bug report.

@guilherme-pg: I believe the log message about MapActivity.java:193 catching an exception is about a _different_ exception. Did you see any other exception stack traces in logcat?

eggpi:
Yes, sorry, there was another one right below it. For some reason the line numbers are not really interesting; they point to the declarations of the classes containing the methods in the stack rather than to the lines that are executing.

However, it's a NullPointerException that has MapActivity$GetGeolocationAndMapItTask.onPostExecute and then MapActivity.access$500 at the top of the stack, so it seems reasonable to assume setPositionAndMarker is crashing due to mMap being null.

I can test that hypothesis later today if it's not easy for you to do it :)

eggpi:
Hmm, for some reason this didn't really fix the problem for me. This time I got a RuntimeException when trying to call setContentView at MapActivity.onCreate. Catching that exception does prevent the crash for me, I'll have a patch later.

I also tried reverting your null pointer handling and applying my change, and that once again prevented the crash. So it looks like the actual fix does not involve the null pointer checks, sorry if I misled you.

merike:
I get the following with 0.7.0 when pressing Locate Me:
W/dalvikvm( 1965): VFY: unable to resolve virtual method 1625: Lorg/mozilla/mozstumbler/MapActivity;.getFragmentManager ()Landroid/app/FragmentManager;
D/dalvikvm( 1965): VFY: replacing opcode 0x6e at 0x0013
D/dalvikvm( 1965): VFY: dead code 0x0016-0057 in Lorg/mozilla/mozstumbler/MapActivity;.onCreate (Landroid/os/Bundle;)V
D/AndroidRuntime( 1965): Shutting down VM
W/dalvikvm( 1965): threadid=1: thread exiting with uncaught exception (group=0x40028a00)
E/AndroidRuntime( 1965): FATAL EXCEPTION: main
[...]

PR #139, opened 2013-10-25
Catch RuntimeException when there's no map

eggpi:
This fixes issue #129 for me. Of course, feel free to not revert the null check for the map if you still think it's necessary; it just didn't seem to help in my particular case.


cpeterso:
@guilherme-pg: What does the `RuntimeException` say? Can you share the stack trace? Is the `RuntimeException` thrown before `getMap()` is even called?

PR #152, opened 2013-10-29
Added 'reports sent' field and text view.

denevell:
This is so the user can see how many reports they've sent to the server.

I was getting frustrated using the app today without this information.

Cheers.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,4 @@
+
 buildscript {
     repositories {
         mavenCentral()
@@ -8,6 +9,7 @@ buildscript {
 }
 
 apply plugin: 'android'
+apply plugin: 'eclipse'
 
 dependencies {
   compile 'com.google.android.gms:play-services:3.2.25'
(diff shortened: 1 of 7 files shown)


denevell:
Okay, I've changed the formatting.

I think keeping the lat uploaded time may be useful for some users, so they can see how recently their data was uploaded. In other words, "I've been walking down this road for x minutes, and the last time someone was x time?". I'd like the information, anyhow - so I can roughly estimate when the leaderboard would have been updated.

Thoughts?

cpeterso:
@denevell: I see what you mean. Displaying both last upload time and the number of reports sent makes sense.

cpeterso:
Thanks!

denevell:
More than welcome.

I intent to offer some more commits in the next couple of days.

1) Nickname in a settings / preferences screen
2) Using actionbarcompat so the play etc can be placed there, leaving more space to play with on the main page.
3) Saving all the reported locations as serialised preferences so the user's reports can be plotted on a map
4) Using some custom views to modularise the layout code, single responsibility principle, etc.
5) Work on the SSLException Connection timed out error you get on uploading sometimes, although I think the  branch on networking may fix that.

cpeterso:
@denevell: Sounds great. Logging the measurements to a proper SQLite database (so we can add an export feature like other Android stumblers) and plotting the user's reports on the map are popular feature requests.

PR #154, opened 2013-10-30
Fix service preemption by putting it into foreground.

### LINKED ISSUE #153: MozStumbler preemption.
Sometimes MozStumbler resets all stuff that is displays. 
This usually happens when I launch something heavy besides MozStumbler.

For example, im walking on street and scanning is turned on. Then I scan a lot and I want to see where Im on leader board, so I start browser (which is heavy app), see some stuff and go back to MozStumbler and see that all counters are zero and big button prompts me to "start scanning", but MozStumbler icon in notification area is still there!
Also, I noticed that just before such "preemption" GPS icon dissappears.

Galaxy Nexus, 
version 4.3
0.7.1

UPDATE:
Im not sure, and dont have much time to investigate, but brief look at logcat and dumpsys output says that your service is not foreground and freely can be killed, so that happens after host activity dies.

pepyakin:
This commit fixes #153.

You can test yourself: just start scanning and drop task (from tasks view) which holds MozStumbler.

### DIFF
--- a/src/org/mozilla/mozstumbler/ScannerService.java
+++ b/src/org/mozilla/mozstumbler/ScannerService.java
@@ -23,7 +23,7 @@ public final class ScannerService extends Service {
     public static final String  MESSAGE_TOPIC   = "org.mozilla.mozstumbler.serviceMessage";
 
     private static final String LOGTAG          = ScannerService.class.getName();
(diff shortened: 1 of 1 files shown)


cpeterso:
Thanks, @pepyakin!

PR #157, opened 2013-10-30
Added preferences screen for nickname.

denevell:
Now the user goes into the settings page to edit their nickname, thereby saving space on the main page and reducing lines of code in the main activity.

Moved preference classes into their own package, something I'd suggest will be needed as the project grows in classes. And made methods on Prefs.java public to allow such.

### DIFF
--- a/AndroidManifest.xml.in
+++ b/AndroidManifest.xml.in
@@ -45,6 +45,10 @@
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
         </activity>
+        <activity
+            android:name="org.mozilla.mozstumbler.preferences.PreferencesScreen"
+            android:label="@string/app_name" >
+        </activity>
 
         <activity
             android:name="org.mozilla.mozstumbler.MapActivity"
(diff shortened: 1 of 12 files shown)


ghost:
Looks great, thx.

dougt:
Thanks for the pull request!!

hannosch:
This change means the nickname is now stored in a different place, right? At least when I upgraded it lost my nickname and had to enter it again.

That's unfortunate. We should at least warn people about this in some way.

denevell:
Same place. I haven't changed any code in the Prefs.java code to change either the location or name of the preference.

Do you know if this happened on updating from previous versions too? I.e. 0.7.0 to 0.7.1, i.e. the text field being empty on update?

I'm trying to work out if it's a problem with the new commits or an outstanding problem.

hannosch:
@denevell This only happened on the 0.7.1 -> 0.7.2 upgrade. But it can also be something silly like the text field not showing the entry until you enter it again (while the nickname is still stored).

denevell:
I've debugged it and you are correct. I did doubt it because all the documentation says this should not be so. I can have another look around, but I fear by the time I get around to fixing it (little time this weekend) most will have updated. The biggest fear I have is that this will happen on the next update.

Some kind of notification would be a better idea, I feel. Again, if I get time I'll look at ways to do this, via a toast or dialog box, etc. Any help would be more than appreciated. Thanks for reporting the bug.

PR #160, opened 2013-10-31
Add support for an API key

dougt:


ghost:
API key is required now or in near future?

dougt:
@freekzy probably not immediately.

PR #162, opened 2013-10-31
Refactor MozAPIKey code

cpeterso:
And fix some nits from earlier PRs.


hannosch:
If you are into cleanup: Can you make sure we sent to API key for both the submit and the search API's? If we want to use it for rate limiting / blocking spammers, we should do so on both.

Issue #172, opened 2013-11-01
Add runtime warning when GoogleAPIKey is unset that map view will be blank


cpeterso:
As per my idea in issue #167, we should add a runtime (and build time?) warning when `GoogleAPIKey` is unset or is `FAKE_GOOGLE_API_KEY` so people will know why the map view is blank. logcat and toast messages would be good.

ghost:
Maybe just disable the button if key is fake?

cpeterso:
Disabling the button is a good idea, though some users will probably wonder why the button is disabled. :)

In another issue, we might consider disabling the button if the Google Play Services library is not installed. Offhand, I'm not sure how we would check that.

ghost:
@cpeterso, ok, all done.

cpeterso:
Fixed by PR #174.

PR #174, opened 2013-11-01
Disable map button if fake google api key found

ghost:
Issue #172


dougt:
Thanks!! https://github.com/dougt/MozStumbler/commit/db7b956464ccc69de00f3b200d281d85ae5fdd01

PR #182, opened 2013-11-03
Add CDMA logging to MozStumbler

lordsutch:
This patch adds logging of CDMA base stations to MozStumbler, per https://github.com/dougt/MozStumbler/issues/71. I took a slightly different tack than the pseudocode for two reasons:
- CDMA phones (even with LTE) continue to track CDMA base stations for voice coverage, so there's no need to watch the current network type.
- EvDO signals may not come from the same site as the 1xRTT (CDMA) signal; since Android doesn't provide a standard API for getting EvDO base station IDs, there's no reason to try to geolocate using EvDO signal strength.

Also, the pseudocode wasn't assigning the "mnc" entry per the documentation at https://mozilla-ichnaea.readthedocs.org/en/latest/cell.html#cdma, so I fixed that to log the CDMA SID as intended.


dougt:
@hannosch, can you review?

cpeterso:
@lordsutch: You might want to rebase your branch on top of the latest master. It looks like your branch might have a merge commit duplicating some code that was already merged to master..?

hannosch:
@lordsutch - this looks awesome and thx for catching my mistakes.

r+ for the intent of the code, someone else has to do a Java/Android review ;-)

PR #187, opened 2013-11-04
Refactor cell scanner

illarionov:
Abstract CellScanner class so to be able to create different implementations of the cell scanner.
An example of such implementation may be support for multimple SIM handsets (dougt/MozStumbler#178) or it can be used for dougt/MozStumbler#72


cpeterso:
@illarionov: Thanks for taking the time to break your PR into these smaller commits. It was definitely made my code review much easier! :+1:

I made many review comments, but please do not be discouraged! I didn't see any major problems with your new design. I think we just need a few revisions before merging.

illarionov:
Thank you for a very detailed review of my code :)
I have comited some of the proposed fixes.
Actually the main goal for me was to show the changes that I needed to 
implement support for multiple SIM. This request is not necessary to merge.

cpeterso:
Looks great! :+1: 

@illarionov: When you say "This request is not necessary to merge", do you mean that I should _not_ merge it and instead I should wait for your big PR for multiple SIMs? If this PR works on its own (preserving current cell scanner behavior), I think we should merge it first so we can incrementally test that this code does not add any surprising bugs before merge your multiple SIM support. :)

Does that make sense? If so, can you please use `git rebase -i master` to squash all your commits (code and cleanup) into one commit? If this PR won't work without multiple SIMs support, then please close this PR with a comment before submitting your PR for multiple SIM support. Thanks!

illarionov:
Squashed and ready for merging

cpeterso:
Thanks, @illarionov! :cake:

Issue #189, opened 2013-11-05
Don't force exit if GPS isn't turned on


cskau:
With 216672f it is no longer possible to open the app without turning on GPS.
This is unfortunate because it prevents the use case of opening the app, and for instance View'ing the Leaderboard, or Test'ing Mozilla Location Service; neither of which _require_ the use of GPS, as far as I can tell?

A simple fix would be to have the No button on the pop-up dialog simply close the dialog, and not the app.

cpeterso:
@cskau: Good suggestion!

@freekzy: Can the `checkGps()` alert's `setNegativeButton()` handler just return without calling `finish()`?

PR #191, opened 2013-11-05
Fix for checkGps

rodmoreno:
It is a small fix for the issue #189, I think you could improve. I do not know, you who think @freekzy


cpeterso:
@rodmoreno: I haven't tested the code, but I thought the problem @cskau described in issue #189 was that the `finish()` method call would exit the app? What happens with your code change if GPS is disabled and you press the "No" button on the `checkGps()` dialog?

@freekzy wrote the original `checkGps()` method, so he would know best how it works.

Issue #144, opened 2013-10-29
Fall back to Google's SupportMapFragment for Gingerbread


cpeterso:
Google's `MapFragment` is only available in Honeycomb and later. We must use the `SupportMapFragment` class:

https://developer.android.com/reference/com/google/android/gms/maps/MapFragment.html#getMap%28%29

cpeterso:
Fixed by PR #202.

PR #202, opened 2013-11-06
Use SupportMapFragment

rodmoreno:
For the issue #144, in my Jelly Bean works perfectly, unfortunately I have no GigerBread to test, but everything should be fine.


dougt:
lgtm!  thanks.

Issue #150, opened 2013-10-29
Stumbler doesn't exit anymore since 0.7.1


whimboo:
Today I have installed version 0.7.1 and now stumbler doesn't exit anymore. The tray icon is still around all the time. If it is a service now, can we at least add a menu entry to fully quit stumbler?

Thanks

cpeterso:
@whimboo: did you open the stumbler's map view? The map starts scanning (so it has WiFi access point names to ask the server about). After viewing the map, you may need to go back to the app's main UI and press the "Stop Scanning" button to be sure. When loading the map, we should probably just scan once.

whimboo:
No, I didn't. But while stumbler was running I got an upgrade notice and upgraded the application to the latest version. Seems like that this caused the problem. When I again click start and stop the notification area entry is gone.

dougt:
closing.  PR #206

PR #206, opened 2013-11-07
Fix for updater

rodmoreno:
For the issue #150

First service is stopped and then terminates the app. With this we get that if the user was scanning and updates the app, not the service is run indefinitely.

Although you should implement something to avoid losing data scanned by the user ...

What do you think?


rodmoreno:
I think I've done, I've done a quick test and everything turned out pretty good, the service was stopped successfully, the application is closed and then showed me the update to be installed.

cpeterso:
Looks good!

PR #215, opened 2013-11-10
Added german translation

DaAllexx:
After several new translations, I thought German was missing ;)

### DIFF
--- a/res/values-de/strings.xml
+++ b/res/values-de/strings.xml
@@ -1,0 +1,40 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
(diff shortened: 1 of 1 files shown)


DaAllexx:
Haven't seen, that's there is already a pull request with a german translation. Never mind, I started yesterday with it, and it seemed, I was too slow :D

guerda:
DaAllexx: I just created it today and pushed it, so I was just more impulsive ;)

hannosch:
On a purely subjective note: I think this translation is slightly better. @cpeterso, you want to merge this one? And then @guerda can suggest changes on this basis.

cpeterso:
LGTM!

PR #222, opened 2013-11-15
Update strings.xml

benoit-l:
Improved French translation


clochix:
(for the record, @benoit-l is one of the main french translator of Mozilla's products, and his corrections make sens, so I think you could merge them)

cpeterso:
Thanks, @benoit-l!

PR #223, opened 2013-11-16
Update further l10n strings for zh-TW

petercpg:


cpeterso:
@petercpg: Travis CI reports an XML error in your zh-TW changes:

```
> Building > :mergeDebugResourcesl.sax.SAXParseException; lineNumber: 40; columnNumber: 3; The element type "string" must be terminated by the matching end-tag "</string>".
```

petercpg:
@cpeterso I just fixed the missing <, it should be good now.

PR #226, opened 2013-11-19
Fix timezone in submitted data

illarionov:
SimpleDateFormat uses the system's default time zone by default.


cpeterso:
oops. Good catch!

PR #236, opened 2013-11-22
Minor additions to SSID blocklist

MikeNicholls:
Added two mobile device and two aircraft(!) SSIDs to the blocklist.


cpeterso:
Thanks, @MikeNicholls! It's interesting to learn the various ways Wi-Fi is used on moving vehicles. :)

Issue #242, opened 2013-12-10
Two preferences files?


xsoh:
I've just looked around the code and I've notice two preferences file, one called "preferences" which is inside "xml/preferences" and the other is inside the prefs.java which is called prefs.class.getName(); I'm I correct?
If that is true, shouldn't we change it and make it one file only?

cpeterso:
@xsoh: Thank you for reporting this bug! @illarionov fixed this is in PR #243.

PR #243, opened 2013-12-10
Save all preferences in a single file [#242]

illarionov:


cpeterso:
@illarionov or @xsoh: Can you explain how we have two different preferences files? Isn't PreferencesScreen.java calling mPrefs, so it should read and write the same preferences file? What behavior does setSharedPreferencesName() change in PreferencesScreen.java?

illarionov:
PreferenceActivity saves all preferences automatically. setSharedPreferencesName() and  setSharedPreferencesName() sets the name and mode of the SharedPreferences file.

Without this patch, the settings are saved twice. mPrefs saves them into PREFS_FILE while the PreferenceActivity saves into default shared preferences file.

cpeterso:
Thank you,@illarionov! :)

Issue #241, opened 2013-12-03
Don't force Wi-Fi always ON


vinitraje:
Every time we turn on scanning MozStumbler switches on Wi-Fi, which should be optional. Because in a country like India Wi-Fi availability is very scarce. 

It consumes a whole lot of battery. Maybe a button should be introduced to turn on/off Wi-Fi when needed.

If the Wi-Fi is turned off manually MozStumbler won't send reports until its turned back on.

cpeterso:
Good idea. Scanning cell towers but not Wi-Fi access points would be a nice feature.

dougt:
thanks xsoh!

PR #248, opened 2013-12-12
Add Wi-Fi Access Points scanning option

xsoh:
Adding Wi-Fi Access Points scanning in setting for enhancement on #241

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -36,5 +36,8 @@
 
     <string name="about_version">Version %1$s</string>
     <string name="about_more">More...</string>
+    
+    <string name="enable_wifi_scanning">Wi-Fi Access Points Scanning</string>
+    <string name="enable_wifi_scanning_longtext">This will take effect on the next start</string>
 
 </resources>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -36,8 +36,5 @@
 
     <string name="about_version">Version %1$s</string>
     <string name="about_more">More...</string>
-    
-    <string name="enable_wifi_scanning">Wi-Fi Access Points Scanning</string>
-    <string name="enable_wifi_scanning_longtext">This will take effect on the next start</string>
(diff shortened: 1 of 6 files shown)


cpeterso:
@xsoh: thanks for your pull request! Your fix looks good, but I just thought of an alternate solution after rereading @vinitraje's issue #241.

Instead of adding a new preference and UI, could we just change `WifiScanner` to check whether the phone's Wi-Fi is enabled? @vinitraje's original concern was that MozStumbler would enabled Wi-Fi. `WifiScanner.start()` could check if Wi-Fi is disabled and then return early without doing anything. With my proposed solution, the user could disable their phone's Wi-Fi if they didn't want to scan Wi-Fi access points.

vinitraje:
@cpeterso That's what I was trying to say. Instead of turning ON Wi-Fi every time, scan only when Wi-Fi is turned ON by user.

And make the reporting independent of Wi-Fi status. Because now when we try to scan with Wi-Fi turned off MozStumbler won't  send any reports until Wi-Fi is turned back ON.

xsoh:
@cpeterso  & @vinitraje Thanks for reporting! I'll try to do it.

cpeterso:
@xsoh: Looks great! To keep our git revision history "clean", can you please "squash" your three separate commits into one? You can use `git rebase --interactive`, but here is a simpler method. The first command will "uncommit" your three commits. The second command will then re-commit them as one single commit with the new commit message.

```
git reset --soft HEAD~3
git commit -m "Don't scan when Wi-Fi turned off"
```

Here is a good tutorial about using git rebase to "squash" multiple commits into one: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

xsoh:
Oops, Sorry I'm new to git :S

cpeterso:
Looks good. Thanks for your help, @xsoh! :)

PR #252, opened 2013-12-12
Custom tile server

thefinn93:
The current version uses OSM, but is locked to MapBox without some code changes. This pull allows any OSM tile server to be specified at build time. I also updated the README to better explain the process. I'm pretty new to Java/Android programming outside of school, so please let me know if this is breaking all the rules of what's normal/good practice/whatever. ~~I'm still trying to dig up a nice list of public(ish) tile servers, so the README can link to that instead, but I'd like to see what people thought of the code in the mean time~~ Edit: linked to the [switch2osm](http://switch2osm.org/providers/) list.

### DIFF
--- a/AndroidManifest.xml.in
+++ b/AndroidManifest.xml.in
@@ -66,6 +66,10 @@
             android:value="MY_MAP_API_KEY"/>
 
         <meta-data
+            android:name="org.mozilla.mozstumbler.TILE_SERVER_URL"
+            android:value="TILE_SERVER_URL"/>
+
+        <meta-data
             android:name="org.mozilla.mozstumbler.API_KEY"
             android:value="MY_MOZILLA_API_KEY"/>
 
--- a/README.md
+++ b/README.md
@@ -19,12 +19,24 @@ StorePassword=<password>
 KeyAlias=<key alias>
 KeyPassword=<password>
 MozAPIKey=test
-MapAPIKey=<your MapBox API key>
+TileServerURL=<OSM Tile Server>
 ```
+For the OSM (Open Street Maps) Tile Server, you have several options:
 
-To obtain a MapBox API Key:
+* *[MapBox](https://www.mapbox.com/)* (easy, secure) is a hosted OSM solution, allowing users to easily create beautiful maps and featuring full SSL. To use MapBox:
 
-1. ?
+  1. Visit [mapbox.com](https://www.mapbox.com/) and sign up
+  2. From the [MapBox Dashboad](https://www.mapbox.com/dashboard/) click the big blue "New Project" button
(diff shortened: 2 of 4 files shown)


cpeterso:
Thanks for your fix, @thefinn93. It looks very good! :+1: 

Can you please use git rebase to squash your branch's changes into one commit? To preserve a clean, linear history, I prefer that branches rebase their changes on top of changes from master instead of merging. Lots of small commits make code reviews easier, but merge commits with conflicts make reviewing the code more difficult.

cpeterso:
@hannosch: Why do we use MapBox's tile server instead of OSM's tile.openstreetmap.org server? Does [y]our MapBox account give us some SLA?

hannosch:
@cpeterso because we aren't allowed to use the OSM tile server. Look at http://wiki.openstreetmap.org/wiki/Tile_usage_policy where it says: "OpenStreetMap data is free for everyone to use. Our tile servers are not." and later "Heavy use (e.g. distributing an app that uses tiles from openstreetmap.org) is forbidden ...".

Since Mozilla already has purchased a mapbox plan for mozilla.org, it made sense to re-use that one, instead of choosing one of the other options at http://switch2osm.org/providers/

thefinn93:
@cpeterso I'll try, I'm not very good at more advanced git stuff like that.. I fiddled around on my own but I think I just screwed up my repo even more. Could you give me a brief overview? sorry...

cpeterso:
@thefinn93, sorry for git confusion. It looks like your branch lost your original changes! Fortunately, I think you can probably restore your lost changes. I used `git reflog` to determine that the commit hash ID of your last good commit was 55e34b7885e3578787d5e7c54d7e15a4b8933358. 
- `git checkout 55e34b7885e3578787d5e7c54d7e15a4b8933358`
- then you can manually copy all the files you change to a safe directory (outside the MozStumbler directory).
- `git checkout master`
- `git branch -d customTileServer` (to delete your bad merge)
- `git checkout -b customTileServer` (to create a new branch with the same name as before so this same pull request will see your new branch)
- then manually copy your original files (from outside the MozStumbler) into the MozStumbler directory
- then add and commit all your changed files

I'm sure there is some fancier git rebase magic, but this method should work. <:)

thefinn93:
Okay, after much flailing, I think I have managed it. However github does not seem to be accepting the fact that I deleted a bunch of commits, so it might try to show them all. Would you like me to open a new pull with just the one commit?

cpeterso:
@thefinn93: thanks for your patience. Your current branch looks great and has just one commit. I have just one small suggestion (noted below).

I think the GitHub warnings about deleting commits are normal. If you try to push a git branch to GitHub after you have changed its commits, you need to use `git push --force` (which is how I assume you bypassed the GitHub warnings).

thefinn93:
Yeah that's what I did

cpeterso:
Thanks! :cake:

PR #254, opened 2013-12-15
Block DB on-train WiFi

evilpie:
See http://www.bahn.com/i/view/mdb/bahnintern/services/mobilitaet_service/zug/wlan/mdb_114824_www_im_ice_hot_spot_04_2013.pdf


cpeterso:
Thanks, @evilpie! :+1:

PR #260, opened 2013-12-27
fixes #171. make sure we sent to API key for both the submit and the sea...

### LINKED ISSUE #171: make sure we sent to API key for both the submit and the search API
If you are into cleanup: Can you make sure we sent to API key for both the submit and the search API's? If we want to use it for rate limiting / blocking spammers, we should do so on both.

dougt:
...rch API.


cpeterso:
LGTM

PR #262, opened 2013-12-29
Activity recognition

dougt:


dougt:
i can rebase some of these down.  Mostly looking for feedback at this point. @cpeterso

cpeterso:
@dougt: I tested your Activity PR this afternoon. It seemed to work. This will be a very cool feature!

The UI needs some love, but for now, some static text mentioning that the stumbler is actually doing something (without pressing a "Start Scanning" button) and that the Quit menu item will stop scanning.

If the new MozStumbler will always be passive, we might want always display a status bar icon (with different icons for the "stationary/not scanning" and "moving/scanning" states). Status bar icons are annoying, but I think one is appropriate if MozStumbler is actually running in the background and consuming battery. But _no_ animated icons ever! :)

dougt:
chris, please review.

cpeterso:
LGTM!

Issue #194, opened 2013-11-05
Show the number of cell towers scanned


marco-c:
At the moment only the AP number is shown.

cpeterso:
We used to display the number of cell towers scanned, but users found it confusing. But it would be straightforward to bring it back.

marco-c:
It'd be interesting for me, maybe you could add an optional indicator that could be enabled in the settings.

CupIvan:
yes, it will be very usefull to display all collected data: total and per interval (for example 5-10 minutes)

Djfe:
can be closed, right? :)

cpeterso:
Yep! Fixed by PR #320.

Issue #168, opened 2013-10-31
Change "GPS Satellites Visible" count to a descriptive string


cpeterso:
such as "GPS Status: Searching" then maybe "GPS Status: latitude,longitude". This UI change should reduce user confusion about whether their phone has a GPS fix and what is the "correct" number of satellites.

merike:
Please keep the count too or at least differentiate between no satellites and a few.. It it very useful to know if it is 0 satellites (terrible weather or otherwise poor conditions, might want to use battery for other things than stumbling) and 3-4 satellites visible (likely to get a gps fix soon).

cpeterso:
Fixed by PR #126. :)

PR #263, opened 2013-12-30
Add "Wi-Fi Access Points Visible" to main display

keverets:
This allows the user to see if any access points are currently visible,
as well as whether the number is changing which prevents stalled polls
from continuing.

### DIFF
--- a/res/layout/activity_main.xml
+++ b/res/layout/activity_main.xml
@@ -27,11 +27,20 @@
            android:textAppearance="?android:attr/textAppearanceMedium" />
 
        <TextView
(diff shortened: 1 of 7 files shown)


cpeterso:
Thanks, @keverets. This is a very good idea!

cpeterso:
Looks good to me! @keverets, thanks for your contribution and taking the time to address my suggestions. :)

If you're looking for other MozStumbler features to work on, you might be interested in issue #168 or #194. They are UI feature requests that are similar to the one you implemented.

PR #264, opened 2014-01-01
Increase minimum report size from 20 to 100 measurements

cpeterso:
This should reduce network usage and power consumption.

### DIFF
--- a/src/org/mozilla/mozstumbler/Reporter.java
+++ b/src/org/mozilla/mozstumbler/Reporter.java
@@ -26,12 +26,12 @@ import java.util.concurrent.locks.ReentrantLock;
 
 import android.content.BroadcastReceiver;
 
-class Reporter extends BroadcastReceiver {
+final class Reporter extends BroadcastReceiver {
     private static final String LOGTAG          = Reporter.class.getName();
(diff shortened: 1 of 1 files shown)


hannosch:
r+
It's an easy first fix. At city driving speeds you collect maybe 1 measurement per second (we collect a sample every 10 meters; 10m/s = 36km/h = 22mph). With a batch size of 20, we do one upload every 20 seconds. The power tail of waking up a cell radio is in the 10 second range, so we keep the radio active for 50% or more of the time (reference http://www.youtube.com/watch?v=dASOm88Wh8g).

With a batch size of 100, we upload every 100 seconds, so the radio can sleep for 90% of the time. That in itself should decrease the power consumption considerably.

cpeterso:
@hannosch: what is the upload limit on the server side?

hannosch:
I increased it to 100mb for the upload API, so basically none. A batch of 100 should work out fine. But I wouldn't increase it further, or we can run into timeout problems, if there's a lot of data to process and the server is generally busy. We'll have monitoring of request times soon enough - once we have real data we can tweak things further.

Issue #69, opened 2013-08-15
Filter known mobile device OUIs in MAC addresses


cpeterso:
Wireshark maintains a list of known OUIs for hardware manufacturers: https://anonsvn.wireshark.org/wireshark/trunk/manuf

dougt:
@cpeterso or do this on the server?

hannosch:
How do you determine if a manufacturer produces only mobile devices? As an example Apple certainly does both mobile phones and stationary devices. And what if a company currently doesn't offer any stationary devices, but introduces something new into the market or one of its chips is used in a device with a different company branding? How would we get notified about those kind of changes?

cpeterso:
I found some lists of purported iPhone OUIs, but I don't think there is a practical way to differentiate mobile and non-mobile OUIs from the same manufacturer.

hannosch:
@cpeterso can you point me to a list of those prefixes? I can check those against our current database. Might be interesting to see if we'd exclude any records via this at all.

cpeterso:
@hannosch: Here is a list of OUI prefixes have been corroborated by multiple websites as iPhone or iPad devices. You probably won't see many of these because MozStumbler filters devices with "iPhone" or "iPad" SSIDs.

```
00:1b:63
00:21:e9
74:e2:f5
78:d6:f0
7c:6d:62
7c:c5:37
88:c6:63
8c:77:12
```

And a longer uncorroborated list: http://community.arubanetworks.com/t5/Unified-Wired-Wireless-Access/CLI-command-for-finding-all-iPhone-MAC-addresses/m-p/55572#M22065

hannosch:
I've checked our DB. We currently have about 5000 WiFi entries for these prefixes and an additional 7 on our blacklist. Many of those 5000 entries have only a single measure, but there are some which have been seen 20 or more times in the same place.

Most likely we picked these up while people are at work. It's unlikely we'll also catch them on their way home or at home.

Not quite sure what to do about this, but the filter seems to be effective, so it's a good idea to pursue this.

cpeterso:
hmm, so those iPhone locations would sometimes be useful for identifying home or office locations. If we eventually had enough measurements, we could filter out single measurements but keep those 20+ measurements in the same location. Though that comes back to privacy issues about knowing a particular phone's home and office locations..

garvankeeley:
Closing due to staleness, and we have a pretty active block list, so I don't think we need to do anything here.

PR #265, opened 2014-01-01
Filter some known iPad and iPhone BSSID OUIs

cpeterso:
Using the partial list of iPad and iPhone OUIs from issue #69.


rtilder:
What's the source for these OUIs?

dougt:
I asked the same question when reviewing the patch.  Read the first comment again.

rtilder:
Ah.  In issue #69.  Should be added to a section in the readme or similar rather than just in a bug along with any other sources for the hard coded lists.  Added #277.

PR #266, opened 2014-01-01
Provide GPS Latitude & Longitude to user

keverets:
Issue #168 suggested a change to the "GPS Satellites Visible" count so the user was aware if a GPS fix has occurred.  This merge request provides an additional line of information informing the user of the current location as reported by the GPS.  That information is only displayed if there are currently visible satellites and at least one location has been observed.


cpeterso:
Looks great! Thanks, @keverets. :+1:

PR #269, opened 2014-01-05
Collect unique data at each location

illarionov:
Now if you stand long on one place, the GPS location may not to be changed for more than one second. 
Once the report with this location is sent, all further scanned WI-FI networks and GSM cells are discarded.

In this patch, sending a report is delayed until a new GPS position is received, and all received durring this time results of scanning are grouped into the one report.


dougt:
Hey @illarionov,

How does this change merge the 3 different signals (wifi, cell, gps)?  As it is now, we basically continually watch all three and when we see new values for each, we report.  

One of the things that I was worried about was if I was stumbling from a car.  As I am driving down the road at 55MPG, what I don't want to have happen is to:

1) get a new gps position, then
2) kick off a wifi/cell scan, then
3) when the scan returns, report.

My worry is that you'll have something like:

```
----------------------------------------------------------[my car]----->
^--- gpsfix here        ^                           ^
                        +--- scan starts here       |
                                                    +--- scan ends here
```

So, basically there will also be a delta between the location and the wifi/cell - dependent on your speed.

 I am wondering if your approach makes what we are currently doing any better.

dougt:
(and ascii art fail, but I hope you get the gist of what I mean).

Also, have you looked at the Activity recognition pr.  Have you thought about how doing something like that, impact this change?

cpeterso:
@dougt: I fixed your ASCII art formatting and added the "scan ends here" marker.

btw, in my local testing, the latency from Wi-Fi scan start to end is ~700 ms. Our current scanner implementation scans on a 3000 ms timer, so those scan results can be up to 3000 ms "too early", compared to ~700 ms "too late" if we start a scan when we receive a GPS fix event. The 700 ms measurement was very consistent, so I suspect Google's implementation starts listening for Wi-Fi beacons, waits a fixed time period (500-700 ms?), then then just report whatever scan Wi-Fi beacons its seen.

A hybrid solution: when we receive a GPS fix event, if the most recent scan data is less that 1000 ms old, then use it. If the scan data is more than 1000 ms old, then initiate a new scan because that data will be closer to the GPS fix location.

dougt:
okay, sounds fine -- @cpeterso do you want to take crack at reviewing this sometime this week?

cpeterso:
@dougt, @illarionov: Sure. I will try to review it over the next few days.

illarionov:
When moving at high speed nothing should change. The GPS scanner is configured to send location updates when the position has changed by at least 10 meters but not more often than once per second. If the position is not changed within 10 meters, the interval between two adjacent updates can be much more than 1 second.

```
        |         .                  .
Wi-Fi   |         .                  .
scanner |         /\/\/\/\      /\/\/\/\
        |         .                  .
Cell    |         .                  .
scanner |         /\/\          /\/\ .
        |         .                  .
        ----------------------------------------->t
                  ^                  ^
                  GPS fix 1          GPS fix 2

```

Currently, when the first cell scan completes (on ASCII art), the Reporter creates report, and then executes resetData() that clears all collected measurements, including data about current position (mGPSData).  Data of the first wi-fi scan will be destroyed if they do not fall within REPORTER_WINDOW,  or otherwise they will be reported as data with coordinates of GPS fix 2.
In proposed solution, the first wi-fi scan will be reported with coordinates GPX fix1,  second wi-fi scan - at GPX fix2, and results of both cell scans will be grouped by Cell ID and reported with coordinates of GPS fix 1.
When driving at high speed, the location is updated mostly every second and all results of scanning are reported with the coordinates of the last received position at the moment the scanning is finished.

I have not looked at the Activity recognition yet, but do not think it has any influence. As I understand it, the scan will be simply turned off when there is no movement. This is a nice addition.

cpeterso:
@illarionov: your PR looks great! I have a couple suggestions that can be fixed in a future PR. I want to merge your branch now to get some testing from developers (before publishing a new MozStumbler release to all users). :)

PR #278, opened 2014-01-08
Brazilian portuguese translation

RicardoMurad:
Sending brazilian portuguese translation


cpeterso:
Thanks, @RicardoMurad!

PR #282, opened 2014-01-09
Some reporter fixes

illarionov:
HttpsURLConnection connection reusing is broken on android 4.0 - 4.4: http://stackoverflow.com/questions/19258518/android-httpurlconnection-eofexception


illarionov:
I do not think that we should touch any other HTTP headers. Maybe it would be better to disable keep-alive connections on the server and not on the client (nginx configuration: keepalive_timeout 0;).

I have rebased from master and added a comment. This request is ready to merge.

cpeterso:
Tweaking the server's settings instead of the client's make sense.

cpeterso:
Thanks, @illarionov!

PR #283, opened 2014-01-09
Do not report accuracy and altitude if they are not available.

illarionov:
Round accuracy and altitude to 2 decimal places.

### DIFF
--- a/src/org/mozilla/mozstumbler/LocationBlockList.java
+++ b/src/org/mozilla/mozstumbler/LocationBlockList.java
@@ -41,22 +41,22 @@ final class LocationBlockList {
             }
         }
 
-        if (inaccuracy < 0 || inaccuracy > MIN_ACCURACY) {
+        if (location.hasAccuracy() && (inaccuracy < 0 || inaccuracy > MIN_ACCURACY)) {
             block = true;
             Log.w(LOGTAG, "Insufficient accuracy: " + inaccuracy + " meters");
         }
 
-        if (altitude < MIN_ALTITUDE || altitude > MAX_ALTITUDE) {
+        if (location.hasAltitude() && (altitude < MIN_ALTITUDE || altitude > MAX_ALTITUDE)) {
             block = true;
             Log.w(LOGTAG, "Bogus altitude: " + altitude + " meters");
         }
 
-        if (bearing < 0 || bearing > 360) {
+        if (location.hasBearing() && (bearing < 0 || bearing > 360)) {
             block = true;
             Log.w(LOGTAG, "Bogus bearing: " + bearing + " degrees");
         }
 
-        if (speed < 0 || speed > MAX_SPEED) {
+        if (location.hasSpeed() && (speed < 0 || speed > MAX_SPEED)) {
             block = true;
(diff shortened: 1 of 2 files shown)


cpeterso:
I like the idea of not uploading accuracy or altitude if the client doesn't have them, but I'm worried that the server code will break if the "accuracy" or "altitude" fields are missing.

@hannosch: In the following server code, is the `process_measure()` parameter `data` a regular Python dict? Will `data['accuracy']` throw a KeyError if the request does not include accuracy:

https://github.com/mozilla/ichnaea/blob/cd05051e645d67f2f0e579d96cdab953071fab88/ichnaea/service/submit/views.py#L126

hannosch:
@cpeterso the client doesn't have to sent those fields. The server code is a bit tricky to understand, but there's a schema validation step which will fill in all "missing" fields with appropriate default values. So in most cases sending a field with a dummy value or not sending the field at all will be no different.

There are very few actually required fields. The definition is at https://github.com/mozilla/ichnaea/blob/master/ichnaea/service/submit/schema.py - any entry that has a `missing=` statement is optional.

illarionov:
@hannosch, altitude=0 (0 meters above mean sea level) is a valid value for altitude.
Shouldn't altitude and accuracy be Decimal instead of Integer?  Accuracy < 1 is rounded to 0 now.

hannosch:
@illarionov using 0 as the default value isn't ideal there. Using -1 doesn't make it any better, since there are land areas where that's a valid value as well. Probably anything to -10 or -20 is valid in some parts of the world (without actually being under water). So we'd have to use a large negative number. So far we haven't bothered to improve this, as we aren't actually using the altitude number yet.

For integer vs. decimal: Trying to be super precise isn't really buying us anything. Most GPS accuracy numbers are in the 5-10 meters range, with a valid long tail to several 100 meters in urban environments. So we can treat anything in the <10 meter range as being about equally precise. And if we don't know the accuracy, we can only assume it to be precise as well. It's about as likely that the phone sensors are lying about the accuracy, as it is to get no value.

illarionov:
Rounding to an acceptable precision can be performed on both the client and the server side. Where will it better to do?
BTW, the lat and lon now sent without rounding. I see 14 decimal places in the reports. It can be rounded to 6 or 7 decimal places.

cpeterso:
@illarionov: Can you please make the following changes?
1. Round the lat/long to 6 decimal places, as you suggested. Rounding on the client lets us reduce the amount of data stored locally and uploaded (if only by a small amount). 6 decimal places (8 and 9 significant digits for latitude and longitude, respectively) should fit each value in a 32-bit floating point number, so I think that's an elegant limit. :) 6 decimal places still gives us better than 1 meter precision.
2. Rebase this PR on top of the latest master.

illarionov:
@cpeterso, done

hannosch:
We store lat/lon as integers on the server side retaining 7 decimal places. That way the data fits into a 32bit signed integer as well. We only really care about 6 decimal places, but 7 doesn't increase the server side storage requirements.

As each decimal place gets sent as one byte over the network, rounding down to 6 decimal places on the client is fine by me.

cpeterso:
LGTM!

PR #284, opened 2014-01-10
Suggestion for new icons

jrconlin:
Goofing around a bit with a better icon for Mozilla Stumbler.

No, my feelings won't be hurt if you don't use them.

Also may try to fix up the main page to look a bit better and be functional in landscape mode.


cpeterso:
Fun! The stumbling sign is a little crooked, but that adds to its charm.

I will let @dougt merge this PR because, as designer of the original icons, I am biased. :)

dougt:
hah.  i'll never forget the older icon.

PR #291, opened 2014-01-13
Update Hindi & Marathi translation

vinitraje:


cpeterso:
Thanks, @vinitraje!

btw, do you have any recommendations on the best way to alert localizers like yourself when new strings are added?

vinitraje:
@cpeterso I'm sorry. There's a small typo in Hindi translation. How should I fix that?

And about alert, we can put a notice in about section of app for contributors.

cpeterso:
@vinitraje: to fix the Hindi typo, just submit a new PR. :)

PR #292, opened 2014-01-13
Set default values on start

illarionov:
This fixes a bug that when open settings on a fresh installation, power saving mode will be shown turned off when in fact it is on.


cpeterso:
LGTM! Defaulting power_saving_mode pref to true was on my wishlist. :)

PR #294, opened 2014-01-13
Bump Android gradle plugin verion to 0.7.+

illarionov:
For Android Studio 0.4.2


cpeterso:
@illarionov:  gradle 1.9 adds an Android Lint build stage, which fails on some MozStumbler warnings. So we can't use gradle 1.9 until we either disable to gradle's Lint build stage or fix all the warnings (which is highly unlikely).

PR #298, opened 2014-01-17
zh-TW l10n update

petercpg:


cpeterso:
Thanks for the update, @petercpg!

PR #301, opened 2014-01-18
Restore ongoing notification with a min. priority (removed in acd2e0f)

illarionov:
Start the service in a foreground state.

Why this notification has been removed? Removing startForeground() and making service a background was intentional?


cpeterso:
Thanks for catching this, @illarionov! The removal of `startForeground()` might explain some of the recent stumbler problems I have been seeing. :)

PR #304, opened 2014-01-19
Rewrite starting and stopping of the activity recognition

illarionov:
Management of a subscription to an activity recognition updates should be done from the main activity.
Activity recognition service can send them even when the application is stopped or unloaded from memory.

In particular, fixed bug that  the activity recognition service continue to send updates after exiting the application and after turning off the power saving mode.


dougt:
the way I see things is that the application can control the service, get information about what the service is currently doing, and can set preferences.  The service should start and stop based on activity monitoring to conserve battery

Why should it be the other way around?

cpeterso:
I like the changes to gracefully handle devices that don't have Google Play Services' Activity Recognition, but I agree with @dougt that the UI activity should just be a client of the smart, activity-aware service.

PR #307, opened 2014-01-20
Show notification when scanning

cpeterso:
The `NotificationCompat.Builder` does not seem to actually do anything on my device, so I resurrected the older, proven `buildNotification()` method that does display the notification icon.

### DIFF
--- a/src/org/mozilla/mozstumbler/ScannerService.java
+++ b/src/org/mozilla/mozstumbler/ScannerService.java
@@ -21,7 +21,6 @@ import android.os.Handler;
 import android.os.IBinder;
 import android.os.Looper;
 import android.os.RemoteException;
-import android.support.v4.app.NotificationCompat;
 import android.text.format.DateFormat;
 import android.util.Log;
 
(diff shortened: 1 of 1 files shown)


illarionov:
It does not show icon with NotificationCompat.PRIORITY_MIN by default. You can try to change to NotificationCompat.PRIORITY_LOW or NotificationCompat.PRIORITY_NORMAL.

cpeterso:
@illarionov: I did not realize that. What is the any advantage of using NotificationCompat Builder over just using the "deprecated" Notification API?

illarionov:
I have used NotificationCompat.Builder to set the priority of the notification trying to make it less annoying.
Besides, it allows to use notification features of the newest Android versions without having to worry about compatibility. http://developer.android.com/design/patterns/notifications.html

cpeterso:
That makes sense. I'll revert my change, but I think we do want to display a notification icon if we are running in the background.

We currently only display the notification when the service is actively scanning. However, this means that in power-saving mode we will only show the notification when the user is actively moving. I'd like to change this so we always display the notification icon when the service is running. Otherwise, we might be draining people's batteries, but when they look at their screen, they don't see any notification that might point the blame to our app.

PR #314, opened 2014-01-21
Some minor UI changes

cpeterso:
These are some minor UI changes I made when I was prototyping how we might bring back the "Start/Stop Scanning" button (or something like it).


phrostb:
sounds good bro!

PR #319, opened 2014-01-22
Fix lint issues with de translations

keverets:
lint really wants "Schliessen" to be "Schließen" and "Fuss" to be "Fuß".
This change makes lint happy.


cpeterso:
Thanks! Android Lint is pretty smart. :)

PR #320, opened 2014-01-22
Display current and total cell info to the user

keverets:
I like to see that even when I'm not getting WiFi APs, there's still valuable info being stumbled.  This is a first stab at satisfying issue #194 ... though that wanted individual cell tower, this is currently counting unique "Cell Info" which includes things like Signal Strength.

The equality and hashCode for CellInfo could be changed to exclude non-tower-identifying info if required.

### DIFF
--- a/res/layout/activity_main.xml
+++ b/res/layout/activity_main.xml
@@ -53,11 +53,29 @@
            android:textAppearance="?android:attr/textAppearanceSmall" />
 
        <TextView
-           android:id="@+id/visible_wifi_access_points"
+           android:id="@+id/current_cell_info"
            android:layout_width="wrap_content"
            android:layout_height="wrap_content"
            android:layout_alignParentLeft="true"
            android:layout_below="@+id/locations_scanned"
+           android:text="@string/current_cell_info"
+           android:textAppearance="?android:attr/textAppearanceSmall" />
+
+       <TextView
+           android:id="@+id/cell_info_scanned"
+           android:layout_width="wrap_content"
+           android:layout_height="wrap_content"
+           android:layout_alignParentLeft="true"
+           android:layout_below="@+id/current_cell_info"
+           android:text="@string/cell_info_scanned"
+           android:textAppearance="?android:attr/textAppearanceSmall" />
+
+       <TextView
+           android:id="@+id/visible_wifi_access_points"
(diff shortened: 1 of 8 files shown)


Djfe:
can you draw the celltowers on the map (test mozilla location service button)?
I know atleast one other app (like mozstumbler) that draws them (by recognizing where they are) and even adds specific data like names/ids of the cell towers

cpeterso:
Thanks, @keverets!

cpeterso:
@keverets: After testing your cell info count, I was wonder why the stumbler claimed that it had scanned of thousands of cells in one day. I now realize what you meant above by "cell info" vs "cell" because the counter is displaying the number of different _scans_, not unique cells. I think we might want to rename the display title or, better yet in my opinion, only display the number of unique cells scanned, like we do for Wi-Fi APs.

Djfe:
so that is the reason why my cell count climbs up so fast and my visible cell count switched between 1 and another number all the time :)
I was really wondering about that for some time

so the cell towers just get scanned in a certain Intervall and during scanning only the used tower is really visible

I'm agreeing with you that it would be better to only display scanned unique cells

btw. if you check 2g only in the options of your Android device
will it be impossible to detect 3g cells for mozstumbler?

while we are at it: do you think it would make sense to differentiate between scanned 2g/3g/4g cells, and to differentiate between scanned b/g/n/ac wifi aps in the list

keverets:
@cpeterso Sorry for not being more clear about the issue.  I'll look at making it track actual Cells rather than Cell Readings (which include the signal strength, etc).  Would you prefer a new "Cell" object to be created that could be incorporated into the CellInfo, or keep the objects as they are and change the equivalence/hashing of CellInfo to disregard non-identifying Cell information (again, such as Signal Strength)?

cpeterso:
@keverets: I think a new `Cell` object that stores the cell's identifying information sounds like a good idea. So a `CellInfo` would have-a `Cell` and measurements like signal strength and scan timestamp? CellScanner.java only uses `mCells` to track the number of scanned cells, so `mCells` would only need to store `Cell`s, not `CellInfos`.

Djfe:
Still not fixed I think.
The cells scanned counter still doesn't show only unique cell towers.

PR #332, opened 2014-01-23
Updated german translations

DaAllexx:
Updated german translations with new values from version v0.10.0, and added a simple comma.


cpeterso:
Thanks, @DaAllexx!

PR #343, opened 2014-01-25
Update zh-tw translation

petercpg:


cpeterso:
Looks good. Thanks, @petercpg and @irvin!

PR #351, opened 2014-01-27
Added Italian translation

musuruan:
I created an Italian translation using android2po and Poedit under Linux. I hope everything is fine - this is my second Android translation.


cpeterso:
Looks great. Thanks for your help, @musuruan!

PR #355, opened 2014-01-28
Downgrade to gradle 1.9 to fix fdroid builds

cpeterso:
fdroid's build system appears to still be using gradle 1.9:

https://f-droid.org/wiki/page/org.mozilla.mozstumbler/lastbuild

```
Cleaning Gradle project...
Could not build app org.mozilla.mozstumbler due to BuildException: 'Error cleaning org.mozilla.mozstumbler:0.10.2'
==== detail begin ====
Available gradle versions: 1.4 1.6 1.7 1.8 1.9
Found 1.10 via gradle plugin version 0.8
/opt/gradle/bin/gradle: line 18: /opt/gradle/versions/1.10/bin/gradle: No such file or directory
==== detail end ====
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,6 @@ StoreFile=<path to file>
 StorePassword=<password>
 KeyAlias=<key alias>
 KeyPassword=<password>
-MozAPIKey=test
 TileServerURL=<OSM Tile Server>
 ```
 For the OSM (Open Street Maps) Tile Server, you have several options:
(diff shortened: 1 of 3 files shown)


cpeterso:
btw, this PR also removes `MozAPIKey` from the readme's example gradle.properties file because it is only needed for official Mozilla builds and the existence of `MozAPIKey` will cause builds by other people to attempt to update to Mozilla-signed builds.

Issue #352, opened 2014-01-27
battery empty -> scan button doesn't switch from on to off state


Djfe:
if the battery of phone is almost empty and mozstumbler stops scanning to save on battery it doesn't switch the buttons light
![screenshot_2014-01-27-14-37-50](https://f.cloud.github.com/assets/5685512/2010134/b3e7f84a-8769-11e3-9d7a-c8a709e7fb00.png)
![screenshot_2014-01-27-14-38-01](https://f.cloud.github.com/assets/5685512/2010135/b4e28a94-8769-11e3-8d52-b80d64b81084.png)

Djfe:
the blue color on the button doesn't appear if you start scanning by opening the map, too.

tamcap:
I believe my commit should fix the latter problem. No idea why updateUI() is not invoked in the first case.

Djfe:
thx ;)

tamcap:
Don't close it just yet - I do not think the fix will be appropriate for the first part of your problem. I have honestly not been able to experience it myself (but I am an obsessive battery charger).

Edit: Actually, let's test it. It seems updateUI() should do the job after all.

Djfe:
works thx for the easy fix ;) (tested with v0.11.0

PR #360, opened 2014-01-28
Issue #352 fix (on/off button state)

tamcap:
This commit should ~~partially~~ fix Issue 352 by replacing Button with CompoundButton in updateUI() method and invoking setChecked().


tamcap:
I believe now that the fix will address all the parts of Issue #352. I misunderstood it previously.

PR #361, opened 2014-01-28
Updated Finnish translations

ojarva:
Fixed jargon and spelling mistakes.


cpeterso:
Thanks, @ojarva!

ojarva:
I think it's something that definitely should be translated - I don't think people who use their phone in Finnish want to start guessing what "Location Service" means. Obviously, Mozilla shouldn't be translated/modified (there's "n" at the end, as that's possessive suffix in Finnish).

cpeterso:
That make sense. Thanks for the explanation. I merged your translations as you submitted them.

PR #362, opened 2014-01-28
Issue #352 fix (on/off button state)

tamcap:
_Got confused previously with multiple commits (sorry)_. This commit fixes the light on the on/off button by replacing the Button class with a CompoundButton and invoking setChecked() during UI update.

PR #368, opened 2014-01-30
Override scanning button's default ON/OFF text

cpeterso:
This fixes a regression from PR #362.


tamcap:
Hey, @cpeterso, what happened?

cpeterso:
@tamcap: calling `scanningBtn.setChecked()` after `scanningBtn.setText()` would reset the "Start/Stop Scanning" button's text to the default the ToggleButton's default checked/unchecked text ("ON/OFF").

The original `scanningBtn.setText()` code was a little hacky. A cleaner fix would be to define the ToggleButton styles in our resource files.

PR #373, opened 2014-01-30
sync zh-CN translation with en-US strings

l-hedgehog:


cpeterso:
Looks good. Thanks, @l-hedgehog!

Issue #370, opened 2014-01-30
Submit Results via WiFi Only 


ghost:
It would be nice if MozStumbler could store scan results even after being closed and submit when WiFi becomes available if the user so chooses in settings.

Djfe:
it stores all informations, but the the UI has a bug so that the statistics don't show collected data anymore
it will be send the next time the phone gets an internet connection
see #305 for more information ;)

wifi only would be a nice enhancement, too (for devices without a flatrate)

AdamHauner:
Similar to #378

tamcap:
Can be closed, fixed by #378.

PR #378, opened 2014-02-03
Introducing WiFi only uploading.

tamcap:
Those changes should introduce a WiFi only upload feature. My only concern is the size of `mReports` getting too large. It's possible that an extreme value (`mReports.length`>?) might cause some buggy behavior down the line.

I'm pretty sure I messed up something with github again. Some of the changes are already incorporated into mozilla:master (related to the on/off button), but as they are by now part of the same commit, I am not sure how to separate them (/ revert them).

### DIFF
--- a/res/layout/activity_main.xml
+++ b/res/layout/activity_main.xml
@@ -98,7 +98,8 @@
            android:layout_alignParentRight="true"
            android:layout_below="@+id/reports_sent"
            android:onClick="onToggleScanningClicked"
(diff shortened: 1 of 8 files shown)


AdamHauner:
Similar to #370

tamcap:
will do

cpeterso:
LGTM. Thanks, @tamcap! Users have been asking for this feature. :)

PR #383, opened 2014-02-03
Geofencing for MozStumbler

tamcap:
This very basic interface allows to specify a point (in a format of `-10.005,10.02`) around which MozStumbler will not report the results. Exclusion zone is currently defined as +/- 0.01 deg lat/lon.


tamcap:
Geofencing works good :+1:

tamcap:
Abandoning the pull request due to mis-formed commits.

PR #386, opened 2014-02-04
Basic geofencing.

tamcap:
This very basic interface allows to specify a point (in a format of `-10.005,10.02`) around which MozStumbler will not report the results. Exclusion zone is currently defined as +/- 0.01 deg lat/lon.

Geofencing works good :+1:

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -57,8 +57,8 @@
     <string name="upload_wifi_only_summary">This option will reduce your data usage over the cellular network.</string>
 
     <string name="geofencing_off">Geofencing off (0,0)</string>
-    <string name="geofencing_title">Geofencing options</string>
-    <string name="geofencing_desc">By entering coordinates here, you instruct MozStumbler to ignore locations around certain point.</string>
+    <string name="geofencing_title">Geofencing settings</string>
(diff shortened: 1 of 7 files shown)


cpeterso:
@tamcap: should I be reviewing this "Basic geofencing" pull request or the `better_geofencing` branch in your tamcp/MozStumbler repository?

tamcap:
This is the basic functionality. Geofencing will work with commit f405761, everything else is just extra features / UI. I think the best way forward is to review this and then any changes will derive from here. If there are some serious (low-level) changes you want to make, I will implement them in both places.

tamcap:
Do you think it's ok to invoke `Prefs.getLat()` and `Prefs.getLon()` everytime `contains()` is called? That was my original motivation for passing it down from GPSScanner.

tamcap:
There are 2 exceptions I am worried about: `NumberFormatException` and `IndexOutOfBounds` (coming from `Substring()`. I will catch them both and just abort in that case.

tamcap:
No complaints on the implicit float + string.

tamcap:
It seems we are moving forward with the more involved geofencing, this pull request can be closed.

PR #389, opened 2014-02-06
Better geofencing

tamcap:
This branch introduces improvement over the basic geofencing. Now with dedicated menu item and display status. [Still] TODO: add indicator when geofencing is actively blocking locations.


tamcap:
Needs work.

PR #390, opened 2014-02-07
Better geofencing

tamcap:
Those are upgrades built on top of the `exp` branch introducing upgrades to the UI and some code cleanup.

New features:
- ability to set geofencing location from the drop down menu during scanning
- geofencing status display on the main screen
- checkbox in the preferences to turn geofencing on/off
- 2 way communication of geofencing status between the blocker and main screen

### DIFF
--- a/res/layout/activity_main.xml
+++ b/res/layout/activity_main.xml
@@ -3,7 +3,6 @@
     android:focusableInTouchMode="true"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
-    android:paddingBottom="@dimen/activity_vertical_margin"
     android:paddingLeft="@dimen/activity_horizontal_margin"
     android:paddingRight="@dimen/activity_horizontal_margin"
     android:paddingTop="@dimen/activity_vertical_margin"
@@ -89,6 +88,15 @@
            android:text="@string/reports_sent"
            android:textAppearance="?android:attr/textAppearanceMedium" />
 
+        <TextView
+            android:id="@+id/geofence_status"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_alignParentLeft="true"
+            android:layout_below="@+id/reports_sent"
+            android:text="@string/geofencing_off"
+            android:textAppearance="?android:attr/textAppearanceMedium" />
+
        <ToggleButton
            android:id="@+id/toggle_scanning"
            android:layout_width="wrap_content"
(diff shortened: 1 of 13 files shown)


cpeterso:
@tamcap: The "Geofence this location" menu item seems to be unused, but it's a very interesting idea.

Could you replace the Settings activity's "Geofencing settings" menu item and text input with a Settings button to "Geofence current location"? That way, the user does not need to figure out their lat/long position so they can manually enter it. And you can remove your code that handles the input and error checking.

I'm imagining the Settings screen could have:
- a "Pause scanning within geofence" checkbox to enable/disable geofencing
- a "Geofence current location" button to set the geofence to the current lat/long (and smaller text that displays the current geofence lat/long)

tamcap:
Not sure what do you mean by unused? It works just fine for me... 

I can make a fake button in the prefernces section with an `onClick` handler. Not sure how well it will look like, but I will give it a shot.

tamcap:
You have to be scanning so that menu option becomes active. -- which probably could be changed to behave in a similar fashion to test location menu option.

tamcap:
Rough draft of the proposed changes. Geofencing is set when scanner is scanning - this avoids the need to start/stop the scanner from the prefs menu - the latter is possible, however, it would complicate the code and execution needlessly (IMHO).

Also, slight adjustment to the bottom margin. The button was becoming squashed in the horizontal orientation otherwise.

tamcap:
~~Tiny detail: it doesn't work. Debugging.~~ All is good.

cpeterso:
Thanks, @tamcap! I will probably release the next minor version of MozStumbler before I merge your changes because I'd like to get some more testing and feedback from other contributing developers.

cpeterso:
@tamcap: can you please rebase your geofencing PR onto the latest master and squash your commits into just one? Your changes can no longer be merged automatically because of other changes on master. Sorry it's taken me a while to review this PR. It's a bit complicated. :)

I'm a little concerned by all the Prefs code necessary for this feature's UI, but I'm not sure if there is a cleaner alternative.

tamcap:
Done.

We might be able to get rid of `geofence_here` pref, because the only thing it does is notify the main  application to do something. There must be a way to do it in an easier and more elegant fashion. Feel free to hold off on the merge yet, if it really gives you heebie-jeebies :)

tamcap:
~~Hold off on the merging. I've got an idea.~~ I was thinking about using a static setter / getter in the Prefs.java accessed at `onResume()` from MainActivity.java. But it's ugly. And honestly doesn't improve much...

tamcap:
Honestly, we are probably stuck as it is. A new intent is an option, but then it will cloud MainActivity code...

cpeterso:
Let's merge your feature and test it on master for a bit. We should watch out for problems where people using "Upload only on Wi-Fi" might geofence their home if that's the only Wi-Fi AP they connect to. :)

Djfe:
you could implement a proper force send button to prevent this problem

PR #391, opened 2014-02-07
translate strings about wifi

CupIvan:


cpeterso:
Looks good. Thanks, @CupIvan!

PR #392, opened 2014-02-09
Poland Koleje Slaskie train hotspot blocklist

merkato:
Koleje Śląskie, local authorities subsidiary railway service in Poland has several trains with wifi hotspots. They are named by "Hot-Spot-KS*" where \* is an rolling stock number.


cpeterso:
Looks good. Thanks, @merkato!

PR #394, opened 2014-02-09
Cleaning up the error output of MainActivity.java

tamcap:
Simple `Log.e->Log.d` cleanup. Most of the intents received by `MainActivity.java` are not errors. They can be simply handled on the debug level to keep the error console cleaner.


tamcap:
Heh, we are getting those noisy stacks already on CDMA/no networks, so I know what you mean. I will get to it in a second.

cpeterso:
Looks good!

PR #400, opened 2014-02-11
Fix to start/stop the service.

### LINKED ISSUE #385: Quit does not Quit
Quitting the application (menu, quit) does not seem to actually quit the application. When you restart the application the "current stats" (Locations reported, Wi-Fi scanned, etc) values are not reset to zero. In addition, after "quitting", the ScannerService process continues to run in the background (eating CPU and battery).

If you force close the application or reboot the phone, the "current stats" are reset (and the application stops using resources). 

I would expect that the user requesting a quit would actually quit the application, unload it from memory, stop all background processes, etc.

If the intent of "quit" from the user interface point of view is to simply exit the UI, then there needs to be another method to actually stop all processing other than doing a force close or using some other task manager. Perhaps another menu option "quit all", "quit and unload", etc.

tamcap:
Fixes #385

### DIFF
--- a/src/org/mozilla/mozstumbler/ScannerService.java
+++ b/src/org/mozilla/mozstumbler/ScannerService.java
@@ -35,7 +35,7 @@ public final class ScannerService extends Service {
     private LooperThread        mLooper;
     private PendingIntent       mWakeIntent;
     private BroadcastReceiver   mBatteryLowReceiver;
-    private boolean             mIsBound = false;
+    private boolean             mIsBound;
 
     private final ScannerServiceInterface.Stub mBinder = new ScannerServiceInterface.Stub() {
         @Override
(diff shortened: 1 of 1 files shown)


tamcap:
This might be a more controversial solution...

tamcap:
Solution description:
- service is started with `startService()` as before
- if anything is bound, it will remain running
- if the service is running in the foreground (scanning actively) it will remain _on_ even if unbound
- if it becomes unbound (main application leaves the foreground) when not scanning, it will stop itself
- it will automagically restart itself next time it is needed
- I think the `stopSelf()` code in `stopScanning()` is redundant, but I would leave it there in case it stops scanning while unbound (not sure however, how it might happen)

Djfe:
the service will stop itself even while unbound if the battery level goes down to 14% ;)

tamcap:
Good catch!

cpeterso:
This looks good. Later, we might want to consider removing the service AIDL glue. I don't think it adds any value, but it complicates the code.

cpeterso:
LGTM. Thanks!

PR #403, opened 2014-02-12
Make MozStumbler beautiful

tamcap:
Adds icons to MozStumbler, removes Quit.


Djfe:
design ideas:

test Mozilla location service and leaderboard can have their own icon ;)

the map should be the default "state"

other Infos should be shown on top of it:
top left:
wifi icon followed by 3/105 (3 visible, 105 scanned)
the same for cells (a cell tower icon)
and a GPS icon as well (with fix/satellites)

you should be able to download the osm map/use downloaded maps of other osm navigation apps (osmand)
-> preload areas over wifi you want to stumble in

the detailed coverage map should be generated from a local database (#3)

there should be a button to switch between online and offline coverage map (online coverage map should get a maximum zoom level to prevent confusion (missing coverage tiles in high zoom levels), if you want to zoom in further you press the button (local coverage) 
-> minimal zoom level needed (over level 13 I think it was (where the online variant doesn't show anything), this way you don't need to combine local and online coverage tiles)

you should be able to see where cell towers are (marked on the map -> you can click on the point and even get additional infos about the tower)

your exact gps position should be shown as well and there should be a button to toggle automatic detection/lock the map to your current gps position

the start/stop scanning button should be displayed at the side or at the bottom of the screen

the number of locations should be displayed at the bottom next to an upload button which will show that the device is uploading (circle animation or progress bar will be pushed in from the bottom)

I hope you likey/these ideas ;) 
a lot of these suggestions have already been mentioned elsewhere (will try to refer to them later), but all of them could help to make mozstumbler more beautiful

Djfe:
@rdlt194070 pls make your own issue this has nothing to do with this one...

cpeterso:
Looks good! If you'd like to discuss other design questions, the dev-geolocation mailing list might be a better forum. Screenshots or ASCII art of new UI layouts would be helpful, too. :)

PR #407, opened 2014-02-13
Enable StrictMode only in debug mode. Call it early onCreate()

illarionov:


cpeterso:
LGTM. Thanks, @illarionov!

PR #410, opened 2014-02-16
Add Finland's VR train SSID to blacklist

aschmitz:
Their SSID is VR-junaverkko, see http://www.vr.fi/en/index/palvelut/palvelut_junissa/WirelessInternet.html


cpeterso:
Thanks, @aschmitz!

Issue #250, opened 2013-12-12
How to see if there are unsent informations and if yes force to report them ?


Valeryan24:
Hello,

MozStumbler is a fantastic idea to collect and provide a collaborative and open-source map of GSM cells, WiFi access points correlated to GPS locations for Firefox OS.

I have a question regarding the reports done :

I noticed that sometimes, there can be several minutes and lots of scanned positions (with a GPS fix) between 2 reports, if we stop the detection (because going into a building for example, or low battery), even with 3G data or wifi available, sometimes MozStumbler does not send the last reports immediately, we have to wait the next session.

Or if I wait many seconds / minutes after stopping detection, I see the number of reports sent increase 2 or 3 times more if many have to be uploaded, and normally number of reports approximately matches number of detections.

But if I quit MozStumbler before all the reports are done, and then restarts it, or if I put the phone on plane mode, or restart manually the phone, or if I have the problem described in issue 249 (crash), often when I restart MZ, all counts (locations, WiFi, reports) are reset back to 0 !

I know that all sent reports are well saved by Mozilla (I can see the Leaderboard points and map coverage increasing), but what happens to the reports not sent at the last close of MozStumbler ?

Is it sure that there are always sent on the next session ? Is there a way (with logs, exploring cache datas of MZ) to verify if there are not, on the several days of scanning, sometimes some locations and datas (GPS, wifi, cells...) that have been well scanned, written on the phone but never sent to Mozilla ?

And if yes is there a way to manually upload them, now ? This is not a matter of points (I don't care), but it can be frustrating when I spend lots of time, especially when walking on the town to catch all the small areas to get locations and wifi points on the most complete way, so having less locations than if I drive by car, so less reports sent per minute, and as I can not manually force all reports before stopping MZ, to think that sometimes all or part of what I've done is useless because never sent..

Thanks in advance.

cpeterso:
hi @Valeryan24: to reduce network traffic, MozStumbler saves measurements locally and uploads a "report" when it has at least 20 measurements. If you exit the app (or it crashes), the on-screen counter is reset to 0, but unsent reports should still be saved, ready for upload. Our UI's terminology is admittedly a little confusing because we show counts for "satellites", "locations", "access points", and "reports" and then "points" on the leaderboard which are all different things! <:\

The "Stop Scanning" button will try to force an immediate report upload, if the network is available.

Valeryan24:
Hello,

[...]

Issue #196, opened 2013-11-05
add button "send report"


CupIvan:
for force send current data, becose when there are no internet on phone and next time it will be available, stored data can't be transfered
I try to push stop/start button, but it often not start transfer mode

cpeterso:
When you press the "Stop Scanning" button, MozStumbler tries to upload your data. If the network is not available, it is supposed to try again later. Do you see your points on the leaderboard web page increasing?

CupIvan:
yes, but on last three updates I have only a few points, but later I score over 3000 for two times...
I don't know how it works, maybe programm transfer only several start coordinates, when internet connection was available
maybe for simplify show current score on the main window? and update it when data will be transfer

marco-c:
I may have the same problem. Should the "Reports sent" number be the same as "Locations scanned"? Sometimes I scan 2000+ locations and 2000+ APs, but when I get home only one report is sent.

cpeterso:
Each report uploaded to the server includes all the locations that have been scanned since that last report was uploaded.

CupIvan:
@cpeterso, sometimes, I went to home and count of locations not one, about ~100-200. maybe, if many data collected, it send several reports with some limit? I don't know all data transfered or not alltime and click button start/stop two-three times for reliability. I think, there is need text with count of transfered _records_ and total _records_ on the phone. If thats numbers are equal, that is good and I can close application. but if are different I will push "send" button.

marco-c:
Good to know. Maybe a visual indication that all the locations were sent would be useful.

Djfe:
there should be an indication that the app is uploading reports right now
so that you won't close the application too early

and another option would be helpful, too:
send reports automatically
which should send reports, if there have been 10 (just an example) or more new locations detected or if a certain amount of time went by
and you should be able to use/not use mobile networks for uploading data

Djfe:
My last comment seems to be described in this issue as well: #250 
The best indication that this issue prevents a lot of people from stumbling are the over 100 entries at the end of the leaderboard containing mostly 1 goes up to 5

People start scanning, think it's interesting, stop it again but only see one (to five) reports being send and they think all data has been delivered or they are wondering whether is still uploading data

If this issue and #250 will get fixed the number of those uploading pretty much nothing will get less in comparison to the ones enjoying the experience of contributing to this awesome project

Djfe:
can be closed now -> implemented with v0.13.0

cpeterso:
[...]

PR #413, opened 2014-02-16
Save on disk + sync adapter

illarionov:
1. Save data to be sent on disk (in the sqlite database)
2. Use the sync adapter for data uploading
3. Add "Upload observations" dialog (#250, #196)

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -18,6 +18,10 @@
     <uses-permission android:name="android.permission.READ_PHONE_STATE"/>
     <uses-permission android:name="android.permission.WAKE_LOCK"/>
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
+    <uses-permission android:name="android.permission.READ_SYNC_SETTINGS"/>
+    <uses-permission android:name="android.permission.READ_SYNC_STATS"/>
+    <uses-permission android:name="android.permission.WRITE_SYNC_SETTINGS"/>
+    <uses-permission android:name="android.permission.AUTHENTICATE_ACCOUNTS"/>
 
     <uses-feature android:name="android.hardware.location"/>
     <uses-feature android:name="android.hardware.location.gps"/>
@@ -59,6 +63,35 @@
             android:icon="@drawable/ic_launcher"
             android:label="@string/service_name">
         </service>
+
+        <service
+            android:name=".sync.AuthenticatorService"
+            android:exported="true">
+            <intent-filter>
+                <action android:name="android.accounts.AccountAuthenticator"/>
+            </intent-filter>
+            <meta-data
+                android:name="android.accounts.AccountAuthenticator"
+                android:resource="@xml/authenticator" />
+        </service>
+
+        <service
+            android:name=".sync.SyncService"
+            android:process=":.ScannerService"
+            android:exported="true" >
+            <intent-filter>
+                <action android:name="android.content.SyncAdapter"/>
+            </intent-filter>
+            <meta-data android:name="android.content.SyncAdapter"
+                android:resource="@xml/syncadapter" />
+        </service>
+
+        <provider
+            android:name="org.mozilla.mozstumbler.provider.Provider"
+            android:authorities="org.mozilla.mozstumbler.provider"
+            android:exported="false"
+            android:syncable="true"
+            android:process=":.ScannerService" />
 
     </application>
 
--- a/res/layout/dialog_upload_observations.xml
+++ b/res/layout/dialog_upload_observations.xml
@@ -1,0 +1,181 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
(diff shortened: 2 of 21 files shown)


cpeterso:
@illarionov, this looks great! This is a lot of code, so I have a few questions. You might want to answer my questions as big comments in the code itself. :)

Can you please describe the advantages of using the sync adapter? How often will the sync adapter upload and how does it interact with "Upload only WiFi"?

Can you please describe the schema of the database tables? AFAICT, we are storing a JSON list of multiple reports in each TABLE_REPORTS row. Can we just store one report or observation per database row? That seems a lot simpler. Batching up `RECORD_BATCH_SIZE` number of reports in `mReports` before flushing them to the Pref store has always been an extra layer of confusion and bugs. `RECORD_BATCH_SIZE` was originally added to throttle network uploads, but it later started throttling writes to the Pref store. I'd love if we could remove all the batching code and just write each report to the database. :)

Why do we need remember a RETRY_NUMBER for each report?

illarionov:
Synchronization adapter was chosen rather inspired by the video on the android developers channel: [Efficient Data Transfers](http://www.youtube.com/watch?v=5onKZcJyJwI) :)
It schedules data transfers in conjunction with transfers from other apps and theoretically this can reduce overall battery usage. And also you do not need to monitor the network connection.
The only drawback I found is that a our stub account is added to the list of accounts in the device settings, and it is not possible to hide it.

Synchronization service is notified by the provider every 10 queued reports (`SYNC_REPORTS_PERIOD`). When `RECORD_BATCH_SIZE=40`, it is about every 3-5 minutes.
Adapter executes sync not immediately.

On stop scanning (or when user clicks "upload" button) the immediate sync is requested.
If "Upload only WiFi" is activated and Wi-fi is not available, the `onPerformSync()` will fail with numIoExceptions=1 and sync adapter should re-schedule this sync.

The scheme all true. `TABLE_REPORTS` is a send queue. Each row is a single JSON request to the server with the measurements.
- `report`  - JSON payload.
- `observation_count`, `cell_count`, `wifi_count` - is all statics about this request.
- `retry` - retry number.

When the server returns a success or returns a fatal error, the row is removed from the database.
On the temporary error, this counter is incremented and query is re-executed on the next sync. But not more than `MAX_RETRY_COUNT` times. That is, the `retry` counter is used so as not to keep a record in the database forever.

`TABLE_STATS` is a simple key-value database with counters of data that is already sent: last upload time, number of observations/cells/wifi sent. I have not used SharedPreferences there because SharedPreferences does not work between 2 processes [on android 2.3](http://stackoverflow.com/a/12144418).

I have left all the batching logic to reduce the number of inserts, rows and overall load on the
database. Although now I also think that this optimization may be neglected in order to simplify the code. I will try to rewrite this part in a few days.

tamcap:
If I can add anything to the discussion: let's give users an option to keep the records in the db despite being sent - that was one of the original motivations behind local storage anyway (so that they can be plotted locally).

Also [edited]: would you be willing to put in lat / lon in the rows too? I mean outside the json string, as 2 separate fields (float/int).

illarionov:
I think it is better to make a separate table for this. Although, I do not really like the idea to implement a functionality of the GPS tracker in this application.

tamcap:
To be honest, neither do I...

illarionov:
Indeed, one of the advantages of the scheme 'observation per database row' will be coordinates in the separate fields. It even may be possible to merge two measurements on the same location (but at the different time) into the one.  Although I'm not sure about performance

tamcap:
I wonder if we could off-load some of the deduplication off ichnaea this way. @hannosch mentioned that some BSSIDs have thousands if not millions of observations already. There is no reason to report the same location 100 times with 10 networks that never changed.

cpeterso:
I like @illarionov's approach to log the observation reports because the server may, in the future, be able to make use of which APs are observed at the same time, not just the observed lat/lon of individual APs. While creating a table of individual APs will be interesting for some use cases, I don't think this PR should implement it.

I don't think having a MozStumbler sync provider listed in the system settings' list of sync accounts is a problem.

@illarionov: do you think I should merge this PR now or should I wait for you to remove the batching code? I'd like someone to remove batching code at some point, but I don't want this PR to bit-rot if there are merge conflicts from other PRs.

illarionov:
It would be better to merge this PR in the current state. I can not say for sure when my new revision will be ready.

tamcap:
Sooo... I cannot build this pr. The `:compileDebugAidl` breaks down by trying to AIDL-interpret the additional .java files (which is something I noticed earlier, and briefly mentioned on IRC). Any ideas how to fix it?

cpeterso:
@tamcap: I can build this PR, but I remember when you and I discussed this same `:compileDebugAidl` problem earlier. Are you build using the `./gradlew` script in the MozStumbler repo?

tamcap:
I made a fresh clone just before the commit a90dad0 merger, then `./gradlew build` --> ok; subsequent `git pull` got the commit a90dad0 and `./gradlew build` failed on `:compileDebugAidl`. 

Another fresh clone later, everything builds just fine... o_0

Djfe:
Thank you very much illarionov!
I really like those changes ;)

tamcap:
I was able to combat `:compileDebugAidl` problems by invoking `./gradlew clean` explicitly in the existing sources. Strange. But works.

PR #418, opened 2014-02-18
Update French translation with Geofence+Upload obs

pl6025:
Source for Geofencing in fr : https://fr.wikipedia.org/wiki/Gardiennage_virtuel


pl6025:
I hope that there is no problems with the Travis "control".

Is it normal that the is 2 or 3 identical strings with differents titles ? Wasn't it possible to re-use existing strings ?
There are strings that are not used for this version (activity strings), will we use it a day or not ? If not, it will be interesting to delete them (some bytes to gain).

cpeterso:
Thanks, @pl6025. Some strings may be identical because they are used in a different UI context (which might have different translations for some languages), but also because of user error, not seeing an existing string we could reuse. Feel free to file a bug pointing out the duplicate strings. <:)

PR #420, opened 2014-02-19
Update SSIDBlockList.java

martiensch:
Added two Dutch train operators and one bus operator with onboard wifi. Furthermore transport list was sorted alphabetically on SSID.


cpeterso:
Looks good! Thanks, @martiensch.

PR #429, opened 2014-02-19
Fixes to the geofencing.

### LINKED ISSUE #423: Can't disable geofencing
I've enabled geofencing, the button in the preferences now says:
"Geofence this location
Geofencing will be set at the next valid location obtained during scanning"
Now, I've changed my mind and I want to disable it, but I can't.

(Actually, I enabled geofencing just out of curiosity, it is still not clear to me what it is)

tamcap:
Changes the non-blocking UI behavior.
Blocks locations from being reported when `geofence here` is requested.
Fixes #423.


cpeterso:
Sounds good to me. Thanks!

PR #431, opened 2014-02-20
Add translation for observation strings

stripTM:
I doubted translate observation by:
1) Observación
2) Muestra
3) Anotación
I do not what is more natural in Spanish.


cpeterso:
Thanks, @stripTM! If someone has a better suggestion, they can take the initiative (like you have) to submit a PR. :)

PR #433, opened 2014-02-20
Remove batch processing logic from Reporter

illarionov:
Save each observation as a single row in the database ('Reports' table).

I'm also restored reporting of wifis and cells in one observation, since memory should not be a problem now.

### DIFF
--- a/src/org/mozilla/mozstumbler/DateTimeUtils.java
+++ b/src/org/mozilla/mozstumbler/DateTimeUtils.java
@@ -8,7 +8,7 @@ import java.text.SimpleDateFormat;
 import java.util.Locale;
 import java.util.TimeZone;
 
-final class DateTimeUtils {
+public final class DateTimeUtils {
     private static final DateFormat sLocaleFormat = DateFormat.getDateTimeInstance(DateFormat.SHORT, DateFormat.SHORT);
     private static final DateFormat sISO8601Format = new SimpleDateFormat("yyyy-MM-dd", Locale.US);
 
@@ -26,7 +26,7 @@ final class DateTimeUtils {
         return sISO8601Format.format(date);
     }
 
-    static String formatTime(long time) {
+    public static String formatTime(long time) {
         return formatDate(new Date(time));
     }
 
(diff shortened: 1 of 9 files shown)


Djfe:
Looks interesting ;)
Thx for the change

One question to how observations will be counted from now on:
4 wifi aps + 4 cell towers = 4 observation
Or:    "  "  = 8 observations
Are you counting them together or alone?

cpeterso:
LGTM! :)

tamcap:
I just got hit by a `NullPointerException`:

02-20 13:03:18.782    7718-7739/org.mozilla.mozstumbler:.ScannerService E/AndroidRuntime﹕ FATAL EXCEPTION: Thread-16347
    Process: org.mozilla.mozstumbler:.ScannerService, PID: 7718
    java.lang.NullPointerException: Attempt to invoke virtual method 'void android.location.Location.setTime(long)' on a null object reference
            at org.mozilla.mozstumbler.Reporter.reportCollectedLocation(Reporter.java:177)
            at org.mozilla.mozstumbler.Reporter.flush(Reporter.java:68)
            at org.mozilla.mozstumbler.ScannerService$1$2.run(ScannerService.java:90)
            at android.os.Handler.handleCallback(Handler.java:733)
            at android.os.Handler.dispatchMessage(Handler.java:95)
            at android.os.Looper.loop(Looper.java:136)
            at org.mozilla.mozstumbler.ScannerService$LooperThread.run(ScannerService.java:154)

tamcap:
Hmm. Now it's gone. Let's keep it in the back of our heads for now.

Djfe:
Maybe it happens once because the database got recreated (v2).
Then you couldn't reproduce it without a database v1.

Maybe you can reproduce it by reinstalling v0.13.0 and then updating again

cpeterso:
But GPSScanner.java looks like it would never send a `GPS_SCANNER_ARG_LOCATION` intent with a null `Location` (because `GPSScanner.onLocation()` would crash with its own NullPointerException first).

tamcap:
The problem is, it happens very randomly... I can't willingly reproduce it despite my best efforts.

tamcap:
My best bet is something, somewhere, calls a `new Reporter()` (which results in implicit setting of `mGpsPosition = null`)  while `reportCollectedLocation()` is running.

PR #435, opened 2014-02-20
Urgent fix.

tamcap:
Prevents a `NullPointerException` happening once in a while.


tamcap:
There seems to be something weird happening once in a while. There is a sanity check early on, but it seems `mGpsPosition` might change to null while `reportCollectedLocation()` is being executed. Probably needs further debugging.

tamcap:
Got it ;-)

Issue #77, opened 2013-08-22
Add "Only upload measurements over Wi-Fi" option


cpeterso:

cpeterso:
This feature would not be that difficult to implement. There are two parts:
1. Add a WiFi preference to PreferencesScreen.java and Prefs.java: https://github.com/mozilla/MozStumbler/blob/master/src/org/mozilla/mozstumbler/preferences/PreferencesScreen.java
2. Check the WiFi preference from NetworkUtils.isNetworkAvailable(): https://github.com/mozilla/MozStumbler/blob/master/src/org/mozilla/mozstumbler/NetworkUtils.java

Djfe:
Introduced with #378

gabrielbauman:
Please consider making this the default. I seem to be uploading nearly 100GB an hour when driving around town.

Djfe:
Wifi only is implemented already
-> look at the settings ;)

For the huge amount of data: look at #408

Djfe:
You are talking about 100mb, right?

gabrielbauman:
Yeah sorry... MB, not GB. I see that WiFi-Only is implemented - I just thought it should be the default.

cpeterso:
Fixed by PR #378.

PR #436, opened 2014-02-20
Enable "Upload only on Wi-Fi" by default

cpeterso:


cpeterso:
I thought there was an existing issue # for this feature, but I can't find it. :\

Djfe:
4th comment in #77 ;)

PR #452, opened 2014-02-24
Design changes

tamcap:
Better adherence to the AOSP / Android design guidelines:
- moved settings to the drop-down menu only
- reorganized action bar with the most common actions first

New icon for `Upload observations`.


cpeterso:
Sounds good!

PR #460, opened 2014-02-27
Show number of GPS satellites used / visible on main screen.

tamcap:
I remember it being requested at some point, however, not sure which issue number. In any case, it's a very simple fix.


cpeterso:
Sure, why not? I anticipate, however, that it will provoke questions or confusion about the number of satellites that MozStumbler "needs" to work. :)

Djfe:
"used" could really lead to confusion as mozstumbler would like to use all satellites
"available" sounds better in my opinion

tamcap:
I am not attached to any particular wording.

PR #461, opened 2014-02-27
Modify the date format to YYYY-mm-01.

tamcap:
Match the mozilla/ichnaea#120 changes as done by @crankycoder and send the date as `YYYY-mm-01` only (overwriting the day value).

### DIFF
--- a/src/org/mozilla/mozstumbler/DateTimeUtils.java
+++ b/src/org/mozilla/mozstumbler/DateTimeUtils.java
@@ -24,11 +24,14 @@ public final class DateTimeUtils {
 
     @SuppressLint("SimpleDateFormat")
     static String formatDate(Date date) {
-        Calendar c = Calendar.getInstance();
-        c.setTime(date);
-        c.set(Calendar.DAY_OF_MONTH,1);
-        date = c.getTime();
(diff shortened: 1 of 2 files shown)


Djfe:
What about the time of day?
-> seconds, minutes, hours

And does it make sense to replace this on the Client if the Server does that already?

tamcap:
We have removed time of day a while ago. This is a minor fix (in theory) improving privacy client-side and during transmission. I will not be very upset if @cpeterso decides it's redundant, but I like to keep client and server side "matched" to avoid some unexpected behavior down the road.

cpeterso:
I don't think the `formatDate()` method should be printing a different date/time value ("YYYY-MM-01") than the parameter value passed in ("YYYY-MM-DD"). Let the caller mangled the dates before it calls `formatDate()`. Perhaps the "YYYY-MM-DD" → "YYYY-MM-01" conversion should be  encapsulated in a helper method in DateTimeUtils.java?

tamcap:
Let me know what you think.

tamcap:
:+1:

cpeterso:
LGTM! Thanks, @tamcap!

PR #462, opened 2014-02-27
Block Adelaide Metro Public WiFi

### LINKED ISSUE #458: Exclude Adelaide Metro WiFi
Adelaide Metro operates a WiFi network on a small number of buses and trams.  As a result, these access points aren't reliable sources of location information.

The SSID of these access points is `PublicWiFi`, and their BSSIDs are in the `a8:54:b2` OUI.

tamcap:
First reported by @micolous. Closes #458.
Decided to use BSSID instead of SSID due to a relatively common name of the SSID: `PublicWiFi`.

### DIFF
--- a/src/org/mozilla/mozstumbler/BSSIDBlockList.java
+++ b/src/org/mozilla/mozstumbler/BSSIDBlockList.java
@@ -22,6 +22,9 @@ final class BSSIDBlockList {
         "7cc537",
         "88c663",
         "8c7712",
+
+         // Adelaide Metro WiFi Network OUIs:
+        "a854b2",
     };
 
     private BSSIDBlockList() {


cpeterso:
LGTM. I agree that the SSID name "PublicWiFi" is too generic to block.

micolous:
Thanks.

I've sent off a message to Adelaide Metro to get them to change their access point names to something hopefully a bit better, pointing out the issues above.

micolous:
I got a response back from Adelaide Metro about actually fixing their systems properly.  They outsource the operation of their network to an advertising firm, APN Outdoor:

> Dear Mark (sic)
> 
> Thank you for your Feedback and sorry in the delay to respond.
> 
> Our Service Provider is APN, who run the Free Wifi within the Adelaide CBD.
> 
> There contact number is 02 8569 3000, this is the enquires and complaint number, which is operational 24 hours a day.
> 
> Kind Regards
> 
> Customer Contact
> Adelaide Metro

I'm a bit disappointed that they don't have an email contact for them (and instead took a long time to respond with an interstate phone number).  I'll do some chasing up on this one to see if I can get contact details for one of their network engineers.

They don't actually run the [Adelaide CBD wireless network, CityLan](https://hotspot.internode.on.net/partners/citylan/) -- that this is just an error in their email.  Those are just fixed access points so don't cause problems for MozStumbler.

cpeterso:
Thanks for checking, @micolous. You don't need to spend too much time on this. Filtering out SSIDs that we know can move is a client-side optimization. Our server has code to (try to) detect other moving SSIDs, so we should be able to catch the "PublicWiFi" on the server. :)

PR #465, opened 2014-02-28
Set upload dialog background to "light"

### LINKED ISSUE #441: Upload observation dialog is barely readable
This is a screen shot from the 0.13.0 (same problem on the 0.14.0) for the report.
screenshot-1392832731318
It's really hard to read.

tamcap:
Fixes #441.
It's possible only certain APIs (<14?) are affected by this. But to be on the safe side, we can simply set it here.

 Alternative is the .xml file, but I wanted to preserve the commented out area in case we want to make it API specific at some point.


cpeterso:
Looks good!

PR #467, opened 2014-03-01
Fixes bug introduced in #400, adds button to notification.

tamcap:
PR #400 introduced a sneaky race-condition bug. It would try to destroy scanner service BEFORE all the resources from the "downstream" scanners could be turned off by the `Looper`. Memory leaks and fatal errors ensure. It would only trigger on a low battery intent when there was no client bound...
- new button to turn off scanning with main activity not running
- fix to the aforementioned bug


tamcap:
The bug fix benefits all, the notification button will only work in devices running ICS and above.

cpeterso:
Interesting. As you can probably guess, the low battery code patch is not tested often. :)

Issue #324, opened 2014-01-22
Only show the update notification once on app start


Djfe:
The update manager built into MozStumbler should only show the dialog box at the start of the app and not every time you switch to the menu/turn your device/open the map and get back.
There should be a button to press (check for updates or something like that) and a setting to disable the automatic search for updates.
Especially since version 0.9 doesn't seem to be bug free, it is a bit annoying.

cpeterso:
This bug is related to how our activity gets stopped and restarted when the device is rotated.

Djfe:
yeah but that is quite often if you press remind me later...
it should be hooked up to the app start at least (and the button solution from above (check for updates) would be nice as an addition)

Djfe:
similar to #280

Djfe:
this issue still happens, if the activity is reloaded (on device rotation or app reload (happens if you come back from the launcher or another app even though the service is running))
but it doesn't happen if you come back from the settings or the leaderboard anymore

crankycoder:
Fixed so long ago by @cascheberg.

Issue #280, opened 2014-01-09
enable GPS message


CupIvan:
when I start programm with GPS disable, there is a message on main screen.
then I disagree with it and click NO
if I go to settings or want view map and will back to main screen, there is that message again and again.

can you upgrade it, that programm will memorize my choice for current session?

or optional maybe show red text "GPS disable" and if all ok green text "GPS enable"
maybe instead of first line "GPS Satellites Visible"

Djfe:
@crankycoder I think the new issue #1151 fits better now since the UI changed so much.
The only thing missing now is an indicator for loosing the GPS connection.
The message described above has been removed already.

crankycoder:
No longer relevant.  GPS icons are black with a fix, red when no fix is available.

PR #469, opened 2014-03-03
Only show dialog to enable GPS on program launch

aaronaxvig:
Show the dialog prompting to enable GPS when the program is launched. Do
not show it when returning back to the main screen.

This is a fix for issue #324.

### DIFF
--- a/src/org/mozilla/mozstumbler/MainActivity.java
+++ b/src/org/mozilla/mozstumbler/MainActivity.java
@@ -47,6 +47,7 @@ public final class MainActivity extends FragmentActivity {
     private int                      mGpsSats;
     private boolean                  mNeedsUpdate = false;
     private boolean                  mGeofenceHere = false;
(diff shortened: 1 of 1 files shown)


cpeterso:
Thanks for the fix, @aaronaxvig!

Instead of adding a new `mHaveShownEnableGpsDialog` flag, can we move the call to `checkGps()` from `onStart()` to `onCreate()`? `onStart()` is called every time we return to the app's main activity, but `onCreate()` is only called once when the app is launched. I don't know if `onCreate()` is too early to test GPS or display the prompt.

Djfe:
This is rather a fix for #280 than for #324, but as cpeterso already stated this is a workaround which isn't as good as calling the alert on create/at the app start but it should work nevertheless

Would be nice to see #324 fixed in the same manner as #280 though ;)

aaronaxvig:
Yes, that is a better way to fix it.  I'll resubmit.

PR #470, opened 2014-03-04
Only show dialog to enable GPS on program launch

aaronaxvig:
Show the dialog prompting to enable GPS when the program is launched. Do
not show it when returning back to the main screen.

This is a fix for issue #280.


cpeterso:
Thanks, @aaronaxvig!

PR #474, opened 2014-03-07
Fixes to communicator package as suggested by @illarionov

tamcap:
Fixes to communicator package as suggested by @illarionov …
- Changing context scope.
- Moving closing cursor and submitter command into `finally{}`
- Deciding that all `IOException` can be temporary (reduces data discarding).
- Cleanup if gzip fails: the `AbstractCommunicator` has to clean up first.

### DIFF
--- a/src/org/mozilla/mozstumbler/sync/SyncAdapter.java
+++ b/src/org/mozilla/mozstumbler/sync/SyncAdapter.java
@@ -38,7 +38,6 @@ public class SyncAdapter extends AbstractThreadedSyncAdapter {
 
     private static final String LOGTAG = SyncAdapter.class.getName();
     private static final boolean DBG = BuildConfig.DEBUG;
-    private final Context mContext;
     private static final int REQUEST_BATCH_SIZE = 50;
     private static final int MAX_RETRY_COUNT = 3;
 
@@ -66,7 +65,6 @@ public class SyncAdapter extends AbstractThreadedSyncAdapter {
 
     public SyncAdapter(Context context, boolean autoInitialize) {
         super(context, autoInitialize);
-        mContext = context;
         mPrefs = new Prefs(context);
         mContentResolver = context.getContentResolver();
     }
@@ -82,7 +80,6 @@ public class SyncAdapter extends AbstractThreadedSyncAdapter {
             boolean autoInitialize,
             boolean allowParallelSyncs) {
         super(context, autoInitialize, allowParallelSyncs);
-        mContext = context;
         mPrefs = new Prefs(context);
         mContentResolver = context.getContentResolver();
     }
@@ -112,7 +109,7 @@ public class SyncAdapter extends AbstractThreadedSyncAdapter {
                 .build();
         queueMinId = 0;
         queueMaxId = getMaxId();
-        Submitter submitter = new Submitter(mContext);
+        Submitter submitter = new Submitter(getContext());
         for (; queueMinId < queueMaxId; ) {
             BatchRequestStats batch = null;
             Cursor cursor = mContentResolver.query(uri, null,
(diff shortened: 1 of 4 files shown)


src/org/mozilla/mozstumbler/sync/SyncAdapter.java
```diff
@@ -135,14 +136,13 @@ private void uploadReports(boolean ignoreNetworkStatus, SyncResult syncResult) {
             } else {
                 syncResult.stats.numIoExceptions += 1;
                 if (submitter.isErrorTemporary()) {
-                   increaseRetryCounter(cursor, syncResult);
+                    increaseRetryCounter(cursor, syncResult);
```
illarionov: Thanks! But increaseRetryCounter() requires cursor to be open.

src/org/mozilla/mozstumbler/sync/SyncAdapter.java
```diff
@@ -112,7 +109,7 @@ private void uploadReports(boolean ignoreNetworkStatus, SyncResult syncResult) {
                 .build();
         queueMinId = 0;
         queueMaxId = getMaxId();
-        Submitter submitter = new Submitter(mContext);
+        Submitter submitter = new Submitter(getContext());
```
illarionov, line 112: Fount yet another problem here.  Instances of URLConnection are not reusable:
java.lang.IllegalStateException: Already connected
at java.net.HttpURLConnection.setFixedLengthStreamingMode(HttpURLConnection.java:754)
at libcore.net.http.HttpsURLConnectionImpl.setFixedLengthStreamingMode(HttpsURLConnectionImpl.java:366)
at org.mozilla.mozstumbler.communicator.AbstractCommunicator.sendData(AbstractCommunicator.java:90)
at org.mozilla.mozstumbler.communicator.AbstractCommunicator.send(AbstractCommunicator.java:107)
at org.mozilla.mozstumbler.communicator.Submitter.cleanSend(Submitter.java:39)
at org.mozilla.mozstumbler.sync.SyncAdapter.uploadReports(SyncAdapter.java:145)
at org.mozilla.mozstumbler.sync.SyncAdapter.onPerformSync(SyncAdapter.java:98)
at android.content.AbstractThreadedSyncAdapter$SyncThread.run(AbstractThreadedSyncAdapter.java:254)

tamcap:
That's strange. `Submitter.close()` should nullify the URLConnection.

illarionov:
That's right. I'll check again, where it can occures

illarionov:
Looks like this exception occured when falling back to plain-text

```
void send(byte[] data) throws IOException {
        setHeaders();
        try {
            sendData(zipData(data));
        } catch (IOException e) {
            Log.e(LOGTAG, "Couldn't compress and send data, falling back to plain-text: ", e);
            sendData(data);
        }
    }
```

tamcap:
OK, this has to be addressed separately.

tamcap:
I might bring submitter back out (basically reverse the cc8a021) as I am guessing this is what was causing the problems, not reuse inside the `for` loop (which is nullifying the `UrlConnection` as it should)

src/org/mozilla/mozstumbler/sync/SyncAdapter.java
tamcap, line 112: This should not be happening anymore.

illarionov:
Yes, sorry for my incorrect assumption.
Thanks for this changes! 
I will test it a little more today.

tamcap:
No worries - I clearly missed most of those too...

tamcap:
An interesting question. While I am not the author of the original concept, I always imagined the code was in place to prevent long-term local database contamination with corrupted values. If for some reason our database contains non-jsonable data, we have to have a mechanism to purge it at some point so it does not remain in it forever. How often does this happen? Should we discard this check whatsoever and assume any error returned by the `Sender` to the `SyncAdapter` is temporary? Not sure. :8ball:

illarionov:
Perhaps I had a bad choice of words. Actually I mean to merge ConnectTimeoutException and IOException catch blocks but to not touch HttpErrorException catch block.
I. e. assume that all errors are temporary except for HttpErrorException.isTemporary().

But I also like the variant that is implemented by @tamcap in this request. Actually, we can remove the division of errors into temporary and permament, and always repeat sending on failure. `deleteObservations()` may be removed in this case.
The logic to remove records from the database is in the `increaseRetryCounter()` now (that is very bad, I know).

tamcap:
I added the commit implementing @illarionov's suggestion to ignore all errors while sending. `increaseRetryCounter()` can handle all the deletions. @cpeterso - feel free to make a call which solution you like more, I can always drop the fcf98c3. However, I would like to ask for at least 488d2bc to be included into the tree before the next release, as it fixes problems I accidentally introduced earlier.

cpeterso:
LGTM! I think both changes make sense.

@tamcap: thanks for your contribution, as always!
@illarionov: thanks for your suggestions and reviews. :)

PR #480, opened 2014-03-09
New position of Start Scanning button in landscape mode

### LINKED ISSUE #479: Landscape mode -> no scroll
I cannot scroll in landscape mode and therefore not access the scan button then

We need a scroll bar or move the gui elements so that they fit on the screen as a whole (by placing the button the right for example)

My device's display: 240 dpi 480x860 (or something like that not sure about that but it is around 4xx times 8xx)

tamcap:
Closes #479.


cpeterso:
Good idea. Thanks!

PR #481, opened 2014-03-10
Display the number of unique cells [#320]

illarionov:


cpeterso:
Looks good. Thanks, @illarionov!

PR #482, opened 2014-03-10
Execute all reporter code in LooperThread

illarionov:
This should fix the arbitrary crashes on start/stop scanning


cpeterso:
LGTM. Thanks!

Issue #477, opened 2014-03-09
No scans for Wifi APs (enabling wifi after scan start)


Djfe:
MozStumbler doesn't scan for Wifi Aps if you enable it after it started to scan

It needs to check for a working Wifi connection each scanning interval instead of only once at the button press
Or we need an intent that gets executed each time the state of the wifi connection changes.

Djfe:
Has somebody tried to reproduce this bug, yet?
Wifi off->start scanning->wifi on->no scans for Wifi aps

guerda:
Yes, I did. I'm currently using the latest version (0.15.0) and did the following steps:
1. Turn off Wifi
2. Start MozStumbler ( http://i.imgur.com/RCXW8aS.png , no Wifi connected)
3. Start scanning
3. Turn on Wifi (I even connect to one in my example, http://i.imgur.com/wQkAh0N.png)
4. No Wifi networks are detected
5. I stop and start scanning ( http://i.imgur.com/L2EZ0XV.png )
6. All Wifi networks are detected

Djfe:
Thx ;)

cpeterso:
Fixed by PR #489.

guerda:
I can verify that this bug is fixed. 
1. Turn wifi on
2. Start scanning
3. Turn wifi off (visible wifi networks 0)
4. Turn wifi on (visible wifi networks rises)

Thank you for the work!

Djfe:
First the newest version is 0.16.0. Second you would need to compile it for your own as the fix has been implemented after the last update...

guerda:
I tested it with 0.16.0 today, it works.

Djfe:
Oh wait actually the fix was included in 0.16.0 so my fault ;)

PR #489, opened 2014-03-11
Resume wi-fi scan after Wi-fi is turned on [#477]

illarionov:


cpeterso:
Looks great. Thanks, @illarionov!

Issue #32, opened 2013-07-24
Investigate "Wi-Fi scan-only mode" (REQUEST_SCAN_ALWAYS_AVAILABLE) in JB 4.3


cpeterso:
REQUEST_SCAN_ALWAYS_AVAILABLE is a new API exposed in JB 4.3:

_Wi-Fi scan-only mode is a new platform optimization that lets users keep Wi-Fi scan on without connecting to a Wi-Fi network, to improve location accuracy while conserving battery. Apps that depend on Wi-Fi for location services can now ask users to enable scan-only mode from Wi-Fi advanced settings. Wi-Fi scan-only mode is not dependent on device hardware and is available as part of the Android 4.3 platform._

https://developer.android.com/reference/android/net/wifi/WifiManager.html#ACTION_REQUEST_SCAN_ALWAYS_AVAILABLE

Djfe:
Has somebody looked at this already?
Should be fairly easy to implement and as it keeps battery usage low...

Does this influence the string shown for each app that uses the device location in Kitkat location settings? (high/low battery usage)
Or does Android only check whether an app will prefer the Google location service over the exact GPS position?

PR #493, opened 2014-03-13
Add "Scan AP's even when Wi-FI is off" [#32]

illarionov:


cpeterso:
Sorry for the long delay, @illarionov! Your PR looks good.

PR #501, opened 2014-03-13
Add cell data to the lookup requests in MapActivity

### LINKED ISSUE #354: "Test Mozilla Location Service" Is problematic
1) When there's no GPS fix, it doesn't do anything and wait for a fix. The location service should use wifi/cell tower data to try to find the location even if GPS location can't be determined.

2) It never finds my location, even in areas that I previously scanned with the apps, even more than 10 times. Once it gets a GPS fix it says it couldn't locate me.

tamcap:
Partially-Fixes #354.


tamcap:
Any comments?

cpeterso:
Looks good! Sorry for the delay, @tamcap. Thanks for reminding me. <:)

PR #505, opened 2014-03-15
quit string deleted

Djfe:
as there is not quit button anymore we don't need the string for it any longer


cpeterso:
@Djfe: the Travis CI build failed because of the following lint error. If you removed a string from the default locale file, you need to remove it from all other locale files, too.

```
MozStumbler/res/values-bn-rBD/strings.xml:7: Error: "action_exit" is translated here but not found in default locale [ExtraTranslation]
```

Djfe:
New pull request: #509

PR #523, opened 2014-03-28
Update French strings (Wifi always/disabled + sat)

pl6025:


cpeterso:
Looks good. Thanks, @pl6025!

PR #527, opened 2014-03-31
Update strings.xml

tamcap:
Typo.


tamcap:
APs might be confusing to some anyway. Would you guys like to spell it out instead?

Djfe:
Theoretically APs is wrong in any case as mozstumbler records logical networks so there might be more than one per ap, according to Hannosch
This is the same case for cells.

And if you want to stick to ap/access point:
Right now the German translation writes it out anyways as there is no shortform for "Zugriffspunkt" .

cpeterso:
LGTM. Thanks.

PR #528, opened 2014-04-03
Update SSIDBlockList.java

martiensch:
Rename of NS train wifi (source: http://tweakers.net/nieuws/95116/ns-neemt-wifi-in-de-trein-over-van-t-mobile.html)


cpeterso:
Looks good. Thanks!

PR #532, opened 2014-04-08
General improvements to MapActivity

tamcap:
- progress bar eye candy
- zoom level depends on the accuracy value reported


cpeterso:
Looks good. Thanks!

PR #543, opened 2014-04-19
Add more on-vehicle SSIDs to the block list

Rondom:
Hello,
I added SSIDs to the block list for:
- Barcelonabusturistic.cat touristic buses
- Flygbussarna.se buses
- National Express coaches
- NSB.no Norwegian state railways
- Oxfordtube.com buses
- SJ.se Swedish state railways

Right now there is only code in place for doing prefix and suffix matching. Given for example the very short length of the Swedish State Railways' on-train SSID "SJ" one can consider a more flexible matching. I know, there will always be some false-negatives, but if others agree, I can add more flexible matching.

### DIFF
--- a/src/org/mozilla/mozstumbler/SSIDBlockList.java
+++ b/src/org/mozilla/mozstumbler/SSIDBlockList.java
@@ -40,6 +40,7 @@ final class SSIDBlockList {
         "Amtrak_",
         "amtrak_",
         "arriva", //Arriva Nederland on-train Wifi (Netherlands)
+        "Barcelona Bus Turistic ", // Barcelona tourisitic buses http://barcelonabusturistic.cat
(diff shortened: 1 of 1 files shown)


Djfe:
I trink SJ is to general but the rest should be ok.

cpeterso:
@Rondom: Thanks for the SSIDs! I agree with @Djfe. The name "SJ" is too generic to block all networks with that SSID. Also, there was a git merge conflict because another pull request added some SSIDs to the same file. Can you please resubmit your pull request using the latest MozStumbler code?

Rondom:
Rebased on top of master and removed SJ due to being too generic. NSB has been added by someone else since.

cpeterso:
Looks good. Thanks, @Rondom!

Issue #555, opened 2014-04-26
Disable upload only on Wi-Fi by default


vinitraje:
In India, some of my friends and I are working to spread word about Mozilla's location service and trying to get MozStumbler installed on maximum phones. We've spread the message via Facebook, WhatsApp and other media.

But the problem is more than 90% of Android users here use mobile network for internet. Even if most of our contacted peoples are scanning with their phone, they don't know about the upload only on Wi-Fi option and they are not able to upload any data so far.

Upload observation on Wi-Fi by default option confuses people with poor technical knowledge of Android phone. It is quite frustrating job to explain the working of MozStumbler to everyone. Instead we're just asking them to install the app, scan their area. It would be better if the option is disabled by default.

cpeterso:
@vinitraje: that's a good point. We made "upload only on Wi-Fi" the default because many users have a mobile data plan that either have a data limit or charges per megabyte. We didn't want to hit someone's data limit or cost them a lot of money. What do you think? What are the common mobile data plans that Android users in India use?

cpeterso:
Maybe we can make a smart default. If the phone's Wi-Fi is disabled when MozStumbler is run for the very first time, then "upload only on Wi-Fi" could default to _false_. If Wi-Fi is enabled (but not necessarily connected to a Wi-Fi network), then default to _true_.

tamcap:
I like it how it is. We could make a first run splash screen saying "uploading on WiFi only by default, go to settings to change" or something like that.

But then, I'm a big fan of the KISS principle.

vinitraje:
In India common plans start from 1GB per month and aren't that expensive.  As I said earlier WiFi coverage is limited to a small proportion of users.

And I have observed that if we upload data after scanning is complete instead of sending after specific interval then data use is less.

vinitraje:
Splash screen idea is good too. Let the users choose either WiFi or mobile network on the splash screen itself.

@cpeterso the top operators in country have data plans in between 139-155 INR for 1GB. And they have unlimited plans too.

vinitraje:
@tamcap if you look over the statistics of Android users in India there are millions of Android users. 

And if you check the WiFi availability it is limited to 100 or more towns only. Others rely on mobile network. We just want to use full potential of Android users in India.

garvankeeley:
@vinitraje very interesting points about the cell vs. WiFi network in India. 
@tamcap I agree with providing a choice on first-run with a splash screen.

[...]

PR #556, opened 2014-04-26
Disable upload only on Wi-Fi option by default

vinitraje:
This will fix the issue #555


cpeterso:
@vinitraje: based on our conversation in issue #555 about a splashscreen to ask users for their upload preferences, I am going to close this pull request. Thanks for starting this conversation! :)

PR #564, opened 2014-04-30
Adding first bus wifi (uk) to ssid block list

miawgogo:
Added missing bus transport


garvankeeley:
Thanks for adding this ioangogo!

PR #565, opened 2014-04-30
Use LocalBroadcastManager for wifi,cell, gps broadcast

garvankeeley:
Switch to LocalBroadcastManager for wifi,cell, gps broadcast in the service module (bug 1002610)

PR #567, opened 2014-05-01
Switching to single process

garvankeeley:
Issue #565, can’t use LocalBroadcast across the processes. One process
should work fine, I think this was a legacy design decision. Changed that, and used LocalBroadcastManager as minimally as possible to reduce changes to the code.


cpeterso:
Looks good to me! :+1:

PR #571, opened 2014-05-04
Changed to the right form

yarons:
As long as case sensitivity matters this should be fine. Sorry for the hassle.


cpeterso:
Thanks for catching this. :)

PR #572, opened 2014-05-04
Always use LocalBroadcastManager to send and receive MESSAGE_TOPIC intents

illarionov:


cpeterso:
@garvankeeley: can you take a look at @illarionov's LocalBroadcastManager changes? I'm not sure how it interacts with your recent LocalBroadcastManager changes.

eggpi:
Since I already had made this refactor myself, I opened PR #575. @illarionov, can you please rebase your patch on top of that once it gets merged? Thanks!

illarionov:
@garvankeeley, `onReceive()` of a `BroadcastReceiver` that is registered with `LocalBroadcastManager` is always executed in the main looper thread of the process. 
The other scan code is running in it's own dedicated looper thread - [mLooper](https://github.com/mozilla/MozStumbler/blob/master/src/org/mozilla/mozstumbler/ScannerService.java#L189) for now.
Such execution of code in different threads has often led to race conditions and strange behavior. For example, stopping the scanning service during the execution of the `reportCollectedLocation()` has often lead to `NullPointerException`. 
Introduction of the mHandler was necessary to ensure that the code is running in the dedicated thread of the scanner, so the `shutdown()` have not to be running at the same time as `reportCollectedLocation()`.

@guilherme-pg, good, I will rebase.  This definitely simplify the code and solve many multithreading problems. Although not sure if this a good idea to execute scan in the main looper thread.

garvankeeley:
@illarionov: thanks for the detailed explanation! I pulled your fix and ran it through some testing (TODO: add a smoke test so we are all on the same page when we say we manually checked out the app :).

eggpi:
Hmm, I thought we were going merge PR #575 and then rebase this one on top of it. Not a big problem though, shouldn't be too hard to rebase #575 now.

garvankeeley:
Sorry, I had assumed a merge of Reporter.java would be straightforward.

PR #575, opened 2014-05-06
Old code removal

eggpi:


eggpi:
@hannosch may be interested in at least skimming through the last patch, as it is the result of a conversation we had on IRC a few weeks ago.

hannosch:
@guilherme-pg f+ looks ok to me

cpeterso:
Looks good!

PR #576, opened 2014-05-06
Simplify formatting of cell data

eggpi:
This is more of a tentative patch, as it breaks the dual-SIM case. Thanks @illarionov for pointing it out!.

Do we want to support dual-SIM devices (I think we do)? Even if we do, we should think about how to deal with that case: do we want to do separate submissions for each type of cell and then one for wifi, or maybe bundle wifi along with one of the cell submissions, and does it even matter? And what should the "radio" attribute be in that case? cc @hannosch

### DIFF
--- a/src/org/mozilla/mozstumbler/Reporter.java
+++ b/src/org/mozilla/mozstumbler/Reporter.java
@@ -9,6 +9,7 @@ import android.content.IntentFilter;
 import android.location.Location;
 import android.net.wifi.ScanResult;
 import android.support.v4.content.LocalBroadcastManager;
+import android.telephony.TelephonyManager;
 import android.util.Log;
 
 import org.mozilla.mozstumbler.cellscanner.CellInfo;
@@ -47,6 +48,7 @@ final class Reporter extends BroadcastReceiver {
 
     private final Context       mContext;
     private final ContentResolver mContentResolver;
+    private final int             mPhoneType;
 
     private Location            mGpsPosition;
     private final Map<String, ScanResult> mWifiData = new HashMap<String, ScanResult>();
@@ -58,6 +60,8 @@ final class Reporter extends BroadcastReceiver {
         resetData();
         LocalBroadcastManager.getInstance(mContext).registerReceiver(this,
                 new IntentFilter(ScannerService.MESSAGE_TOPIC));
+        TelephonyManager tm = (TelephonyManager) mContext.getSystemService(Context.TELEPHONY_SERVICE);
+        mPhoneType = tm.getPhoneType();
     }
 
     private void resetData() {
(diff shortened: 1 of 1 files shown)


cpeterso:
@guilherme-pg: I just merged your PR #575. Does this PR conflict with #575?

eggpi:
Just rebased so it no longer does, thanks. But let's not merge this until we figure out whether we actually want it, please :)

hannosch:
@guilherme-pg uhm, I thought even dual or triple-sim phones where either all gsm or all cdma. Turns out there are cross gsm/cdma phones. But the way I understood multi-sim devices, only one sim card is ever active at a time. So any location report (lat/lon + cell, wifi data) should be consistent in itself, that is only have gsm data and wifi or only cdma data and wifi.

If that is true, than the original behavior makes sense. We still need to group submission by phone/sim card type, but it should be rare to have data for both groups at the same time. So I wouldn't worry about trying to send wifi data for only one of them.

As for the top-level radio type, we would end up having two different groups. One where it's set to gsm and the cell entries are all for gsm/umts/lte cells. And one group where it's set to cdma and the cell entries are all about cdma cells.

To the server this will basically look like two different phones sending data.

Djfe:
@hannosch for the future
It might not be possible right now to use two connections at the same time
But it will be in the future:
Buzz word: multipath tcp (mptcp)
You can surf over wifi and mobile network at the same time then, even 2 cells at a time should be possible
You will receive a connection speed of all devices together (maybe not the latency though)

However I don't know whether there are countries/regions where you have to use cdma and GSM at the same time but dual SIM devices for one standard will probably use both connections at the same time in the future to increase the speed.

We probably need someone with such a phone here to test whether it is possible to scan with two Sim systems at the time (dual SIM in general). Especially a device with cdma and GSM would be good as Google separates cdma and GSM as well and all implementation on the market could be different -> probably no general standard out there to access the telephony manager in the same way to scan cdma and GSM at the same time

eggpi:
@illarionov , you have a dual SIM phone, right? Can you please confirm that you do get cell scans of different radio types while the stumbler is running, and which radio type is returned in the TelephonyManager API, so we know whether we need to group them or not? You've probably run this test before, just wanted to double-check before closing this pull request.

illarionov:
@guilherme-pg I have several Dual SIM Dual Standby phones on the MTK chipset. The scan on both SIM cards at the same time is possible (using undocumented API).  There is a separate instance of the TelephonyManager for each of the SIM card. The default TelephonyManager API returns radio phone type of the default SIM card.
But all my phones are GSM only. Theoretically, CDMA+GSM dual SIM phones are exists. For example **HTC One dual SIM 802D** or **Lenovo A600e**.  I made a handling of this cases. 
But while there is support of the Dual-SIM in the main repository, this handling can be removed in order to simplify the code as it makes no sense to maintain it.

P.S. I have implemented support for my devices ([geminiScanner](https://github.com/illarionov/MozStumbler/commit/0ea14a2d771308cfca098e82b0723bca4152885f)). It works fine for me, all the SIM cards are participating in the scanning.  But this only works on the MTK-based phones and the undocumented API is used. That is why I'm not sure that this will be useful in the main repository. And I have to roll back this pull request in my fork.

eggpi:
Thanks, @illarionov .

So if I understand it correctly, it should be fine to merge this PR for now, and leave the CDMA+GSM case for later, if it ever comes up? @cpeterso , if you agree, can you please review? Thanks!

cpeterso:
LGTM. Thanks!

PR #577, opened 2014-05-08
 Code Cleanup, and client/service split

garvankeeley:


garvankeeley:
I noticed the manifest still has a ScannerService entry. Harmless, but needs cleaning out. (ScannerService is now the StumblerService)

eggpi:
You probably meant to open this pull request against my repository rather than this one, right?

garvankeeley:
I thought you did that, this would be getting comical if I wasn’t so annoyed with myself, closing :)

PR #585, opened 2014-05-11
Disable hardware acceleration to reduce memory usage

illarionov:
This significantly reduces initial memory usage. From 29MB to 5MB on Nexus 5

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -33,7 +33,8 @@
         android:icon="@drawable/ic_launcher"
         android:label="@string/app_name"
         android:theme="@style/AppTheme"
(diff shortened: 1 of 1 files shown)


garvankeeley:
I have always just had it on, so I had to read into this. The android docs and this stack overflow provide some background:
http://stackoverflow.com/questions/16432565/android-hardware-acceleration-to-use-or-not-to-use

Unless the application becomes more complex in its drawing, there is no need for it. I tried to find information as to the battery consumption difference, as hardware graphics rendering (GPU) is theoretically more efficient, and  would use less power than cpu-bound graphics rendering. I couldn't find any mention of this, I assume it is inconsequential.

PR #587, opened 2014-05-12
Added SSIDs for Swiss PostAuto and German coach companies

JensErat:


cpeterso:
Looks good. Thanks, @JensErat!

Issue #408, opened 2014-02-13
Data usage on 0.12.1


leogaggl:
There seems to have been another regression on the latest update (which I just updated yesterday).

The app chewed through 1GB of mobile data within 3-4 hours. I have used this for months on mobile data and newt had an issue with data usage. 

Since this happened while overseas - there goes the data allowance. Not happy ...

tamcap:
I strongly recommend using the use wi-fi only option when concerned with data usage. However, 1 GB in 3 hours does sound excessive...

leogaggl:
Thanks. Yes - I have already done that and blocked the app from mobile data.

However this has never been a problem over months. And the whole 1GB was used by the stumbler app. So it seems to me to be an issue introduced in the latest upgrade which co-incided with the problem.

Excessive is a nice way of putting it - have mumbled some stronger terms ... ;-)

Djfe:
Have you used it in this time?
Have you opened the map
or have you only uploaded locations?

or did it always run in the background?

cpeterso:
I wonder if this bug is related to the problem reported with the "Reports Sent" counter was surprisingly greater than "Locations Scanned". :(

leogaggl:
&gt;Have you used it in this time?
No.

&gt;Have you opened the map
No.

&gt;or have you only uploaded locations?
Not many - mainly stationary in hotel

&gt; or did it always run in the background?
Yes.

tamcap:
I can definitely say I consistently see approximately 2x reports sent vs location scanned. I was pretty busy and didn't have a chance to adb it yet.

illarionov:
This is because each location is sent as 2 reports now: wifis and cells.

cpeterso:
That must an unintended side effect from my reporter refactoring. :(

stripTM:
0.5GB consumed me in two days, I am now working only on wifi
https://twitter.com/stripTM/status/434789779864301568/photo/1

rfugger:
I'm having the same issue. I've been using the app as normal, but my data usage in a few days after updating to 0.12.1 has been 375 MB. I never noticed any significant data usage previously. This is a definite problem.

gabrielbauman:
I was bitten by this recently - about 100MB an hour consumed over several hours of driving around.

Perhaps making wifi-only uploads the default would be a good idea? Data plans here in Canada are mostly in the 600MB - 1GB monthly range, so even 50MB or 25MB per hour is far too much.

[...]

PR #592, opened 2014-05-16
URGENT bugfix

### LINKED ISSUE #591: 0.20 generates big data file and crashes when trying to upload data
I had a long trip and I took collect more data.  It worked relatively fine for the first ~2000 data reports. 

Disclaimer: My Internet connection is quite bad at this moment, so I have not  been able to search properly. So, please bear with me.

At some point I got a notification of low internal space.  I had left ~20 MB.  I cleared data from some applications (like Firefox), I made ~60 MB.  MozStumbler had around 16MB, considering the history of thousands reports it should have been more than enough.

After I while I noticed that MozStumbler stopped silently to collect data. I re-started every time.  That is, until I noticed I had ~1900 pending reports and I tried to upload it manually and it crashed.

So, I enabled to upload only on WiFi and MozStumbler started to collect data continuously again. Until again I was running out of space.  This was is bad sympton, but I uninstalled several other applications to make space and continue tracking.

Long story short: MozStumbler whent from 16MB to 106MB of data in less than one day. I ran out of space and I don't know how to extract manually to inspect it.  I have 4700 observations that can't be uploaded because MozStumbler crashes.

I wish MozStumbler could store the data in the SD, like I can do with other applications.

For context: I have a Nexus 1 with Cynamodgen. I used the file manager to locate a file, with no luck.  And no luck by grepping in the source code.

In short, there are 3 problems:
- Crash when uploading a file (maybe the data storage got corrupted at some point)
- Big increase of the data file between 0.19 and 0.20 (I hardly believe it could have collected 3 times (or more) data than I had done in months)
- No possibility to store the data in an external storage where to back it up (maybe I am missing something).

tamcap:
We never cleaned up the mCellData and mWifiData  after reading from them - making them become HUGE and exponentially increasing storage necessary by MozStumbler by having thousands if not millions of duplicates.

I think this should do the trick. Fixes #591. Please merge & release ASAP.

### DIFF
--- a/src/org/mozilla/mozstumbler/Reporter.java
+++ b/src/org/mozilla/mozstumbler/Reporter.java
@@ -148,6 +148,7 @@ final class Reporter extends BroadcastReceiver {
         Collection<CellInfo> cells = mCellData.values();
         Collection<ScanResult> wifis = mWifiData.values();
 
+
         ContentValues values = new ContentValues(10);
         values.put(Reports.TIME, mGpsPosition.getTime());
         values.put(Reports.LAT, Math.floor(mGpsPosition.getLatitude() * 1.0E6) / 1.0E6);
@@ -189,7 +190,8 @@ final class Reporter extends BroadcastReceiver {
                 Log.w(LOGTAG, "JSON exception while serializing wifi data: " + exc);
             }
         }
-
+        mCellData.clear();
+        mWifiData.clear();
         values.put(Reports.WIFI, wifiJSON.toString());
         values.put(Reports.WIFI_COUNT, wifiJSON.length());
 


src/org/mozilla/mozstumbler/Reporter.java
```diff
@@ -146,7 +146,9 @@ private void reportCollectedLocation() {
         }
 
         Collection<CellInfo> cells = mCellData.values();
+        mCellData.clear();
         Collection<ScanResult> wifis = mWifiData.values();
+        mWifiData.clear();
```
illarionov: I think this should be somewhere later. The Collection returned values() is backed by the map so changes to the map are reflected in the collection.

tamcap: ugh, i just assumed it was a simple copy

src/org/mozilla/mozstumbler/Reporter.java
```diff
@@ -189,7 +190,8 @@ private void reportCollectedLocation() {
                 Log.w(LOGTAG, "JSON exception while serializing wifi data: " + exc);
             }
         }
-
+        mCellData.clear();
```
tamcap, line 193: @illarionov - that should be late enough :anguished:

eggpi:
I don't think I'm qualified to do a release, but the code looks good to me. Can you please squash your changes into a single commit?

tamcap:
:+1:

Djfe:
Does this fix #408 as well?
If the issue was older than one version then that is likely.

I guess nobody looked further into it since the gzipping reduced the upload size greatly (which makes sence since the data has been duplicated...)
Nobody reported issues about the internal data usage after gzipping was introduced for the upload

tamcap:
I don't think so. High data usage was more of an anomaly instead of a
regular occurrence.

Djfe:
Well than the old issue could be closed, I would say.

garvankeeley:
I started by pulling the current release. Just catching up on this issue right now.

PR #611, opened 2014-05-29
Remove old preferences keys on upgrade

illarionov:
This may fix a slow start of the application


garvankeeley:
Yikes, that would certainly slow down the startup :), thanks for the commit.

PR #613, opened 2014-06-01
Determine whether the cell location is updated when the screen is off

illarionov:
I found that some phones do not update the current сell location when display is off. The `mTelephonyManager.getCellLocation()` returns the same value until the screen is turned on, though, in fact, the current cell changes. 
If you go with such phone at a high speed for a long time, the phone incorrectly reports signal from the cell tower over the long distance.
My solution is to watch for changes when the screen is off and do not take into account cell location data until it is at-least one change.
Related bug: https://code.google.com/p/android/issues/detail?id=10931

### DIFF
--- a/src/org/mozilla/mozstumbler/cellscanner/DefaultCellScanner.java
+++ b/src/org/mozilla/mozstumbler/cellscanner/DefaultCellScanner.java
@@ -33,6 +33,7 @@ public class DefaultCellScanner implements CellScanner.CellScannerImpl {
 
     private static final GetAllCellInfoScannerImpl sGetAllInfoCellScanner;
 
+    private final ScreenMonitor mScreenMonitor;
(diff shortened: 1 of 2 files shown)


garvankeeley:
Good find! What device and version are you testing this on? I'll try it on mine, Nexus 5 + 4.4.2.
I wonder if this just older android, only one person mention 4.x, specifically 4.1 being a problem.

illarionov:
Nexus 5 is fine.
I have noticed this behavior on the  Alcatel One Touch 4030 S'Pop (android 4.1.1) and Cubot GT72E (android 4.2.2)

garvankeeley:
illarionov, this all looks good. I'm not able to test it, so if anyone reading this can provide further confirmation that this works for them as expected, that would be helpful. We have a few items now for a release, so I'll do one tomorrow.

PR #615, opened 2014-06-03
Update SSIDBlockList.java

VolMi:
I encountered the SSIDs "MeinFernbus1" and "MeinFernbus2" recently. There is no ambiguity without the "-" as Fernbus means "long distance bus" in german.


cpeterso:
Looks good to me. Thanks, @VolMi!

Issue #610, opened 2014-05-27
Include an attributation for OSM (again)


porjo:
Issue #239 dealt with this however I don't believe it went far enough. Currently there is just a link that reads _"Learn more about our mapping provider"_ which links to Mapbox, the supplier of the rendered tiles. The primary source of the raw map data is Openstreetmap and I think that deserves a specific mention on the about page, including a link to openstreetmap.org

garvankeeley:
Thanks, we can certainly add the added attribution in the next release.

Djfe:
@garvankeeley you should definitely do that before releasing the next version tomorrow ;)

garvankeeley:
Will do thanks for the reminder

Djfe:
Can be closed now. Has been introduced with 0.20.5 ;)

porjo:
Thanks, that looks better :)

PR #616, opened 2014-06-04
issue #610: better credit to OSM on about screen

garvankeeley:
issue #610: OSM credits added and a link to OSM.


garvankeeley:
Very minor, so I'll self-merge as I am about to kick a release build

PR #633, opened 2014-06-19
Android Studio

keverets:
This will block on Travis upgrading to the Android Build tools version 19.1.0.

Upgrades to gradle, gradle android plugin, and android build tools. Required for Android Studio 0.6+

Also ignore AS files (.idea and .iml)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 .DS_Store
 .gradle
 .settings
+.idea
+*.iml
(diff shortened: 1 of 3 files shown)


garvankeeley:
@keverets, we have to update travis to match, I just updated Android Studio and had the same problem

garvankeeley:
issue #637, updated travis

PR #640, opened 2014-06-23
Added log activity to options menu.

garvankeeley:
I find this useful for understanding what the app is doing.
Also switched to black UI style.


garvankeeley:
Lint was complaining that when using appcompat, the namespace of showAsAction in menu.xml should be changed to 'app', like this:
 <item
        android:id="@+id/action_view_map"
        android:icon="@drawable/ic_action_location_searching"
        app:showAsAction="always"

Problem with this, is that now everything is in the menu, showAsAction is no longer respected. I'll have to push another fix for this. Thank-you Mr. Lint.

Issue #621, opened 2014-06-09
Attach a scrollbar to the about page


Djfe:
Since OSM is mentioned on the about page, the CC license at the bottom is cut off on my device.
I think it would make sense to define the page as scrollable.

garvankeeley:
https://github.com/mozilla/MozStumbler/pull/644

anishgkanchan:
I had pulled a request with the scrollbar added. However the Travis CI build failed. I deleted the branch both locally and remotely. Kindly guide me on what is to be done next.

Djfe:
@garvankeeley this is fixed now so you can close it as well ;)

PR #644, opened 2014-06-23
Attach a scrollbar to the about page #621

anishgkanchan:
A ScrollView has been added to activity_about.xml.


garvankeeley:
This looks good so far, thanks! I'll wait until travis completes and then close it.
Note that I added a reference to the original issue, so they are now linked

PR #650, opened 2014-06-24
Translate missing strings (fi)

ojarva:
Translated all strings that were marked as missing / needed update.


garvankeeley:
Thanks @ojarva!

Issue #581, opened 2014-05-10
Pull SSID blocklist from web


garvankeeley:
We should pull the blocklist from github. Periodically update based on connection preference for uploading data: wifi-only or wifi and cell.

The Updater.java has most of the code we need for doing this.

hannosch:
Great idea. Could you make sure the blocklist is in a simple data format? Ideally the blocklist could be re-used by different implementations, like geoclue or FxStumbler in the future. First step is making sure the file is in a simple general format.

I think @cpeterso also wanted to pull out the list into its own repository, so it can be versioned / updated independently of MozStumbler.

garvankeeley:
Could be a json array, with regex elements (my preference).

{
“ssid_blocklist": 
[
{ "pattern":"._i[Pp]hone$", "info":"Ad-hoc iphone networks" }, 
{ "pattern":".__nomap$" , "info":“Standard identifier to request SSID is ignored by scanners" }
]
}

Or, with no regex, it should specify if it is a prefix, suffix, and if it is case-sensitive. 
Or could just be newline-separated file, with a # for comment lines. 
Any preference?

garvankeeley:
Implementation:
- have the Blocklist as a JSON file as suggested above
- parse this file to build the Blocklist logic
- update the list: when we upload data, do a separate request to a github page where the Blocklist sits, if newer, grab it

garvankeeley:
openSourcer: since we will need a default block list, I would put the file in res/blocklists, so we can distribute the current list with the app. 
The plan is to update this file periodically of course, but if there is any problem with getting a block list the app can fallback to using the built-in block list. 
When looking for updates to the file it will look in github, in res/blocklists. Perhaps github has a way to query file dates using the REST interface, and if the date is newer, grab the new file.

garvankeeley:
Another consideration is how the service will use it, without the client. The service is pure .java files for use in Firefox for Android. We could have a default version of this file in service/blocklist, like DefaultBlockList.java, that file would just have a string:
public static final String defaultListJSON = "{\"ssid_blocklist\":[" +
"{ \"pattern\":\"^AndroidAP._\" , \"info\":\"Mobile devices\" }," +
"{ \"pattern\":\"^(Android)| Hotspot._\" , \"info\":\"Mobile devices\" }," +
"{ \"pattern\":\"^barnacle.*\" , \"info\":\"Android tether app\" }," + etc.

This list could be a minimal list, the most important item is the _nomap.

[...]

PR #653, opened 2014-06-24
Pull blocklist from web #581

anishgkanchan:


garvankeeley:
I pulled your branch and it did build, still we need to find out why this won't build before it gets into the codebase. We depend on Travis. The prolog of blocklist.txt is the problem, perhaps it needs to be a .json file and properly declared as such in the first line. Not sure

garvankeeley:
The Blocklist is getting downloaded, but where is it being applied? Probably the code is there somewhere, but because of all the commits being split up, I would have to go digging for it. 
(BTW, I made this mistake plenty of times!)

anishgkanchan:
Would you please tell me how I'm supposed to use the blocklist downloaded? One way would be to add a method in SSIDBlocklist.java which would read the blocklist.

cpeterso:
hi @anishgkanchan, are you still working on this downloadable blocklist feature? I spoke with @garvankeeley offline and I think this feature might be more complicated than what we need. While we do update the SSID blocklist more often (from contributors' pull requests) than we release new MozStumbler builds, I don't think waiting until the next MozStumbler release to get the new blocklist is a big problem.

anishgkanchan:
@chrispeterson actually I was working on it but then @garvankeeley told me
that I could work on it later because I was a bit hazy on it's further
implementation.
So I started reading up on the 'Display the user's "tracks" on
MozStumbler's map issue.
Later after an Android Studio update the 'MozStumbler' project stopped
building altogether due to some Gradle issue.
Do you suggest that I don't work on the 'Pull blocklist' issue any further?

Regards,
Anish Kanchan
Student, University of Mumbai

cpeterso:
hi @anishgkanchan, since you were blocked on the implementation, I think there are other MozStumbler features you could work on that would have more immediate benefit for users.

I think issue #664 ("Move SSID blocklist from Java code to user-configurable text file on /sdcard") might be an interesting bug for you because you are already familiar with the SSID blocklist code. If we move the blocklist from Java code to the sdcard, then people could edit it on their own devices. They could add private SSIDs to their blocklist or even remove the default SSIDs. :)

I'm going to closed this PR for now. If you have any questions, just add an @cpeterso question in the issue or, better yet, drop by the #geo IRC channel on irc.mozilla.org.

anishgkanchan:
The  'Move SSID blocklist from Java code to user-configurable text file on
/sdcard' issue was closed few days back.
I had already done some work on it before. I wished to know if that portion has already been coded.
I'd like to work on it.

crankycoder:
@anishgkanchan  there is no code for this yet.

anishgkanchan:
Yes cool. I'll continue my work on issue #664.

Issue #622, opened 2014-06-09
Make about page more consistent


Djfe:
The link to the osm project has a different color than the one to mapbox and it is underlined.
This should be more consistent, I think.

Another thing:
Maybe we should mention the provider of our cards directly instead of only linking to their page. (the name "mapbox" should be mentioned in my opinion)

Djfe:
Why isn't the osm string not available for translation btw.?
And why is the text and URL in a layout XML file instead of being in the about activity in the src folder like everything else?

The structure is definitely not consistent.

Other people who want to add something will be confused about this.

garvankeeley:
Djfe, feel free to address these issues if you have time, thanks.

pl6025:
I can do it if you confirm the syntax:
In res/layout/activity_about.xml
Line 34 android:text="Map view tiles © OpenStreetMap contributors\nhttp://www.openstreetmap.org "/>
--> android:text="@string/MapViewTiles"

In res/values/strings.xml 
Addition of:
<string name="MapViewTiles">Map view tiles © OpenStreetMap contributors\nhttp://www.openstreetmap.org </string>

PR #654, opened 2014-06-24
Make about page more consistent #622

anishgkanchan:


garvankeeley:
This should be a single commit, but I'll give you a pass on this one :).
Issue #622 fixed.

PR #655, opened 2014-06-24
Attach a scrollbar to the about page #621

anishgkanchan:


garvankeeley:
Lint is just out to get you today! I can't see from looking at this what might be the problem. See if you can find out how to view the lint results on travis if you can't see the error locally.

anishgkanchan:
Haha! 
Could you please guide me on using Lint with Android Studio. I did Analyze>Inspect Code, but I could not find any serious issues in it. I might be missing something. Also gradlew just doesn't work at my place.

PR #660, opened 2014-06-26
Switch from osmdroid 4.1 to mapbox-android-sdk 0.3.0

keverets:
WIP: adding an additional tile source doesn't seem to combine them.  Asked Mapbox support about this, as it may be necessary to make one of the tile sources into an Overlay.

This requires dropping support for Android API level 8, so for mapbox integration, it will need to be 9+.

Also removed the querying of MLS for position info.  That can be added back in a separate activity, as this Map Activity should be used by stumblers to determine where in their area has already been stumbled.  The testing of MLS responses was never clear in the MapActivity.

The "Refresh" button now refreshes the map back to the user's location.  This is useful if they start to pan around the map a bunch and get lost.

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -5,7 +5,7 @@
     android:versionName="0.20.5" >
 
     <uses-sdk
-        android:minSdkVersion="8"
+        android:minSdkVersion="9"
         android:targetSdkVersion="19" />
 
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION"/>
--- a/build.gradle
+++ b/build.gradle
@@ -20,8 +20,12 @@ android {
     buildToolsVersion '19.1.0'
 
     dependencies {
-        compile "org.osmdroid:osmdroid-android:4.1"
-        compile 'org.slf4j:slf4j-android:1.7.7'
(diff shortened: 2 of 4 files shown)


garvankeeley:
@keverets Thanks! Grabbing now to try out locally...

garvankeeley:
It is looking good. I'll have to try it with our map box provider. I can see the overlay tiles are loading faster, and the view is more responsive, the base map tiles are slow to come in.
BTW, there is something wrong with travis-CI at the moment, all the builds are failing with:
Execution failed for task ':preBuild'.

> failed to find Build Tools revision 19.1.0

garvankeeley:
Travis Tweet about the problem:
https://twitter.com/travisci/status/482255297738985473
Need a new android license key

keverets:
Right now the overlay tiles are set to the base tile layer, as the
"addTileSource" doesn't do as I expected and merge the the tile layers.  If
you swap the tile layers in the constructor you should see the base map
show up quickly and with better caching, but then the overlay tiles won't
load.

Once I get the tile merging/overlay issue sorted out I'll update the PR.

garvankeeley:
Closing until we come back to this. Let me know if that isn't the right github for procedure for WIP stuff after it has been explored.

PR #666, opened 2014-07-02
Added ScrollView to the activity_about.xml #621

Jimmar:
Added ScrollView to the activity_about.xml file and reorganized it a little bit


garvankeeley:
Thanks, worked for me!

PR #675, opened 2014-07-03
Bug 1024708 - Pre: Add MPLv2 license headers to code that ships with Fennec.

ncalexan:
Trivial, but needs to be done.  I have a tool that does this in one line if you want to land such license headers in more than just `service/`.


garvankeeley:
@ncalexan Thanks for doing that.

Issue #671, opened 2014-07-02
0.20.6 settings got reset


Djfe:
the update resetted all my mozstumbler settings
is there a way to transfer/import the settings into the new format?
If so then it would be nice if somebody implements it before the next version is released ;)

garvankeeley:
I don't know if I'll be able to recover your old settings, but I will make sure to fix the settings upgrade bug

Djfe:
you don't need to recover my old settings but it would be nice for all other users if they could keep their settings after updating to the next version (since 0.20.6 was only a pre-release most of them still have their settings)

PR #677, opened 2014-07-04
try to preserve old prefs on upgrade

garvankeeley:
Manually copy from the expected old location to the new one.


garvankeeley:
Fix for issue #671

PR #680, opened 2014-07-04
Persist the "Upload only on WiFi" and the "Geofencing" preference

keverets:
Prior to this change, the user couldn't de-select "Upload only on WiFi" setting and have it persist.  Going back into settings would show that the "Upload only on WiFi" setting was re-selected.

Geofence Enabled had a similar fate.  Sometimes it would not disable if the GeofenceEnabledHere was not false.


garvankeeley:
@keverets much appreciated!

Issue #663, opened 2014-06-29
Upload silently fails multiple times and the data is lost


DonTomika:
I've been playing with MozStumbler in the past few days and collected about 4000 observations, from which about 1000 succeeded to upload, but the others failed and eventually got deleted. I did some experiments and it seems that uploading 4-5 observations always succeeds, but uploading more (like 100) fails most of the time, at least for me. After 3-4 retries, the data is deleted.

**Device:** HTC One SV (Android 4.2.2), using MozStumbler v0.20.5

**Network:** tried multiple Wi-Fi connections, and also cellular networks. Others apps that rely on the same connection are working fine.

**Related issues:** https://github.com/mozilla/MozStumbler/issues/546 and https://github.com/mozilla/MozStumbler/issues/651

**Logcat messages:**

```
06-29 18:53:26.781: E/org.mozilla.mozstumbler.communicator.AbstractCommunicator(6941): Couldn't compress and send data, falling back to plain-text: 
06-29 18:53:26.781: E/org.mozilla.mozstumbler.communicator.AbstractCommunicator(6941): org.mozilla.mozstumbler.communicator.AbstractCommunicator$HttpErrorException
06-29 18:53:26.781: E/org.mozilla.mozstumbler.communicator.AbstractCommunicator(6941):  at org.mozilla.mozstumbler.communicator.AbstractCommunicator.sendData(AbstractCommunicator.java:96)
06-29 18:53:26.781: E/org.mozilla.mozstumbler.communicator.AbstractCommunicator(6941):  at org.mozilla.mozstumbler.communicator.AbstractCommunicator.send(AbstractCommunicator.java:104)
06-29 18:53:26.781: E/org.mozilla.mozstumbler.communicator.AbstractCommunicator(6941):  at org.mozilla.mozstumbler.communicator.Submitter.cleanSend(Submitter.java:40)
06-29 18:53:26.781: E/org.mozilla.mozstumbler.communicator.AbstractCommunicator(6941):  at org.mozilla.mozstumbler.sync.SyncAdapter.onPerformSync(SyncAdapter.java:91)
06-29 18:53:26.781: E/org.mozilla.mozstumbler.communicator.AbstractCommunicator(6941):  at android.content.AbstractThreadedSyncAdapter$SyncThread.run(AbstractThreadedSyncAdapter.java:254)

06-29 18:53:27.992: E/org.mozilla.mozstumbler.communicator.Submitter(6941): Error submitting: 
[...]

PR #681, opened 2014-07-04
Fix issue 663 (I hope). Don't record cellinfo with unknown network type.

garvankeeley:
Fix Issue #663.
Also added upload info in the log activity (and some color!).
Renamed a file to be more consistent.

### DIFF
--- a/src/org/mozilla/mozstumbler/client/LogActivity.java
+++ b/src/org/mozilla/mozstumbler/client/LogActivity.java
@@ -5,14 +5,22 @@ import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
+import android.os.Handler;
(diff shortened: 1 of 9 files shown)


KAMI911:
However I am using the latest version I still lost measurements when I upload huge amount of data.

garvankeeley:
What does the Log Activity show? It should indicate what the error is.

I added a test case that I used to create 100000 records and upload them, which doesn't fail. To test, I increase the number of generated records in: 
https://github.com/mozilla/MozStumbler/blob/master/src/org/mozilla/mozstumbler/tests/AppTest.java

I need to add more stress conditions to this case to get failure, particularly to simulate intermittent network problems.

naoliv:
Yesterday I also lost some data after the wifi went down while uploading.
In the log I could see this:

screenshot_2014-07-26-23-20-40

PR #682, opened 2014-07-04
New Android Studio 0.8.1 requires a newer gradle plugin 0.12+

keverets:
Hopefully these revisions will slow now that AS is in beta.


garvankeeley:
Just updated, I need this too.

PR #686, opened 2014-07-06
MapActivity fixes

keverets:
Here's some fixes to the MapActivity to make it more useful to stumblers.  It seems that before it was intended to be just a test of the MLS to see what the server responds with.  I think a test is less important than a stumbler being able to determine where they should stumble next, hence using the GPS fixes to animate the map around, and making rotation a bit less jarring.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,7 +20,7 @@ android {
     buildToolsVersion '19.1.0'
 
     dependencies {
-        compile "org.osmdroid:osmdroid-android:4.1"
+        compile "org.osmdroid:osmdroid-android:4.2"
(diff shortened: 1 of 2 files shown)


garvankeeley:
Thanks! Grabbed the patch, applied, going for a walk to try it out...

garvankeeley:
Usability is much improved, rather than regularly seeing the world view (when rotating, after stopping and leaving the app, etc), after the first zoom, it stayed on my area correctly. Animating to location looks good.

If we add back in MLS location, we should have one marker clearly marked 'GPS', and the other marker marked 'MLS'.

PR #718, opened 2014-07-23
MainActivity ui update: change to ensure it runs on the main thread

garvankeeley:
Wrap the update in a main thread runnable. Allows for safe updating of the ui from secondary thread.


garvankeeley:
Addressed review from @keverets and re-pushed

PR #719, opened 2014-07-23
Change to using a pin-point (with accuracy circle) to indicate stumbler's GPS position

keverets:
This drops the ItemizedOverlay for indicating position, and re-introduces use of the AccuracyCircleOverlay, which is also given a mild make-over to draw the exact pin location and keep its Paint objects around rather than recreating them on each draw.

To get the accuracy circle, the full Location is now passed out from GPSScanner all the way back to the MapActivity, rather than grabbing the Latitude & Longitude piecemeal.

The center point could be improved further, and the accuracy circle often gets obscured by it, but hopefully it's an improvement over the old push-pin to indicate location.


garvankeeley:
Tried it out, looks good, thanks!

PR #721, opened 2014-07-23
cleaned up SSIDBlockList

Djfe:
cleaned up the SSIDBlockList and applied @cpeterso s changes (switched tabs with spaces)


cpeterso:
Looks good. Thanks, @Djfe!

PR #722, opened 2014-07-24
Translation updated

yarons:
Added missing strings


cpeterso:
Looks good. Thanks, @yarons!

PR #723, opened 2014-07-24
Translated missing Finnish strings

ojarva:


cpeterso:
Thanks, @ojarva! Looks good.

Issue #708, opened 2014-07-14
Sometimes doesn't capture anything


naoliv:
I am seeing something strange with MozStumbler: I used to run it together with GPS Logger (https://play.google.com/store/apps/details?id=com.mendhak.gpslogger) and Tower Collector (https://play.google.com/store/apps/details?id=info.zamojski.soft.towercollector) but saw that it wasn't capturing data sometimes.

Since both GPS Logger and Tower Collector were properly capturing data, I thought that it could be some kind of concurrence (two apps trying to access the same resource, maybe).

Then I started to use only MozStumbler but sometimes I still see the same problem: even having 11 visible GPS/Glonnas satellites, sometimes it just doesn't capture anything.
The scanned locations remains at 0, while I can see the GPS satellites, cells and wifi points being updated.

garvankeeley:
Is there a GPS entry in the log activity in MozStumbler? I wonder if the device doesn't have a non-fused GPS location yet (Fused: using Google Location Services as assistance).

naoliv:
The phone is a Xperia Mini Pro running Android 4.0.4, just in case.

With adb logcat this is what I see when opening MazStumbler:
http://paste.debian.net/plain/109709

And after pressing to start scanning (with wifi disabled):
http://paste.debian.net/plain/109710

With "GPS" I can only see this:
loc_eng_send_gps_status: status = 1
loc_eng_agps_data_conn_failed
loc_eng_send_gps_status: status = 4

It's also strange that I see it loading TTS stuff that I don't see with GPS Logger, for example:
http://paste.debian.net/plain/109711

garvankeeley:
Can you check the new Log Activity in the app?

naoliv:
The log is empty (even with 12 sattelites visible).
Should it contain info only when it acquires a position or it should be displaying something else?

garvankeeley:
I'll have to do an update that shows more info in the log, this should be easy to figure out if I display the GPS source attribute (network/fused/GPS-only).

Djfe:
The log should contain the same stuff that is written into the system log in my opinion.
So it would be nice if you could add that info to the log as well.

naoliv:
Testing now it worked (like said initially, it's sporadic when it works or not).
I can see this in the log:

![screenshot_2014-07-15-13-59-43](https://cloud.githubusercontent.com/assets/982338/3588398/21ca02a0-0c43-11e4-9805-864b317e38bb.png)

[...]

PR #725, opened 2014-07-26
Put initialization of GPS broadcast reception in onStart()

keverets:
It's removed in onStop() so should be initiated in onStart().

Without this change, if the user left the map (via notification, home button press, etc) and came back, they would no longer get GPS updates.

The only obvious way to restart updates was to go back from the map activity, then re-enter it to force a re-creation.


naoliv:
Could this be causing #708?

garvankeeley:
I don't think this is causing #708.

PR #727, opened 2014-07-26
Add PathOverlay to the map to draw the GPS track

keverets:
This is a work-in-progress, which still needs refinement and is posted to solicit feedback.

It will remember the user's location changes while the Map Activity is up and running, and draw that in the same color as the GPS dot, persisting across screen rotations, etc.

I'm planning to later change this to grab historical data from the service, or some other local store if the user enables it in the setting.

An issue is that OSMDroid is no longer recommending using PathOverlay and now says to use the OSMBonusPack Polyline, however MozStumbler does not yet include that library, and it doesn't seem to have a nice maven import available.

If the Map Activity is changed to use the Mapbox Android SDK, that SDK uses their own implementation of the PathOverlay which may be sufficient, though I'd suspect it would still be prudent to clip the paths that are passed to the overlay to only those that are within some boundry of the currently visible map as I imagine there will be performance implications with users having their entire stumbling history drawn.

### DIFF
--- a/res/values/colors.xml
+++ b/res/values/colors.xml
@@ -1,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <color name="gps_track">#FF6464FF</color>
+</resources>
--- a/src/org/mozilla/mozstumbler/client/mapview/MapActivity.java
+++ b/src/org/mozilla/mozstumbler/client/mapview/MapActivity.java
@@ -24,6 +24,8 @@ import android.widget.ImageView;
 import android.widget.TextView;
 
 import java.net.URL;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Timer;
 import java.util.TimerTask;
 import org.json.JSONException;
@@ -48,6 +50,7 @@ import org.osmdroid.util.GeoPoint;
 import org.osmdroid.views.MapView;
 import org.osmdroid.views.Projection;
 import org.osmdroid.views.overlay.Overlay;
+import org.osmdroid.views.overlay.PathOverlay;
 import org.osmdroid.views.overlay.TilesOverlay;
 
 public final class MapActivity extends Activity {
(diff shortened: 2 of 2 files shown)


garvankeeley:
Looks like a good start, walked around a bit with it today. The path isn't reloading when I exit and return to the map activity, although, I am working on my branch, which is heavily modified.

keverets:
Yes, for the moment the list is stored in the Activity (just passed through
the onSaveInstanceState() during rotation) so once the Activity is
destroyed it goes away.

I was still pondering how best to persist track history, which I'd like to
make an option to the user. It seems like it should be the job of the
Service, which is already keeping short history for uploading purposes. I
think extending that to hold all history when desired, and exposing an
interface so that the Map can query it, might be a good way forward.

What are your thoughts on that?

Djfe:
That's the plan already but no one has started working on such a database, yet.
See #3

garvankeeley:
I would prefer to keep this client-side, the track history just be persisted in-memory, make a static list in the MapActivity class?

keverets:
A static list in the MapActivity would mean that if the Map Activity wasn't being displayed, the track wouldn't be recorded, since the broadcast receiver will only be active while the Map Activity isn't stopped.

Should there be another service, with the sole purpose of recording the track (if the user selects it)?

Djfe:
Not neccessary if the map will be the default view
The map could be integrated into the Main activity

garvankeeley:
My static idea was not too well thought out :). Yeah a MapGPSTracker service or something would be good to have for this. 
Somewhat aside, I can foresee the Map Activity getting its own location updates in the future, and we would indicate which location updates came from the service (as pure GPS). This is a common point of confusion it seems, as to why the app location isn't like other apps. Walking around in Toronto, a lot of the locations come in as Fused locations, and stumbler will not use these for a stumbled location. Maybe we can do something in future in the Map Activity that clarifies what the app is doing.

garvankeeley:
Kevin, can I merge this, then I'll add the GPS receiver for the map activity tracks? I can whip something up based on your info on the bug. Then I'll put out a beta with the latest changes today.

keverets:
Sure

PR #730, opened 2014-07-27
Use the service's last location if available and the center hasn't been set in other ways

keverets:


garvankeeley:
Makes sense, thanks Kevin.

PR #737, opened 2014-07-30
Fennec Work: pull database back-end, replace with pure file-based storage

garvankeeley:
Fennec will not allow the use of a database for a minor feature such as this, the overhead of a DB is too great, and existing use of a db was there for convenience.

Unfortunately, this required a rather hefty refactor. The bonus being, a lot of the problems with missing uploads and disappearing data should be gone. I hope. 

As part of this, MozStumbler client is now using the same upload method as the stumbler service, which does not use Android sync.


Djfe:
Android Sync was introduced to reduce the power usage. People said it saves power because it bundles syncs of several apps.
But I think it only introduced problems, so I think it's great that you remove it again ;)

garvankeeley:
I am rebasing this commit and cleaning it up, I did a bit before the PR, but in retrospect it needs a cleaner history for the reader.

garvankeeley:
Totally reformatted the PR, not sure what the best github protocol is, so I am closing this one, opening another.

PR #738, opened 2014-07-30
Fennec Work: pull database back-end, replace with pure file-based storage

garvankeeley:
[Refactored (rebase -i) my previous PR. Tried to break out the minor commits, which one can quickly review, and the BIG commit which is fea7490 'The grand refactor' -that one requires more thought].

Fennec will not allow the use of a database for a minor feature such as this, the overhead of a DB is too great, and existing use of a db was there for convenience.

Unfortunately, this required a rather hefty refactor. The bonus being, a lot of the problems with missing uploads and disappearing data should be gone. I hope.

As part of this, MozStumbler client is now using the same upload method as the stumbler service, which does not use Android sync.

p.s. I had planned to also squash 008d38e, just realized I left that intermediate work in. Sigh. Leaving it for now, I don't think it is too confusing to see that step in the development.


garvankeeley:
I think this is ready to merge. I am going to put up a beta today.

PR #739, opened 2014-07-31
Use the last map center (not necessarily the last user location), instead of (0,0)

keverets:
Rather than having the user see (0,0) every time they start, persist the last map center, and use that if the stumbler service doesn't yet have a real location.

Persisting some preferences for the Map separate from the the rest of the app/service seemed to make sense in this context.  Can move that if desired.


garvankeeley:
I found this was overriding the lat long from the prefs in onCreate:

@Override
        protected void onPostExecute(Location result) {
            setUserPositionAt(result);
            if (!mUserPanning) {
                positionMapAt(result); <-- I get (0,0) arrive after onCreate
            }

...

I'm not sure if I am testing this the right way. I shut off GPS, then I reboot the app, then go to the map view. I was expecting that it would centre on my last known location, but it doesn't.

garvankeeley:
I just saw this was updated! Thanks Kevin. I wish github had a notification that PRs were updated.

PR #741, opened 2014-07-31
Amendment to 727, track locations when the activity is not showing.

garvankeeley:
Also snuck in call to make the line thicker.

See issue #727


garvankeeley:
Kevin, does this look ok?

PR #752, opened 2014-08-01
Fennec code standard

garvankeeley:
Addressing issues from here:
https://bugzilla.mozilla.org/show_bug.cgi?id=1038843


garvankeeley:
Note to reader: I will be rebasing this heavily before it goes in, rather than adding 20+ cleanup commits to the tree :)

PR #769, opened 2014-08-19
Coverage overlay zoom

keverets:
This moves the coverage overlay into its own class which, for the moment, extends the TilesOverlay.  Once it is changed to be purely vector based, it no longer need do so, but there should not be changes necessary to the MapActivity at that point as it sees the coverage map as a simple Overlay.

When asked to draw zoom levels greater than those provided by the coverage map (currently level 13), this will now determine a scaling factor, grab the lower zoom level image, and do a simple scale to draw the overlay.

Once the vectorization is at a happy point, it should be able to be swapped into the mCoverageTileLooper with minimal fuss.

### DIFF
--- a/src/org/mozilla/mozstumbler/client/mapview/CoverageOverlay.java
+++ b/src/org/mozilla/mozstumbler/client/mapview/CoverageOverlay.java
@@ -1,0 +1,27 @@
+package org.mozilla.mozstumbler.client.mapview;
+
+import android.content.Context;
+import android.graphics.Color;
+
+import org.osmdroid.DefaultResourceProxyImpl;
+import org.osmdroid.tileprovider.MapTileProviderBasic;
+import org.osmdroid.tileprovider.tilesource.ITileSource;
+import org.osmdroid.tileprovider.tilesource.XYTileSource;
+import org.osmdroid.views.overlay.TilesOverlay;
+
+/**
+ * This class provides the Mozilla Coverage overlay
+ */
+public class CoverageOverlay extends TilesOverlay {
+
+    public CoverageOverlay(final Context aContext, final String coverageUrl) {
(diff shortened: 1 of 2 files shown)


garvankeeley:
Great work @keverets! Finally I can stop switching to the Ichnaea Map View when I am stumbling. I can actually see gaps in collection in my neighbourhood I hadn't noticed before. This is a huge step forward for the app. 

Do you know if the Mapbox OSM sdk still has tile overlay problems. The performance is the last piece of the puzzle I think.

keverets:
The Mapbox tile overlay is still a problem, sadly: https://github.com/mapbox/mapbox-android-sdk/issues/380

hannosch:
@keverets Would it help, if I'd add a tilejson file describing the coverage data overlay on the service end? I could extend https://location.services.mozilla.com/map.json to be valid tilejson as well.

Not sure if that helps though :)

keverets:
@hannosch TileJSON is almost certainly better to parse and deal with rather than doing image manipulation on the .png files... if that's offered I'd be very glad to make use of it.

keverets:
Hrm, oh, wait. I think I confused GeoJSON for TileJSON...

Djfe:
Thx a lot for your submission! ;)
Like garvan I've missed seeing the coverage status in the higher detailed levels of the map.

I really hope that there will be a good balance/compromise between privacy protection and coverage detail once vectoring is implemented :)

hannosch:
@keverets yeah, I only meant tilejson (a description of where a tile layer is hosted).

GeoJSON in itself isn't enough, as it has no standard way to split the data into multiple chunks, like the zoom level, y, x system used in image based layers. And the data we have is too big to download in one chunk.

garvankeeley:
This is working great for me. The only problem is the performance of the mapview with the composited tiles, which is not related to this PR.

PR #770, opened 2014-08-19
Use MapQuest OSM tiles instead of Mapnik

keverets:
Members of the OSM community recommended making use of the free MapQuest tiles instead of leaning on the mapnik infrastructure.  More information available at:

http://developer.mapquest.com/web/products/open/map

### DIFF
--- a/src/org/mozilla/mozstumbler/client/mapview/MapActivity.java
+++ b/src/org/mozilla/mozstumbler/client/mapview/MapActivity.java
@@ -231,7 +231,7 @@ public final class MapActivity extends Activity {
     @SuppressWarnings("ConstantConditions")
     private static OnlineTileSourceBase getTileSource() {
         if (BuildConfig.TILE_SERVER_URL == null) {
-            return TileSourceFactory.DEFAULT_TILE_SOURCE;
+            return TileSourceFactory.MAPQUESTOSM;
         }
         return new XYTileSource("MozStumbler Tile Store",
                                 null,


garvankeeley:
For non-mozilla builds, anything seems to be better than hammering the default non-commercial OSM tile source (which is intermittent, possibly by design, since they would prefer the public host their own mirrors).

garvankeeley:
Will need to review terms of use:

Terms of Use
The MapQuest Open Tiles are available for use for free under the following conditions:

If using the MapQuest-OSM tiles, OpenStreetMap must be given credit for the data – see this section of OSM's legal FAQ for details.
Please place “Tiles Courtesy of MapQuest ” on your page, or in the copyright area of the map and link the word “MapQuest” to http://www.mapquest.com. 
If your application will get heavy usage (currently defined as more than 4,000 tiles per second), please let us know by sending us an email at open@mapquest.com. Please include the estimate of your expected usage so that we will be aware and accommodate the extra traffic.
If there will be an announcement (for example – a press release or a corporate announcement), please contact MapQuest at open@mapquest.com to let us know of the good news and any expected jumps in map tile usage.
Usage of these tiles are at your own risk.
Usage of these tiles is governed by our Terms of Use.
http://developer.mapquest.com/web/info/terms-of-use

keverets:
I'm kind of surprised that OSMDroid doesn't include the copyright area on the map by default.  I'll look at the code and figure out if there's a good way to do that.

Do we know how many tiles per second are being sucked from the coverage map (or the mapbox base map)?

garvankeeley:
No idea, Hanno can likely us help come up with some guesstimates.

hannosch:
MozStumblers map usage is currently between 500-1000 mapbox map views per day. They count one map view as 15 tiles being loaded, so you get 7.5k-15k tiles loaded by day or at max 0.2 tiles per second.

That is only counting the Mozilla builds, which should account for one third to half the MozStumbler usage.

So we are nowhere near the 4k/sec tiles mark.

garvankeeley:
Will file the bug to get the copyright info shown, and do so prior to next release. Might still do a beta before this is done to get feedback.

PR #774, opened 2014-08-23
Keep the screen on when in the Map Activity

keverets:


cpeterso:
@garvankeeley plans to make the Map Activity the app's primary screen. I don't know if we would still want to force the screen on then or not.

Is setting `FLAG_KEEP_SCREEN_ON` programmatically preferred over setting `android:keepScreenOn="true"` in layout/activity_map.xml? I don't have a strong opinion. :)

keverets:
I thought about possibly making it an option, but the user still has control (can manually put the device to sleep or move to a different activity) and the device will no longer keep the screen on.  In https://developer.android.com/training/scheduling/wakelock.html it mentions:

> The advantage of this approach is that unlike wake locks (discussed in Keep the CPU On), it doesn't require special permission, and the platform correctly manages the user moving between applications, without your app needing to worry about releasing unused resources.

Being that many other mapping apps I use don't bother the user with the option and just keep the screen on while they're driving/exploring, I thought it might be best to do the same.  In the worst case, we have a discussion about it ;)

I don't have a strong preference either way for setting the flag in the layout or in the onCreate().  Either way is one line, but from the page referenced earlier it offers one possible advantage:

> The advantage of setting the flag programmatically in your activity is that it gives you the option of programmatically clearing the flag later and thereby allowing the screen to turn off.

Just keeping the door open in case we want to make it user selectable.

cpeterso:
Since other map apps keep the screen on, it makes sense for us, too. The user can always shut their screen off manually. :)

I would vote against adding yet another user option, but I'm not strongly opposed. We could keep the screen on as long as the device has moved in the last minute or two.

garvankeeley:
I use a lot of mapping apps, so I have my screen sleep set to a 10 min timeout. For me, this would only cause confusion (well, not really, because I know the code), I wouldn't want the app to wake lock the screen. Others, might want this on though.

If we ever did a first run screen, I would like 2 questions answered from users, 
- can we upload on cell?
- (and now this one) do you want the screen kept on?
  And in both cases, point the users to the settings if they want to change these settings later.

crankycoder:
I for one would rather not have the screen stay on unless I set the preference. I don't expect _any_ of my applications to keep the screen on unless there it's doing something where I would be expected to be watching the screen but not touching the screen for extended periods of time.

That amounts to video players - but that's pretty much it.

keverets:
Most map applications will also do this without bothering the user for a
preference. You can try them and see that they don't fade automatically,
but it rarely comes up as the user usually manually puts the device to
sleep or switches applications.

I stumble while driving and having the screen on helps me make decisions on
routes as I go without having to touch the screen.

keverets:
@garvankeeley updated with preference.

garvankeeley:
Thanks! Now that the map is getting more interesting, I am finding I don't want it shutting off when I am looking at it.

PR #775, opened 2014-08-23
Make default zoom 16 rather than 13

keverets:


garvankeeley:
Agreed, this is about a 2x2 km area on my device, which is a sensible starting zoom level.

Issue #772, opened 2014-08-20
Update app icon


cpeterso:
@hannosch has new MLS logos from [bug 1032897](https://bugzilla.mozilla.org/show_bug.cgi?id=1032897) here:

https://github.com/mozilla/ichnaea/tree/master/ichnaea/content/static/images

garvankeeley:
Looks great!

PR #776, opened 2014-08-23
Include new logos

keverets:
Taken from https://github.com/mozilla/ichnaea/tree/master/ichnaea/content/static/images

For issue #772


cpeterso:
Looks good. Thanks, @keverets!

We should probably update our status bar icons (`res/drawable-hdpi/ic_status_scanning.png`), too.

Issue #761, opened 2014-08-12
Task Notification list item: The stop function on this is not in sync with the app 


garvankeeley:
The task notification list item has stop button when expanded, this is not in sync with the state of the app. It says "[X] Stop Scanning" when the app is currently stopped.

garvankeeley:
Originally reported as issue #710.

ioana-chiorean:
I entered a issue earlier here https://github.com/mozilla/ichnaea/issues/293 . probably wrong place :S 
Thanks for following up my comment with this issue.

Djfe:
#746 is also part of this issue, I think

And what's also irritating:
Since one of the latest versions you cannot see whether there is a GPS fix or not (it still shows x/y satellites if I deactivate GPS in my status bar) 
This was different before:
No value or a "-" instead of "x/y"
EDIT: this has been listed as #685 already

crankycoder:
Tapping repeatedly on the 'X' in the notification pulldown eventually works.  The UI shows that the scanning has stopped, but the stumbler stays in the notification area.  This is different than hitting the stop button which stops scanning _and_ removes the stumbler from the notification pull down.

crankycoder:
closed in #782

PR #782, opened 2014-08-28
fix stop scanning button in the Notification pulldown

crankycoder:
Fix for issue #761 - Ignore ctags files

This patch fixes the "Stop Scanning" button in the notification pulldown.

The old code had a secondary code path to disable the scanning in TurnOffReceiver.
This patch changes the TurnOffReceiver to just send an intent to disable the scanning using a local broadcast.

The MapActivity sent an ACTION_UNPAUSE_SCANNING intent which was being accepted by the MainApp, this has been moved into the MainActivity instead..

When MainActivity::onToggleScanningClickedView() is called, we force the UI to update.  This was previously causing problems where the MapActivity would enable the scan, but the button UI was not properly refreshing.

### DIFF
--- a/src/org/mozilla/mozstumbler/client/AboutActivity.java
+++ b/src/org/mozilla/mozstumbler/client/AboutActivity.java
@@ -1,3 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
 package org.mozilla.mozstumbler.client;
 
 import android.app.Activity;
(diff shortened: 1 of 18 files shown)


crankycoder:
@garvankeeley r?

crankycoder:
I'm seeing more problems with the map view and the start stop button state.  I'm going to add a patch to this PR that reverts the changes with respect to intent handling of unpause_scanning

crankycoder:
@garvankeeley can you try testing this now? The mapview was driving me bonkers in #783 and I think I've got it licked now.

garvankeeley:
Thanks @crankycoder, this patch works well, bug is gone, sanity is restored to notifications.

crankycoder:
@garvankeeley I've reworked the patch into one commit 6380ff2 so that we can merge cleanly back to dev.  Other than that - there are no changes w.r.t the previous set of changes.

crankycoder:
@garvankeeley ok, now that I've got fennec building, I'm coming back to this.  I'll add a scan check and tweak some of the names so that it's clear that we only disable  the scanning.

crankycoder:
@garvankeeley ok - tiny clarifying patch put in.  The service scanning is checked in https://github.com/crankycoder/MozStumbler/commit/3eb87edcab5b45858831c47a80978f27b14a2b34#diff-f75c3084f4ad8d4abf81ee96673acf3bR67

PR #800, opened 2014-09-03
Create client side prefs that extend the service prefs.

garvankeeley:
vng=r?

I am adding more prefs on the client-side, specifically hardware acceleration.


src/org/mozilla/mozstumbler/client/ClientPrefs.java
```diff
@@ -0,0 +1,47 @@
...
+public class ClientPrefs extends Prefs {
+    private static final String LOG_TAG = ClientPrefs.class.getName();
+    private static final String LAT_PREF = "lat";
+    private static final String LON_PREF = "lon";
+
+    protected ClientPrefs(Context context) {
+        super(context);
+    }
+
+    public static void createGlobalInstance(Context c) {
```
crankycoder: I generally like to have singletons to be synchronized.  ClientPrefs is only instantiated in one place anyway, and that's single threaded, so this is more of a nit.

src/org/mozilla/mozstumbler/client/ClientPrefs.java
```diff
@@ -0,0 +1,47 @@
...
+        if (sInstance != null) {
+            return;
+        }
+        sInstance = new ClientPrefs(c);
+    }
+
+    public static ClientPrefs getInstance() {
```
crankycoder: Does it make sense to change this to `getInstance(Context c)` and merge this with `createGlobalInstance`?

src/org/mozilla/mozstumbler/client/ClientPrefs.java
```diff
@@ -0,0 +1,47 @@
...
+        assert(sInstance != null);
+        assert(sInstance.getClass().isInstance(ClientPrefs.class));
```
crankycoder, line 27: Asserting the type of the instance seems unnecessary.

crankycoder:
@garvankeeley this basically looks ok aside from my own paranoid feelings about unsynchronized singletons.

src/org/mozilla/mozstumbler/client/ClientPrefs.java
garvankeeley: I have a few reasons for this design, to make the prefs decoupled from context (and Android framework awareness). If the only thing tying a .java file to the Android platform is getting a context in order to call a function, I don't like that design by matter of aesthetics (sorry should word this better). In terms of application vs. activity context, in this case, we can easily ensure that the app context is used for long-lifetime stuff. And finally, why keep passing in the context, when it is only need once, on construction of the singleton, it creates an unnecessary contract on the caller.

src/org/mozilla/mozstumbler/client/ClientPrefs.java
garvankeeley, line 27: It is necessary to catch any bugs of prefs creation. If the Pref singleton is created from the service, it is a 'Pref' type, if it is created from the client-side, the singleton is a 'ClientPref' type. Insidious bugs could ensure if this was not enforced early (someone moves creation code around and triggers this bug). We could just have a proxy that is a mini-factory to enforce this instead, and ensure the creation is of the right object type. But that would be more typing :).

src/org/mozilla/mozstumbler/client/ClientPrefs.java
garvankeeley: Will do, we should be locking this codebase down for concurrency -as it grows, more concurrency seems inevitable. And in places with no performance cost, I can't see why not (other than I try to avoid too much of it where 'synchronized' might imply concurrent ops where there is none, and I have found that has confused me on other codebases). I think it is useful in this case.

PR #801, opened 2014-09-03
MapActivity: last location not always set when switching views

garvankeeley:
MapActivity: setLastMapCenter in onDestroy as well as onSaveInstanceState.

Issue #800 should land first.

Github note:
Attempting to split issues on the same commit history on a different branch. If this approach fails, i'll choose another.


src/org/mozilla/mozstumbler/client/ClientPrefs.java
```diff
@@ -0,0 +1,47 @@
+package org.mozilla.mozstumbler.client;
```
crankycoder, line 1: Needs an MPL2 license.

garvankeeley, line 1: I always forget that, thanks for catching.

PR #802, opened 2014-09-04
Hardware acceleration option in prefs, on by default

garvankeeley:
Ideally we can make a conclusion as to whether this preference is needed, and whether hardware acceleration should be on or off. I vote for on, the user experience is much better (on the devices I have tested), and the RAM usage looks comparable to similar apps.

Issue #801 should land before this, the previous commits in this chain belong to other issues.


garvankeeley:
I rebased this, still says it can't merge, just opening another PR. issue #803

PR #805, opened 2014-09-05
Limit the gps location dot update to 1 Hz max

garvankeeley:
1 Hz is the typical max rate of a GPS, I can see it update more frequently than this, which only conflicts with pan and zoom gestures if done too often.

### DIFF
--- a/src/org/mozilla/mozstumbler/client/mapview/MapActivity.java
+++ b/src/org/mozilla/mozstumbler/client/mapview/MapActivity.java
@@ -103,38 +103,39 @@ public final class MapActivity extends Activity {
         }
     }
     private class ReporterBroadcastReceiver extends BroadcastReceiver {
+        private boolean isUpdateDotScheduled = false;
+        Handler updatePositionDotHandler = new Handler();
+
         public void reset()
         {
             mMap.getOverlays().remove(mAccuracyOverlay);
         }
 
-        public boolean isUpdateDotScheduled = false;
+        public synchronized void setUpdateDotScheduled(boolean val) {
+            isUpdateDotScheduled = val;
+        }
 
         private Runnable updatePositionDotRunnable = new Runnable() {
             @Override
             public void run() {
-                isUpdateDotScheduled = false;
-                final ClientStumblerService service = ((MainApp) getApplication()).getService();
-                final Location result = service.getLocation();
-
-                // Don't map (0,0)
-                final double latitude = result.getLatitude();
-                final double longitude = result.getLongitude();
-                if (isValidLocation(latitude, longitude)) {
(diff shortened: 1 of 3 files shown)


src/org/mozilla/mozstumbler/client/mapview/MapActivity.java
```diff
@@ -102,20 +102,53 @@ public void onReceive(Context context, Intent intent) {
...
         }
     }
-
     private class ReporterBroadcastReceiver extends BroadcastReceiver {
         public void reset()
         {
             mMap.getOverlays().remove(mAccuracyOverlay);
         }
 
+        public boolean isUpdateDotScheduled = false;
```
crankycoder: `ReporterBroadcastReceiver::isUpdateDotScheduled` is updated from one thread in the `updatePositionDotHandler` and is read from a separate thread when the `GPSScanner.ACTION_GPS_UPDATED` intent is received.  Reads and writes need to be done with synchronized getter and setters.

src/org/mozilla/mozstumbler/client/mapview/MapActivity.java
```diff
@@ -102,20 +102,53 @@ public void onReceive(Context context, Intent intent) {
...
+
+        private Runnable updatePositionDotRunnable = new Runnable() {
+            @Override
+            public void run() {
+                isUpdateDotScheduled = false;
```
crankycoder: use a synchronized setter here.

src/org/mozilla/mozstumbler/client/mapview/MapActivity.java
```diff
@@ -102,20 +102,53 @@ public void onReceive(Context context, Intent intent) {
...
+        Handler updatePositionDotHandler = new Handler();
+
         @Override
         public void onReceive(Context context, Intent intent) {
             String action = intent.getAction();
 
             if (action.equals(GPSScanner.ACTION_GPS_UPDATED)) {
+
+                // Only update location dot at 1 Hz max.
+                if (!isUpdateDotScheduled) {
```
crankycoder: use a synchronized getter here.

src/org/mozilla/mozstumbler/client/mapview/MapActivity.java
```diff
@@ -102,20 +102,53 @@ public void onReceive(Context context, Intent intent) {
...
+                    isUpdateDotScheduled = true;
```
crankycoder: And the setter here.

crankycoder:
Dang it.  You're fast. :)  r+ after the minor synchronization nits.

crankycoder:
Wait a second - I think, there's a off by 1 second error here when the GPS signal is lost.  I've sent you a PR with a change in the way 1Hz updates work.

PR #807, opened 2014-09-05
Fix #510 - Only collect StumblerBundle instances if we've got valid NMEA data

### LINKED ISSUE #510: Check for test/fake GPS data
rdandu suggested this in email:

Can we do some basic checks in MozStumbler to make sure the following is not corrupting data on the device:
a) Mock Location: Enable mock locations should not be turned on https://developer.android.com/training/location/location-testing.html
b) Fake GPS: https://play.google.com/store/apps/details?id=com.lexa.fakegps

crankycoder:
This should fix #510 

This adds an NMEA listener to the MainActivity and the Reporter.

An incoming NMEA data stream is a pretty good indicator that we're using a real GPS.

The MainActivity has stubs so that we can add visual feedback that NMEA data is being read.

The Reporter checks that we've received GGA and GSV sentences within the last 60 seconds or else StumblerBundle instances are not created.

I also did some minor code cleanup along the way.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@ build
 gradle.properties
 *.keystore
 local.properties
+
+# ignore vim swap files
+*.sw?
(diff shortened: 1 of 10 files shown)


garvankeeley:
r+ with the mods.
I was able to patch this on top of the dev trunk locally, not sure why github is having difficulty with the merge of these changes.

PR #810, opened 2014-09-05
Setting initial pan and zoom requires a delay

garvankeeley:
Addressing this bug in osm droid: https://github.com/osmdroid/osmdroid/issues/22

Setting initial pan and zoom requires a delay. Not much to review code-wise, it will have to be tested on-device. Zooming and panning to the correct location, even with a short delay, is big UI improvement when opening the map view.

### DIFF
--- a/src/org/mozilla/mozstumbler/client/mapview/MapActivity.java
+++ b/src/org/mozilla/mozstumbler/client/mapview/MapActivity.java
@@ -34,6 +34,7 @@ import java.util.Timer;
 import java.util.TimerTask;
 import org.json.JSONException;
 import org.json.JSONObject;
+import org.mozilla.mozstumbler.client.ClientPrefs;
 import org.mozilla.mozstumbler.client.ClientStumblerService;
 import org.mozilla.mozstumbler.client.MainApp;
 import org.mozilla.mozstumbler.service.AppGlobals;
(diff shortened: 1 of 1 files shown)


src/org/mozilla/mozstumbler/client/mapview/MapActivity.java
```diff
@@ -402,26 +418,34 @@ protected void onStart() {
...
         bundle.putInt(ZOOM_KEY, mMap.getZoomLevel());
         IGeoPoint center = mMap.getMapCenter();
-        ((MainApp) getApplication()).getPrefs().setLastMapCenter(center);
         bundle.putDouble(LON_KEY, center.getLongitude());
         bundle.putDouble(LAT_KEY, center.getLatitude());
+        saveStateToPrefs();
+    }
+
+    @Override
+    protected void onPause() {
```
crankycoder, line 437: Is it ok to clear all the caches everytime we pause? That seems over aggressive to me, but I'm not entirely sure.

crankycoder:
r+ after irc chit chat

src/org/mozilla/mozstumbler/client/mapview/MapActivity.java
garvankeeley, line 437: Yeah, too aggressive. I have some more refactoring coming, let me move at that for my next PR.

Issue #809, opened 2014-09-05
Review the threadsafety of StumblerService


crankycoder:
There's a bunch of possible threading issues in StumblerService. We should do a proper review and expose a threadsafe interface.

crankycoder:
Closed in #812

PR #812, opened 2014-09-05
StumblerService should have all the public accessors synchronized

garvankeeley:
issue #809: I agree, as we increase the concurrency of the app (currently this interface is not used concurrently), it makes sense to have the public interface to the service (which is represented by StumblerService.java) synchronized.

### DIFF
--- a/src/org/mozilla/mozstumbler/service/stumblerthread/StumblerService.java
+++ b/src/org/mozilla/mozstumbler/service/stumblerthread/StumblerService.java
@@ -55,57 +55,57 @@ public class StumblerService extends PersistentIntentService
         super(name);
     }
 
-    public boolean isScanning() {
+    public synchronized boolean isScanning() {
         return mScanManager.isScanning();
     }
 
-    public void startScanning() {
+    public synchronized void startScanning() {
         mScanManager.startScanning(this);
     }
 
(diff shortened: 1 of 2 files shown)


src/org/mozilla/mozstumbler/service/stumblerthread/StumblerService.java
```diff
@@ -55,57 +55,57 @@ public StumblerService(String name) {
         super(name);
     }
 
-    public boolean isScanning() {
+    public synchronized boolean isScanning() {
```
crankycoder, line 58: There's a couple places in the StumblerService where you are still calling `mScanManager.isScanning()`, those should be changed to just `isScanning()` to get the synchronized behavior.

There's also a call to `if !mScanManager.isScanning() { startScanning() }` at line 226.  That should just call `startScanning` as the scan manager internally does it's own check of the `isScanning` flag prior to starting a scan.

`mScanManager.setPassiveMode` and `mScanManager.isPassiveMode()` should also probably be synchronized as the flag may be read in one thread, and set by another thread using an Intent.

crankycoder:
@garvankeeley review is complete

garvankeeley:
Thanks! Will update as per review and merge

PR #816, opened 2014-09-09
Issue #791 - drop geofencing code

### LINKED ISSUE #791: Pull geofencing code out and target a post 1.0 revision

crankycoder:
This fixes #791 and removes all geofencing code and geofencing related resources.


src/org/mozilla/mozstumbler/service/stumblerthread/scanners/GPSScanner.java
```diff
@@ -215,21 +201,13 @@ public void onLocationChanged(Location location) {
...
         if (AppGlobals.isDebug) {
             Log.d(LOG_TAG, "New location: " + location);
         }
 
         mLocation = location;
 
-        if (!mAutoGeofencing) {
-            reportNewLocationReceived(location);
-        }
+        reportNewLocationReceived(location);
```
crankycoder, line 217: @garvankeeley just want to make sure I got this right - we should always report a new location now right?

src/org/mozilla/mozstumbler/service/stumblerthread/scanners/ScanManager.java
```diff
@@ -178,13 +178,4 @@ public Location getLocation() {
         return (mGPSScanner == null)? new Location("null") : mGPSScanner.getLocation();
     }
 
-    public void checkPrefs() {
```
crankycoder, line 181: All the checkPrefs did was fiddle with the geofencing, so all calls and methods were pulled out.

crankycoder:
@garvankeeley I've pulled the `LocationBlockList` back in, renamed it to `StumblerFilter` and renamed the `contains` method to `blockLocation` because that's what it actually does. :)

`checkPrefs` is still pulled out because after following the call chain, the only thing that happens is an eventual invocation of `LocationBlockList::updateBlocks` here: https://github.com/mozilla/MozStumbler/pull/816/files#diff-1f7fbb934196c505ddf6deeb96e631acL30

which only does things related to geofencing.

crankycoder:
@garvankeeley r?

garvankeeley:
Thanks, looks good, hopefully we can get geofencing back in as a UI-based feature after 1.0.

Issue #765, opened 2014-08-19
Don't draw line connecting discontiguous tracks


cpeterso:
I turned stumbling on and off during a road trip. The map view drew long straight lines crossing the desert to connect discontiguous tracks that were many miles and hours apart. When drawing tracks, the map should not connect measurements that are more than X meters or Y minutes apart.

Djfe:
Same issue I noticed and reported here: #733

It's good to have an independent issue now to keep track of it :)

garvankeeley:
We will likely change to just drawing collection points, and not joining the lines

crankycoder:
Closed as of #817

Issue #763, opened 2014-08-13
the GPS "glitches out" at the beginning of the scanning


Djfe:
Since we are now able to see the track MozStumbler records:
You can see in the following screenshot that the first scans at the beginning are garbage since the GPS values need to plateau first once the device get a GPS fix

are other people experiencing this, too?

if so, then there needs to be a timeout before a GPS position is considered valid for use

or mozstumbler would need to check functions like http://developer.android.com/reference/android/location/Location.html#getAccuracy%28%29 to determine whether the accuracy of the current location is accurate enough

of course the accuracy could also be lower if your device is covered by something or if it is in your jeans/backpack.

In this screenshot you can see what I' talking about:
![glitch](https://cloud.githubusercontent.com/assets/5685512/3907072/713e3182-22f8-11e4-81f7-19c012a12f9b.png)

the red line is where I moved approximately

garvankeeley:
I think this approach would fix it:
- show the current gps as a dot, and connect the last point in the track to this dot
- only store track positions that are used for stumbler reports. When we actually show the track vertices as dots, this is a visual indicator to the user of where reports were made, and how frequently.

crankycoder:
Closing this as #817 was merged

Djfe:
Does #817 really fix this issue or does it just obscure it by only showing the points and not connecting them?

the real issue here is that Mozstumbler should ignore faulty GPS positions at the beginning of the scan/ignore faulty positions in general (ignore too unaccurate outliers)

garvankeeley:
Sorry, I saw this as a UI bug, but it is actually a observation accuracy issue. 
I do see some wandering but the reported accuracy of the GPS still looks reasonable (by looking at the accuracy radius), so I don't see an obvious way to filter this.

crankycoder:
I'm closing this as waiting for better GPS fixes doesn't help us in the context of collecting data for Ichnaea.  Waiting just means we aren't collecting data.  We might as well just start collecting right away.

Issue #750, opened 2014-08-01
Marker for position is too big


Gitschubser:
Hmm, i think the marker for the position could be smaller.

garvankeeley:
On my screen it is 25% smaller than the marker on google maps, can you screenshot it? Maybe there are screen size issues we need to consider

Djfe:
The way it looks on my screen:
![2014-08-01 22 48 41](https://cloud.githubusercontent.com/assets/5685512/3787976/14938c60-1a48-11e4-8015-d7a1aecfdc49.png)

keverets:
Likely due to drawing based on px rather than dp.

Djfe:
even on my galaxy tab 3 10.1 (1280x720px) the pointer is bigger than the pointer of Google Maps
the track could be thinner, too in my opinion
and the pointer and the track are not transparent, which will hopefully change in the future so that you are able to read street names below the track and and the pointer

garvankeeley:
I believe this to be fixed by #817

Issue #733, opened 2014-07-28
MapActivity: show my gps track, and equivalent MLS track


garvankeeley:
Show the gps track so I know where I have been. Also show the MLS track so I can see the value of my contribution (and I can see where we have greater error and need more stumbling). With repeated stumbling I should see the MLS track improve over time, which is a valuable indicator. 
Also, Firefox OS QA can use this information to better understand MLS location error.

Sorry, my colours are swapped on my drawing,  the GPS track should be blue.

![screen2](https://cloud.githubusercontent.com/assets/5495083/3716336/b2043f00-1609-11e4-8b25-486aebc42a00.jpeg)

Djfe:
I noticed a bug in 0.21.0:
![screenshot](https://cloud.githubusercontent.com/assets/5685512/3906737/c57d532a-22f5-11e4-96de-cc3595b9d1f7.png)

The map page only loads current points and connects them -> it is possible that it misses some if the devices screen was turned off
->it will connect the last and the next point and doesn't care about the data that was collected inbetween

if you reload the map page it rebuilds the complete track so that the correct track is shown

garvankeeley:
Great report, with this info it will be easy to fix

crankycoder:
Closed as of #817 landing

PR #817, opened 2014-09-10
Map view work: GPS location, Observation points, and MLS locations

garvankeeley:
This is the appearance showing the GPS location in blue, the observation points in green, and the MLS locations in red (attached to their parent observation points with a line).
Also shown is the "locate me" button in the lower left.

screenshot_2014-09-09-19-35-55

The map view fixes addresses issues: #733, #750, #763, #765, #706

The instance count violation was originally issue #7, it recurs from time to time as we accidentally add code that keeps a reference to an activity.


garvankeeley:
Follow up bugs: 
- animate the locate me button on tap
- credit in about dialog for 'locate me' icon from AndroidAssetStudio (Apache License): http://romannurik.github.io/AndroidAssetStudio
- Limit the number of observation points, on an old device I estimate that performance starts to suffer above 250 points
- Draw the observation and MLS points to a raster tile, right now the vectors are redrawn every time.

Djfe:
Could you make those dots transparent, please? I would like to be able to see the street through everything
What do you think about placing the locate me button in the task bar in the top-right corner?

garvankeeley:
Just making the dots transparent isn't enough, if objects are overlaid, transparencies will sum and obscure data. Moreover, for my eyes, things would have to be highly transparent before I could make out data underneath. Our map style is a bigger issue for me, I can't make out details on it like street names and one-way indicators.
ATM I am just making sure the data is displayed and the code underneath supports these UI refinements in future (hopefully by a visual designer).

Addendum: This does transparent dots with similar data to stumbler, might be something we can copy:
http://www.neteffmon.com/?zoom=17&lat=48.87346&lon=2.34396&layers=B0TFF

Issue #792, opened 2014-09-02
UI Redesign: Start/Stop button on Map Activity


garvankeeley:
The redesign of the app is to make the map the main screen, therefore it needs a start/stop scanning button.

PR #832, opened 2014-09-13
Add Play/Pause to the menu on the map

keverets:
Also, don't always unpause scanning when entering the map.


keverets:
This fixes #792

garvankeeley:
Thanks again @keverets!

PR #833, opened 2014-09-13
Place the marker for the current point over top of the previous points

keverets:
Without this, the current position can get swamped by the previous points
when zoomed out.

### DIFF
--- a/src/org/mozilla/mozstumbler/client/mapview/MapActivity.java
+++ b/src/org/mozilla/mozstumbler/client/mapview/MapActivity.java
@@ -183,11 +183,11 @@ public final class MapActivity extends Activity {
             mMap.getOverlays().add(mCoverageTilesOverlay);
         }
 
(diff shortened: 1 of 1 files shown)


garvankeeley:
I wanted to emphasize the importance of the observation points, so I made them the top layer. Can you screenshot what you are seeing? 
I did try this out but I can't see the observation point being collected as I am walking, I have to move off the dot. And because the app doesn't zoom in very high (making it hard to 'move off the dot'), sometimes I am standing and wondering if a collection was made.

keverets:
I'll see if I can get a good screenshot of what I'm seeing today.

I'd like the green historical dots to be below the blue dot (since the blue
dot is more important to me and can become obscured by the green dots), but
the red MLS returned locations to be above the blue dot (so I can see how I
compare to MLS, even when it's an exact match).

This will take a bit of thinking.

garvankeeley:
I'll close this for now, with some more user feedback we can decide how to best display this

PR #834, opened 2014-09-13
Fix auto-follow

keverets:
The map no longer re-centers on the current location, making the user constantly press the centerMe button as they move.  After this change, pressing the centerMe button will keep centering until the user manually pans at which point it will stop following until they press the centerMe button again.


garvankeeley:
Thanks @keverets! This is an important detail that we missed.

Issue #773, opened 2014-08-22
MapQuest copyright on mapview (only on non mozilla builds)


garvankeeley:
issue #757 is for OSM copyright. We are switching to MapQuest's free tile service for non-mozilla builds.

This follows from issue #770

crankycoder:
Closed as of #835

PR #835, opened 2014-09-13
Add copyright area to the map, and populate it as requested when using MapQuest tiles

keverets:
This is a fix for issue #773


garvankeeley:
Thanks, verified it is on when source tile is MapQuest, off with MapBox. 
(I am surprised we don't need to attribute the latter on the map itself, I should verify that)

Issue #830, opened 2014-09-12
Map view performance: limit the number of observations points drawn


garvankeeley:
We can just draw observation points with no upper limit, on older devices this could bog them down. 

@keverets suggestion is to time bound the drawing code. This would be the simplest approach to start with. The points are stored in order, so we could iterate from newest to oldest, and stop after 100ms (or something similar).

PR #837, opened 2014-09-13
Limit ObservationPointsOverlay drawing to 150 ms

keverets:
Too many points, and the drawing can cause lag on slower devices.  This
will limit the impact.

Performance testing should be done to determine appropriate values for
the maximum duration, as well as how often to check the time.

This is a fix for #830


garvankeeley:
Thanks Kevin, this will prevent this feature from breaking older droids.

Issue #629, opened 2014-06-16
Unit Testing -Help needed!


garvankeeley:
The code base is growing and we have zero unit testing. Retrofitting this is not easy, so going forward we should be unit testing as we commit.
Any tests will be accepted, unit or functional. There is no code coverage requirement, just hit the major use cases.

PR #839, opened 2014-09-13
Bump Android build tools and Android support library to version 20

cpeterso:


crankycoder:
Why are we bumping to add support for Android Wear?

Djfe:
Android wear would provide a great way for checking the current scan stats!
Pretty cool use case in my opinion.

cpeterso:
I bumped the Android build tools and Android support library to their latest versions (simply because they are available), but I did not bump the Android API level. We're still using API level 19 (KK 4.4.2), not level 20 (Android 4.4W or L Preview).

crankycoder:
I would like to defer this change at least until the unittest changes for #629 go in as the gradle build is going to change pretty substantially.  I'm not particularly interested in having people change their build requirements when the most recent API we actually use is API 18 and we have no code that actually requires bumping the support library version up.

cpeterso:
OK

PR #840, opened 2014-09-14
Don't check for app updates if Wi-Fi is disconnected and "Upload only on WiFi" setting is true

### LINKED ISSUE #766: Auto-updater should honor "Only upload over Wi-Fi" setting
The auto-updater will currently try to download new updates from GitHub over a cell connection, even if the user prefers to only upload data over Wi-Fi.

cpeterso:
- Fixes #766
- Minor clean up of MainActivity.java


garvankeeley:
Thanks Chris, LGTM. 
Android Studio will auto-alphabetize and strip unused imports, might be time for me to run that again project-wide.

PR #845, opened 2014-09-15
added SimpleKML and SimpleXML dependency

### LINKED ISSUE #844: Hookup SimpleKML
First step in data export (and import?) is to have this lib in the codebase: 
https://github.com/Ekito/Simple-KML

crankycoder:
This fixes #844 to add a dependency to SimpleKML

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -83,6 +83,7 @@ android {
     }
 
     lintOptions {
+        disable 'MissingTranslation'
         abortOnError false
     }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -27,7 +27,6 @@ libDir.mkdirs()
 task updateJars(type: Download) {
     src([
         'https://github.com/mozilla/osmdroid/releases/download/4.3pre0/osmdroid-android-4.3-SNAPSHOT.jar',
-        'https://github.com/crankycoder/SimpleKML/raw/master/SimpleKML/libs/simple-xml-2.6.3.jar',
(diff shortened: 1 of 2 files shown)


crankycoder:
Note that this also adds: com.android.support:appcompat-v7 as a dependency.

From https://developer.android.com/tools/support-library/features.html : 

`In general, we recommend including the v4 support and v7 appcompat libraries, because they support a wide range of Android versions and provide APIs for recommended user interface patterns.`

build.gradle
```diff
@@ -75,8 +83,8 @@ android {
     }
 
     lintOptions {
-        abortOnError true
         disable 'MissingTranslation'
+        abortOnError false
```
crankycoder, line 87: I've disabled lint error aborts because of the dependency on simplexml.

simplexml relies on javax.xml.  Those libraries are contained in the `stax-api` which have a dependency on `javax.xml.stream` which isn't in Android.  

This is 'ok' for us, as simplexml never uses javax.stream anyway.  Annoying though.  Oh java.

proguard.cfg
```diff
@@ -44,3 +44,5 @@
 -keepnames class * implements android.os.Parcelable {
     public static final ** CREATOR;
 }
+
+-libraryjars <java.home>/lib/rt.jar(java/**,javax/**)
```
crankycoder, line 48: This is the magical incantation required to get proguard to not throw a hissy fit at us because of javax.xml

crankycoder:
@garvankeeley  r?

garvankeeley:
Tried it out, looks good! Will code something up that uses this today

Issue #842, opened 2014-09-15
Map view only draws observation points when the map view is open and screen is on


cpeterso:
This is frustrating because I will shut off my screen when I'm driving and, when I reach my destination, I have no intermediate points.

garvankeeley:
Thanks for the report, confirmed here also.

Note to self, hook this up properly in ObservedLocationsReceiver.java

PR #848, opened 2014-09-16
Fix show observation points with the screen off

garvankeeley:
Issue #842.

Fixes screen off case, and the case where the map activity has not been shown yet in the app's lifetime.

This code (the previous PR) was written with the intention that the map activity is the main activity -which will be the case soon- but it complicated this hookup. I can simplify things a bit when we refactor out the current main activity.


garvankeeley:
r? vng.

src/org/mozilla/mozstumbler/client/mapview/ObservedLocationsReceiver.java
```diff
@@ -12,45 +12,38 @@
...
 import java.util.LinkedList;
 
-class ObservedLocationsReceiver extends BroadcastReceiver {
-    ArrayList<GeoPoint> mLocations = new ArrayList<GeoPoint>();
+public class ObservedLocationsReceiver extends BroadcastReceiver {
+    private static final String LOG_TAG = ObservedLocationsReceiver.class.getSimpleName();
     private WeakReference<MapActivity> mMapActivity = new WeakReference<MapActivity>(null);
     private LinkedList<ObservationPoint> mCollectionPoints = new LinkedList<ObservationPoint>();
     private LinkedList<ObservationPoint> mQueuedForMLS = new LinkedList<ObservationPoint>();
+    private final LinkedList<ObservationPoint> mPointsMissedByMapActivity = new LinkedList<ObservationPoint>();
```
crankycoder, line 30: Only change I'd suggest is to use `mPointsMissedByMapActivity`, `mCollectionPoints,` and `mCollectionPoints` as FIFO queues using the LinkedLink push() and pop().

garvankeeley:
updated in issue #875

PR #854, opened 2014-09-17
Use the spelling "Wi-Fi" in the UI and fix some lint resource warnings

cpeterso:
Android uses the spelling "Wi-Fi", with a hyphen and capital letters, regardless of the system locale. Our app UI should be consistent with the system UI conventions. (Fun fact: some iOS sold in China use the term "WLAN" instead of "Wi-Fi".) I didn't change any code identifiers, so they, like the Android's APIs, use a mix of "WiFi", "Wifi", and "wifi". :)

I also fixed a bunch of lint warnings about unused resources and string formats.


garvankeeley:
Thanks Chris!

Issue #862, opened 2014-09-19
Get rid of the Updater code for the 1.0 release


crankycoder:
The Play store will handle code updates of versions for us, no need to run our own updating code.

lua86:
I do not think it whould be a good idea, as some people do not want google
play store.

crankycoder:
@lua86 are you talking about the F-droid release?  There's a related issue with releasing the MozStumbler with unofficial Mozilla builds - the unoffical builds do not use our much faster performing tile server for maps.  

@garvank I've made pretty big changes to the build system as of landing #864.  We could split the builds into different 'flavors'.  The change would amount to adding an extra buildConfigField and having a guard around using the Updater code.

Thoughts?

cpeterso:
@crankycoder: I think @lua86 is referring to the actual Google Play Store. @hannosch says that only 30% of MozStumbler users are using our GitHub builds, so the other 70% are likely using F-Droid or their own local builds.

I think we should follow our users' lead and:
- Remove our GitHub-based auto-updater.
- Use F-Droid to publish frequent test releases of our Free Software build flavor (no Google Play Services, no MapBox, etc). F-Droid watches our GitHub repo for new release tags and auto-builds their package from our repo. I don't think we need to do anything.
- Use Google Play Store to publish stable snapshots of our "full-featured" build flavor (that uses Google Play Services, MapBox, etc).

elkos:
Please in case of removing the auto-updater could there be a viable way to prompt the use to either install Mozstumbler from F-Droid or Google Play? I think that some of our users (for example our top3 data contributors) might miss a crucial update ;)

Djfe:
Pls use Google groups to distribute alphas/betas through the play store:
https://support.google.com/googleplay/android-developer/answer/3131213

Facebook practices this as well.

crankycoder:
@elkos thanks,  that's a good point.  :-) 

We'll keep the updater code after all.  I'll tweak the build to conditionally check for updates if you get a github release. 

@cpeterson, do we have an fdroid contact as the new build is pretty different now?

crankycoder:
Disable the updater code unless we have a release flavour set to github.

Djfe:
there should be something like an information that an update is available even in the Google Play Store version -> it should link to the Google Play Store page so that users can update

something like:
Update | Stay on version xy

other apps like Whatsapp do this as well
not everyone has autoupdates turned on

Issue #826, opened 2014-09-12
Updater Unit Test: update Mozilla builds only


garvankeeley:
MainActivity onCreate has this:
 if (BuildConfig.MOZILLA_API_KEY != null) {
            Updater.checkForUpdates(this);
 }

This would be a good item to unit test so it doesn't get broken

crankycoder:
I'm going to add this into my fix for #629

PR #864, opened 2014-09-19
Features/629+826 unittest mozbuild updates

### LINKED ISSUE #629: Unit Testing -Help needed!

crankycoder:
This fixes #629 to add JUnit, robolectric and jococo coverage analysis
#826 needed a unittest around the Updater code, but that's actually going away because of the play store, so this actually closes off #862 as well.

I need to add some documentation for this, but the short of it is that the Makefile has reasonable targets.

The default `all` target will build a release.
`test` will compile and run unittests

There is now a split between debug and unittest builds as robolectric puts extra constraints on what is possible.


android-library/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest
+	package="cz.posvic.bootstrap.lib"
```
crankycoder, line 3: Rename this to org.mozilla.mozstumbler

android-library/src/main/res/menu/main.xml
```diff
@@ -0,0 +1,11 @@
+<menu xmlns:android="http://schemas.android.com/apk/res/android"
+	xmlns:app="http://schemas.android.com/apk/res-auto"
+	xmlns:tools="http://schemas.android.com/tools"
+	tools:context="cz.posvic.bootstrap.lib.LibraryActivity">
```
crankycoder, line 4: rename this to org.mozilla.mozstumbler as well.  just grep for the string and replace all instances.

crankycoder:
@garvankeeley I'm just waiting for travis to get back before merging this.

garvankeeley:
On a fresh pull I am getting modifications locally:
On branch dev
Your branch is up-to-date with 'origin/dev'.

```
deleted:    MozStumbler.iml
modified:   android-library/android-library.iml
modified:   android/android.iml
```

garvankeeley:
Will this commit also break the updater system? 
android/src/main/java/org/mozilla/mozstumbler/client/Updater.java 29:  
private static final String VERSION_URL = "https://raw.github.com/mozilla/MozStumbler/master/VERSION";

garvankeeley:
And the app icon is gone.

crankycoder:
Let's change the VERSION_URL to point to something that _isn't_ in source control.  If we accidentally push a broken change into source control, I don't want to deal with hundreds of broken clients.

Let's have it point to some github wiki page instead.  I've opened ticket #865 to review and rework the Updater code.

crankycoder:
I still have the real app icon, maybe a stale build?  I'll double check.  Those pngs I dropped look like the stock generic android robot icons.

crankycoder:
@garvankeeley The icon seems to be ok for me - I've deleted all my build directories and reconstructed the APK and I'm still getting the new-ish stumbler icon.  The icon is referenced from the AndroidManifest.xml file here:

https://github.com/mozilla/MozStumbler/blob/dev/android/src/main/AndroidManifest.xml#L29

Which should just pull the proper ic_launcher.png file from the resources directory.

crankycoder:
VERSION file re-added in #871

garvankeeley:
I got the icon to work once I copied it to the XDPI resources.

Stripping the icon from android-library project would be good to avoid confusion.

PR #870, opened 2014-09-22
Added Mercedes in-car WiFi to blocklist

JensErat:


crankycoder:
Is the SSID configurable on these things?  The doc at:

http://www.crownautomobile.com/clientuploads/specials/service/wifi-faq.pdf

mentions that the hotspot has an SSID with `Autonet` in it.

@garvankeeley - I think we should add some support in the stumbler to keep tabs of a bounding box for any given BSSID or cell tower. That would at least give us a hint that something.

JensErat:
The "Autonet" SSID seems to be some US located service, I observed the SSID in Germany. [Other reports on "MB WLAN" can be observed](https://duckduckgo.com/?q=%22MB+WLAN%22+mercedes), might be configurable, but seems to be the default.

crankycoder:
Thanks @JensErat!

PR #878, opened 2014-09-22
Features/849 email address

### LINKED ISSUE #849: Leaderboard: capture email addresses
Although we would like to have Fx accounts integration, we can't get that for 1.0. We should as a stopgap, grab email addresses.

crankycoder:
Adds an email address field and header when uploading the to ichnaea.
Fixes #849 on the client, needs support on ichnaea now.


crankycoder:
https://github.com/crankycoder/MozStumbler/tree/features/preference-unittests has the start of unit tests, waiting for follow up from the robolectric list.

crankycoder:
@garvankeeley r?

crankycoder:
Hrm.  I think we should at least minimally hash the email address with SHA1.  I don't think we really care about salting the value, but this would at least prevent the server from ever accidentally leaking email addresses.  

We only care about making sure that individual MozReps can get their statistics out, and this is only a stop gap until Galaxy+FxA is ready to roll.

Thoughts?

elkos:
I know I am too late into this conversation but would it make sense to use Personas for email authentication... is it's a huge overkill?

crankycoder:
We're targetting Firefox Accounts going forward, but we need something in
the short term to enable some stumbling 'parties'.

PR #894, opened 2014-09-29
Add a message to the map view if no map tiles are available

garvankeeley:
If either no network, or no SD card, notify the user no map tiles are shown, but stumbling is ok.

This is the no network message (there is another message for no sd card storage available):
screenshot_2014-09-29-13-36-27


crankycoder:
This should be tested on a Samsung GT-S5670L (Galaxy Fit) as per #918

PR #910, opened 2014-10-01
Performance bug with hardware accel + drawPath() 

garvankeeley:
I use drawPath to draw triangles for cell-only collection indicator.
With as little as 300 primitives the drawing starts to jank heavily.
Turns out drawPath() is poorly supported:
http://developer.android.com/guide/topics/graphics/hardware-accel.html
Changed code to use just draw rect and circle, and tested with 3000
primitives.
Also, lowered the point overlay drawing timeout even further to 30 ms, this only
represents part of the drawing pipeline, there is still raster overlays, so we should be
aggressive about stopping the drawing early.

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/mapview/ObservationPointsOverlay.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/mapview/ObservationPointsOverlay.java
@@ -8,11 +8,12 @@ import android.content.Context;
 import android.graphics.Canvas;
 import android.graphics.Color;
 import android.graphics.Paint;
-import android.graphics.Path;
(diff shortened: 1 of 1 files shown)


crankycoder:
Sweet.  Testing this tonight as I did a long stumble and started to see enough lag that android tried killing the stumbler several times.

garvankeeley:
Did you have the additional map layer turned on? If you only had the green dot layer, then it would be a different bug.

crankycoder:
I had the shaped does showing up.  I always have that on when I do stumbles.

garvankeeley:
Shockingly, drawing a triangle was ambitious.

PR #911, opened 2014-10-01
Reporter/HttpUtil cleanup items for easier testing/debugging

garvankeeley:
Reporter/HttpUtil cleanup
While debugging MLSLocationGetter I ran into a few cleanup issues
- wasSent() doesn't seem to serve a purpose, removed
- changed some Log.e(...) to be more concise as the code I was debugging
  was spamming the log when throwing exceptions
- added errorMLSResponse on MLSLocationGetter to clear the state after
  an error
- In Reporter.java, I added a check for zero wifi and cell count, it is
  hard to repro, it happens when I modify the code for load testing (as I
  typically do), but it helps prevent false errors when testing at the
  very least.

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/mapview/ObservationPoint.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/mapview/ObservationPoint.java
@@ -12,6 +12,7 @@ import org.mozilla.mozstumbler.client.ClientPrefs;
 import org.mozilla.mozstumbler.service.utils.NetworkInfo;
 import org.mozilla.mozstumbler.service.stumblerthread.datahandling.DataStorageContract;
 import org.osmdroid.util.GeoPoint;
+import org.mozilla.mozstumbler.service.core.logging.Log;
 
 public class ObservationPoint implements MLSLocationGetter.MLSLocationGetterCallback {
     public GeoPoint pointGPS;
@@ -57,5 +58,9 @@ public class ObservationPoint implements MLSLocationGetter.MLSLocationGetterCall
             mMLSLocationGetter = null;
             pointMLS = new GeoPoint(location);
         }
+    }
+    public void errorMLSResponse() {
+        Log.i(ObservationPoint.class.getSimpleName(), "Error:" + mMLSQuery.toString());
+        mMLSLocationGetter = null;
     }
 }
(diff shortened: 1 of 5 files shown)


garvankeeley:
@crankycoder r?
Sorry for the miscellany-type PR.

android/src/main/java/org/mozilla/mozstumbler/client/mapview/ObservationPoint.java
```diff
@@ -58,4 +59,8 @@ public void setMLSResponseLocation(Location location) {
             pointMLS = new GeoPoint(location);
         }
     }
+    public void errorMLSResponse() {
+        Log.i(ObservationPoint.class.getSimpleName(), "Error:" + mMLSQuery.toString());
+        mMLSLocationGetter = null;
```
crankycoder, line 64: Why do we care about setting mMMLSLocationGetter to null?

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/Reporter.java
```diff
@@ -250,6 +247,9 @@ public synchronized void flush() {
             Log.d(LOG_TAG, "Received bundle: " + mlsObj.toString());
         }
 
+        if (wifiCount + cellCount < 1)
```
crankycoder: Doh.  thanks.

crankycoder: Nit. Can we add braces here?

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/datahandling/StumblerBundle.java
```diff
@@ -27,12 +27,6 @@
     private final Map<String, ScanResult> mWifiData;
     private final Map<String, CellInfo> mCellData;
 
-    public void wasSent() {
```
crankycoder, line 30: Thanks.  This was left overs from the old HTTP code.

crankycoder:
r+

android/src/main/java/org/mozilla/mozstumbler/client/mapview/ObservationPoint.java
garvankeeley, line 64: After a failure, the ObservationPoint will hold an instance of mMLSLocationGetter, which is not re-usable (it is an asynctask), so it should be set to null.

PR #916, opened 2014-10-01
KML I/O added in Developer Menu

garvankeeley:
This adds KML import and export. 

[In debug-only, if you long-press the version number in the about dialog, you get the KML I/O dialog.] EDIT: no longer true.

I have been using this feature to export data for verification, and to load test 1000s of points into the map view for load testing. The feature is not ready for prime-time, as the UI needs more work, but other than that it is functional.

I don't want to keep merging this locally for testing, so it is best to get it in the codebase.

@crankycoder r?


android/src/main/java/org/mozilla/mozstumbler/client/serialize/ObservationPointSerializer.java
```diff
@@ -0,0 +1,332 @@
...
+                if (isGps) {
+                    points.add(observationPoint);
+                    gpsList.put(placemark.getId(), observationPoint);
+                } else {
+                    mlsList.put(placemark.getId(), coordinate);
+                }
+            }
+        }
+
+        Iterator it = mlsList.entrySet().iterator();
```
crankycoder: This should be typed.

android/src/main/java/org/mozilla/mozstumbler/client/serialize/ObservationPointSerializer.java
```diff
@@ -0,0 +1,332 @@
...
+        while (it.hasNext()) {
+            Map.Entry pairs = (Map.Entry)it.next();
```
crankycoder: And this should also be typed.

crankycoder:
This branch isn't compiling for me when I grab it from your repo, can you take a look at it?

crankycoder:
@garvankeeley  - finished my review, minor nits.

garvankeeley:
Now it is available though a button ''Developer" on the Settings screen.

PR #923, opened 2014-10-02
Automate a bunch of build problems

crankycoder:
- automated the versionName and versionCode nonsense.
- write out the VERSION file while we're updating the AndroidManifest.xml
- dynamically pull in rootProject/private.properties to load in
  customized parameters without clobbering parallel, daemonized
  compilation
- added a release_check.sh script to check the existence of a
  private.properties file prior to compilation of a release
- added release_check as a Makefile target and dependency to release


garvankeeley:
To proguard.cfg, add:
-dontwarn org.joda.time.**

To makefile, change:
rename_release.sh to
sh rename_release.sh

And the version number thing

r+ after that, because it worked!

PR #924, opened 2014-10-02
queued MLS locations were being removed too early

garvankeeley:
I think i moved a line during a rebase, the queued MLS locations were being removed from the queue too early. (one line was out of place, fixed).


crankycoder:
Isn't the previous version correct?  Shouldn't you remove the last item no matter what?

crankycoder:
r+ after clarification over vidyo.  This isn't really a queue, it's a bag of things that need to be queried against MLS.

garvankeeley:
Sorry queue was a bad name, will fix at some point.

Issue #930, opened 2014-10-02
700 observations stuck in queue


kbrosnan:
Using 1.0, pressing the upload button does not upload the data.

garvankeeley:
Yikes. Can you check the Settings>Log for messages? 

@crankycoder, what build tag is this? I'll take a look and see if something was unhooked in that build.

kbrosnan:
The log is blank. Tried pressing the upload button and re-checking the log and it is still blank.

garvankeeley:
Can you try toggle the scanning on and off and see what happens? This should also trigger an upload (although this case will respect the Settings>Auto-upload only on Wi-Fi)

crankycoder:
@kbrosnan Which build is this?  We've never released a 1.0 version yet.

kbrosnan:
Via dropbox from Karl.

kbrosnan:
This has grown to over 1300 in queue.

garvankeeley:
Should be fixed now.

kbrosnan:
I have not seen this on v0.40.0, seemed fairly common on the build QE was handed.

PR #937, opened 2014-10-07
Empty content body not handled after successful post.

garvankeeley:
This would be my guess as to what is causing this also: https://github.com/mozilla/MozStumbler/issues/930


garvankeeley:
@crankycoder r?

PR #939, opened 2014-10-07
Created a developer settings screen. 

garvankeeley:
Previously the KML Activity was being used for this. Currently it only has the KML fragment.


garvankeeley:
@crankycoder just some fluffy GUI stuff, can I just push

Issue #921, opened 2014-10-02
Add pref for turning on the "Show MLS locations" switch


garvankeeley:
As it stands now, even with the MLS locations off, extra allocations are taking place. There is no quick way to avoid this, as the JSON bundles need to be stored in memory to query MLS. It would be best to be safe and have this feature off by default. This feature should be stress-tested for memory and network usage before it gets removed as a pref (and permanently on).

The pref would be:

Enable show MLS locations on map [x]
You'll be able to toggle this layer on/off in the 'Map layers' options, but increases memory and network usage.

PR #941, opened 2014-10-07
MLS location layer requires additional memory and network.

garvankeeley:
Make this layer toggle an option that defaults to false.

Issue #921

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/ClientPrefs.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/ClientPrefs.java
@@ -12,7 +12,7 @@ public class ClientPrefs extends Prefs {
     private static final String LON_PREF = "lon";
     private static final String IS_FIRST_RUN = "isfirstrun";
     public static final String KEEP_SCREEN_ON_PREF = "keep_screen_on";
-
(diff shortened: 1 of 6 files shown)


android/src/main/java/org/mozilla/mozstumbler/client/ClientPrefs.java
```diff
@@ -12,7 +12,7 @@
     private static final String LON_PREF = "lon";
     private static final String IS_FIRST_RUN = "isfirstrun";
     public static final String KEEP_SCREEN_ON_PREF = "keep_screen_on";
-
+    public static final String ENABLE_OPTION_TO_SHOW_MLS_ON_MAP = "enable_show_mls_on_map";
```
crankycoder, line 15: Can you rename `ON_MAP_SHOW_MLS_PREF` to `SHOW_MLS_ON_MAP` and `ENABLE_SHOW_MLS_ON_MAP_PREF` to clarify that one is a flag to show MLS locations, and the other is a preference to toggle the availability of the flag.

crankycoder:
r+ after renaming the constants.

garvankeeley:
I have another PR in my queue which renames this pref

Issue #938, opened 2014-10-07
mozilla api key not being set


garvankeeley:
Prefs.getInstance().setMozApiKey() is never called

crankycoder:
merged in #945

PR #945, opened 2014-10-07
Features/938 missing api key

crankycoder:
This should close out #938 and set the mozilla api key as a pref on MainApp creation.

I've also cleaned up some of the logging so that HTTPResponse doesn't spam the GUI log.


garvankeeley:
r+

PR #961, opened 2014-10-08
Making main app the global upload progress listener.

garvankeeley:
- AsyncUploader now has a static upload listener, this is set once at
  MainApp startup to the MainApp
- also more AsyncUploadListener as sub interface of AsyncUploader, this
  is just one function.


garvankeeley:
@crankycoder r?

Issue #963, opened 2014-10-08
When switching OSMDroid tile providers, we get receiver leakage


garvankeeley:
E/ActivityThread(  802): Activity org.mozilla.mozstumbler.client.navdrawer.MainDrawerActivity has leaked IntentReceiver org.osmdroid.tileprovider.modules.MapTileFileStorageProviderBase$MyBroadcastReceiver@407439d8 that was originally registered here. Are you missing a call to unregisterReceiver()?
E/ActivityThread(  802): android.app.IntentReceiverLeaked: Activity org.mozilla.mozstumbler.client.navdrawer.MainDrawerActivity has leaked IntentReceiver org.osmdroid.tileprovider.modules.MapTileFileStorageProviderBase$MyBroadcastReceiver@407439d8 that was originally registered here. Are you missing a call to unregisterReceiver()?
E/ActivityThread(  802):    at org.mozilla.mozstumbler.client.mapview.tiles.AbstractMapOverlay.<init>(AbstractMapOverlay.java:39)
E/ActivityThread(  802):    at org.mozilla.mozstumbler.client.mapview.tiles.CoverageOverlay.<init>(CoverageOverlay.java:20)
E/ActivityThread(  802):    at org.mozilla.mozstumbler.client.mapview.MapActivity.mapNetworkConnectionChanged(MapActivity.java:335)
E/ActivityThread(  802):    at org.mozilla.mozstumbler.client.mapview.MapActivity.access$900(MapActivity.java:62)
E/ActivityThread(  802):    at org.mozilla.mozstumbler.client.mapview.MapActivity$5.onReceive(MapActivity.java:341)

garvankeeley:
From log in Issue #869

Issue #953, opened 2014-10-08
Toggling "Keep screen on while in map" needs restart to take effect


cascheberg:
Toggling "Keep screen on while in map" only takes effect after restarting the app (v0.40.0.0)

crankycoder:
Merged in #964

Issue #950, opened 2014-10-07
Side pannel data is blank with don't keep activites


kbrosnan:
![screenshot_2014-10-07-15-09-51](https://cloud.githubusercontent.com/assets/250273/4551252/e24a0d78-4e6e-11e4-9c73-18631e8b2ee5.png)
- Enable the Android developer options
- Select the "Do not keep activities" option
- Launch MozStumbler
- open the sidebar note the values
- Switch apps or open the Settings or Leaderboard
- open the sidebar, no data is shown

crankycoder:
Merged in #964

Issue #960, opened 2014-10-08
UI Reporting is slightly off for 'This Session' KB


garvankeeley:
From left to right here is a sequence that shows the bug.

I started the app fresh (zero for all stats). The second screen shows the observations sent as 1.8 KB, but 'this session' is 1.4 KB. Screen 3 looks ok, and then screen 4 shows the Sent higher than this session again

![screenshot 2014-10-08 14 10 49](https://cloud.githubusercontent.com/assets/5495083/4564340/a974a622-4f16-11e4-98ae-003f4c489e2b.jpg)

garvankeeley:
Also, why in the first screen: no size reported for the 'ready to send' observations.

PR #964, opened 2014-10-09
Fix lifespan with "Don't keep activities" on

garvankeeley:
- This option simulates low memory states where non-visible activities are destroyed rather than backgrounded. Well-behaved apps should pass this test without issues.
- Required changing fragment creation, and properly handling onDestroy to
  avoid tons of warning messages in the logs
- _May_ be a serious issue on old devices, which actually do destroy activities more aggressively.
- Previously, because we didn't have proper lifecycle management, the main activity was set in the manifest to not re-create on orientation changed. Removed this requirement.

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -30,8 +30,6 @@
         <activity
             android:name=".client.navdrawer.MainDrawerActivity"
             android:label="@string/app_name"
-            android:launchMode="singleTask"
-            android:configChanges="orientation|screenSize|keyboard"
             >
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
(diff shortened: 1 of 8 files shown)


garvankeeley:
Fixes: https://github.com/mozilla/MozStumbler/issues/950

Also snuck in fix for: https://github.com/mozilla/MozStumbler/issues/960
Because it sort of relates to metrics view life cycle.

I think the changes to onDestroy fix https://github.com/mozilla/MozStumbler/issues/963

garvankeeley:
https://github.com/mozilla/MozStumbler/commit/8cce7355e7e0c375066c822c7f62b3f7728a68b0

Fixes https://github.com/mozilla/MozStumbler/issues/953

garvankeeley:
@crankycoder added back the launch mode, and changed the 'keep screen on' pref to not require restarting the activity

PR #966, opened 2014-10-09
Store osmdroid tiles in cache dir

### LINKED ISSUE #955: Store tiles in getExternalCacheDir() (with a fallback to getCacheDir())
It is not possible to clear the MozStumbler (v0.40.0.0) map cache without clearing all app settings, since map cache is listed as "Data".

mozstumbler

cascheberg:
Fixes #955

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/MainApp.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/MainApp.java
@@ -79,19 +79,10 @@ public class MainApp extends Application
     }
 
     private File getCacheDir(Context c) {
-        File dir = null;
-        if (Build.VERSION.SDK_INT >= 8) {
-            dir = c.getExternalCacheDir();
-        } else {
-            dir = c.getExternalFilesDir(null);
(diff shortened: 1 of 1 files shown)


garvankeeley:
Thanks for the PR! You can remove the sdk check as our min sdk is 10.
Is the external cache dir publicly accessible without rooting the device? I need that for development, as I inspect the tile dir regularly during development.

cascheberg:
I am not an experienced Android developer, so I just look things up.
The API description for `getExternalCacheDir` is pretty much the same as for `getExternalFilesDir`:

> There is no security enforced with these files. For example, any application holding WRITE_EXTERNAL_STORAGE can write to these files.

Though, `getExternalCacheDir` is not quite as nice as `getCacheDir` when it comes to cleaning up the space, [see API](http://developer.android.com/reference/android/content/Context.html#getExternalCacheDir%28%29)

garvankeeley:
getExternalCacheDir() isn't read protected, so this good, lets give it a shot

Issue #934, opened 2014-10-06
Crash reporting tool


garvankeeley:
https://github.com/ACRA/acralyzer is the one that I know of, I'll take anything.

crankycoder:
Merged in #967

PR #967, opened 2014-10-09
Features/934 acralyzer

crankycoder:
This should close off #934

Adds ACRA crash reporting and build changes required to pull in ACRA_URI, ACRA, USER and ACRA_PASS from private.properties.


garvankeeley:
r+

PR #969, opened 2014-10-09
add build and os info to user agent

garvankeeley:
Looks like this:
google/hammerhead/hammerhead:4.4.4/KTU84P/1227136:user/release-keys__MozStumbler/0.40.0.0

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/ClientPrefs.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/ClientPrefs.java
@@ -16,6 +16,8 @@ public class ClientPrefs extends Prefs {
     public static final String ENABLE_OPTION_TO_SHOW_MLS_ON_MAP = "enable_the_option_to_show_mls_on_map";
     private static final String ON_MAP_MLS_DRAW_IS_ON = "actually_draw_mls_dots_on_map";
 
+    public static final String ENABLE_SAVING_JSON_IN_KML = "save_json_in_kml";
+
     protected ClientPrefs(Context context) {
         super(context);
     }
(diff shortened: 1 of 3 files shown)


garvankeeley:
@hannosch any preference as to how this is formatted?

hannosch:
A typical Android example is: `"Dalvik/1.6.0 (Linux; U; Android 4.1.2; DROID RAZR Build/9.8.2O-72_VZW-16-5)"`. Knowing the Android version and device type is really useful. So I'd use something like this and append `MozStumbler/0.40.0` to it.

For example:

`"Dalvik/1.6.0 (Linux; U; Android 4.1.2; DROID RAZR Build/9.8.2O-72_VZW-16-5) MozStumbler/0.40.0"`

Edited: Use slash between product / version, as per http://tools.ietf.org/html/rfc7231#section-5.5.3

garvankeeley:
The device type is there 'hammerhead', the build '4.4.4/KTU84P/1227136', what part should I change? Is it the string formatting that should change?

Here are the available fields:
http://ec2-54-172-135-65.compute-1.amazonaws.com/acralyzer/_design/acralyzer/index.html#/report-details/stumbler/053af68c-9941-4807-9afd-d85b4cf967aa
No shortage of them :)

garvankeeley:
According to StackOverflow this will do it. Which would be MozStumbler-only due to the heavy weight approach.
String ua = new WebView(this).getSettings().getUserAgentString();

garvankeeley:
Fennec stumbler is like so:
Mozilla/5.0 (Android; Mobile; rv:35.0a1) Gecko/35.0a1 Firefox/35.0a1

I could append the "FINGERPRINT" string, which is like so:
motorola/falcon_retuglb/falcon_umts:4.4.3/KXB21.14-L1.32/33:user/release-keys

But it seems a tad verbose. I would be happy with device (falcon_umts) and version (4.4.3)

garvankeeley:
new pr opened

PR #984, opened 2014-10-10
#932 - Removes the XYTileSource and enables the MapBoxTileSource

### LINKED ISSUE #932: Use MapBoxTileSource for map box tile layer
OSMDroid has a built-in tile source where we just set the API key, and it looks better than the source we are using, as it has multiple content urls.

Requires a gradle change, so, hot potato to vng.

crankycoder:
This fixes #932.

The MapBoxTileSource randomly pulls from one of four different tile servers which should be more responsive.

The gradle build has been modified in a strange way, but it's been commented.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -91,6 +91,30 @@ android {
         unittest {
             buildConfigField "boolean", "ROBOLECTRIC", "true"
         }
+
+        applicationVariants.all { variant ->
+            variant.processManifest.doLast {
+                println("configuring AndroidManifest.xml");
+                copy {
+                    from("${buildDir}/intermediates/manifests") {
+                        include "${variant.dirName}/AndroidManifest.xml"
+                    }
+                    into("${buildDir}/intermediates/filtered_manifests")
+                }
+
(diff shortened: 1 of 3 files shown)


crankycoder:
@garvank - I'm on the fence about merging this. I don't see any dramatic improvement in performance and the URLs on the mapbox developer page indicate that we should be using the old URL.

I'll dive into the mapbox-android-sdk code to see what they're doing in there before we merge this.

garvankeeley:
Your points are good:
- no clear performance benefit
- added a bunch of gradle code
- the urls used in the osmdroid code may not be the latest urls

On the other hand the line where we instantiate the tile provider looks cleaner! 

LGTM, I'll leave it up to you.

crankycoder:
Mapbox's own SDK only uses the one URL, so I'm going to say "lets not do this".

https://github.com/mapbox/mapbox-android-sdk/blob/mb-pages/MapboxAndroidSDK/src/main/java/com/mapbox/mapboxsdk/constants/MapboxConstants.java#L12

PR #988, opened 2014-10-12
check if WiFi is mobile hotspot

cascheberg:
WiFi can be declared as mobile hotspot, you might want to check that case. I am not sure how `ConnectivityManager.isActiveNetworkMetered()` works for mobile data connections, so I leave the type comparison.

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/service/utils/NetworkInfo.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/service/utils/NetworkInfo.java
@@ -39,6 +39,7 @@ public final class NetworkInfo {
 
         if (Build.VERSION.SDK_INT >= 16) {
             if (mConnectivityManager.isActiveNetworkMetered()) {
+                Log.d(LOG_TAG, "Network is metered");
                 return false;
             }
         }
--- a/android/src/main/java/org/mozilla/mozstumbler/service/utils/NetworkInfo.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/service/utils/NetworkInfo.java
@@ -6,6 +6,7 @@ package org.mozilla.mozstumbler.service.utils;
 
 import android.content.Context;
 import android.net.ConnectivityManager;
+import android.os.Build;
 import android.util.Log;
 import org.mozilla.mozstumbler.service.AppGlobals;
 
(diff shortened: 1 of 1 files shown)


Djfe:
You know that this only checks whether your app should avoid too much network traffic or not, right?
It has nothing to do with whether the device shares it's mobile connection over wifi or not.

And if you want to see whether the used wifi connection is a hotspot of another android device:
I'm not sure whether this is even possible...

I think this method returns a value that depends on the settings made on the settings page that shows the amount of internet traffic as a graph. There is a point where you can select to avoid too much background network traffic on your mobile connection.

cascheberg:
You can tag WiFi connections as mobile hotspots, see [here](http://www.androidcentral.com/how-tag-wifi-access-points-hotspots-your-android-device), which should influence the return value.

I think the option to restrict background network traffic is forced and does not need to be checked / cannot be bypassed by apps.

Djfe:
`if (Build.VERSION.SDK_INT >= 16 && mConnectivityManager.isActiveNetworkMetered()) {
Log.e(LOG_TAG, "Network is metered!");
return false;
}`
Maybe a log entry for debugging. Could be helpful, if a user complains, that uploading over wifi doesn't work.
And I think you could put the arguments in one line.
If the first argument is false the second one is skipped anyways, so this shouldn't cause any issues on devices using Android API15 and lower, I think.

cascheberg:
That is a good idea, I added the debug log entry.
I used two separate `if` statements because Android Studio does not seem to recognize the one line version and warns about API level requirements.

crankycoder:
The `NetworkInfo` class in stumbler is actually used to only check if Wifi is available for use in case the user has asked to only use wifi instead of mobile data plans.

If I understand the intent of this patch, you're trying to filter out mobile wifi data from entering the stumbler.   I don't think you can do that without binding in with DHCP and listening to the DHCP hint that the connection is metered.  Without establishing a DHCP connection, I don't think we have enough information on the client side to determine if a wifi network is metered and a mobile network connection.

cascheberg:
I am not sure if I understand your comment correctly, but this was **not** meant to blacklist mobile WiFi hotspots during the scanning process.

My idea was to treat WiFi as mobile network if there is a chance that it's just consuming data off another phones mobile data plan via hotspot.

To my knowledge, `isActiveNetworkMetered()` will return `true` in the following cases:
- all mobile network connections (non-WiFi)
- WiFi connections that are using a hotspot from an Android host >= 4.1 (announced over DHCP)
- WiFi connections that are manually tagged on the client side as being mobile hotspot

crankycoder:
I don't think we should really need to be concerned about treating wifi as mobile data plan connection.

PR #991, opened 2014-10-12
 Cell/WiFi count is only set when the MLS layer is enabled

cascheberg:
Currently the cell/WiFi count is only set when the MLS layer is enabled. As a result the points overlay always draws the green dot icon, never any cell/WiFi only icon.


cascheberg:
This is not working as expected, sorry, will check again

cascheberg:
OK now, fixed a stupid bug

cascheberg:
Yes, I think you are totally right. I modified the code as suggested.

garvankeeley:
Thanks for the PR!

Issue #974, opened 2014-10-09
show whether stumbler is able to scan wifi and/or cells


Djfe:
there should be an evidence if the stumbler isn't able to scan for wifi/cell whether it be because of:
- wifi is turned off and the always scan setting isn't enabled
- airplane mode is on
- a wifi hotspot is running on the antenna

maybe we could grey out the icons or make them red like the GPS is at the beginning

EDIT:
there should be listeners so that the stumbler realizes after the start/during scanning that the wifi antenna (or something else) is available again

AgentSteel:
or maybe display a message to ask the user to enable wifi

Djfe:
this is about issues where it might not be obvious that MozStumbler doesn't have wifi access.

Another issue is:
if you allow MozStumbler/apps in general to scan while wifi being off while MozStumbler is scanning already MozStumbler won't scan until you restart it (if wifi is off)

if you have that option is off and you are in airplane mode or have a hotspot on it might not be obvious that MozStumbler isn't able to scan

if the font/icon of wifi/cell turn red if MozStumbler isn't able to scan then that would be enough for me (no alert)

PR #996, opened 2014-10-13
Use different color for points with no data

cascheberg:
Currently, points that have WiFi and cell data are drawn in green, just as well as points that have no observed network data. This changes it so that points with no observed network data are drawn in gray.
This needs #991 to work well.

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/mapview/ObservationPointsOverlay.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/mapview/ObservationPointsOverlay.java
@@ -26,6 +26,7 @@ import java.util.LinkedList;
 class ObservationPointsOverlay extends Overlay {
     private static final String LOG_TAG = ObservationPointsOverlay.class.getSimpleName();
     private final Paint mRedPaint = new Paint();
+    private final Paint mGrayPaint = new Paint();
     private final Paint mGreenPaint = new Paint();
     private final Paint mCellPaint = new Paint();
     private final Paint mWifiPaint = new Paint();
@@ -51,6 +52,9 @@ class ObservationPointsOverlay extends Overlay {
 
         mRedPaint.setColor(Color.RED);
         mRedPaint.setStyle(Paint.Style.FILL);
+
+        mGrayPaint.setColor(Color.GRAY);
+        mGrayPaint.setStyle(Paint.Style.FILL);
 
         mBlackStrokePaint.setColor(Color.BLACK);
         mBlackStrokePaint.setStyle(Paint.Style.STROKE);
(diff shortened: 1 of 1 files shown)


Djfe:
An entry in the map legend is missing still ;)

garvankeeley:
The underlying bug would be a report with no data, it shouldn't be possible. What are you seeing happen?

cascheberg:
I can reproduce this by activating airplane mode. The observations are **not** listed as reports in the upload queue, but they show up on the map, and I think that is quite useful.

garvankeeley:
That's a bug, so I'd rather we fix that, where an observation point is being created when there is no scanning data. I thought we already had a guard in the code for this but I guess not.

Djfe:
@garvankeeley take a look at:
EDIT: forgot the issue id #974 

@cascheberg like Garvan I would prefer not drawing any points since we got a normal GPS track
but it could be useful for debugging, as a dev setting, but that's optional and only useful if we get empty reports

cascheberg:
What are your objections?
My idea is this: the user will know that the app is still making correct observations, and that the result of the observation is that there are no networks at that position. Otherwise it's not clear whether there are no networks or if the app lost the GPS signal and made no valid observations at all.

garvankeeley:
We are defining an observation point as a point that has a location +
cell/wifi. We should use some other means to indicate there are collection
problems. There have been a few requests to indicate this in the icons at
the top of the map.

cascheberg:
> We are defining an observation point as a point that has a location + cell/wifi.

This obviously makes sense on the server side (at least).

> There have been a few requests to indicate this in the icons at the top of the map.

Well, this only helps though if I am looking at the map at that exact moment.

But then again, maybe it really doesn't matter all that much if it's lost GPS or no cell/wifi...

cascheberg:
@garvankeeley Would it be a good idea to make `Reporter` send a new intent at the end of `flush` that can be received by `ObservedLocationsReceiver`? It looks like there was some code that did something like that. Currently they both receive and independently process `SUBJECT_NEW_LOCATION` concurrently, is that correct?

garvankeeley:
Yes, I believe that is so that when collection starts, we can show a dot,
rather than when a collection is complete. It provides another clue to the
user that collection is occuring.

cascheberg:
Referring to

> when collection starts, we can show a dot, rather than when a collection is complete. It provides another clue to the user that collection is occuring.

and

> I'd rather we fix that, where an observation point is being created when there is no scanning data. I thought we already had a guard in the code for this but I guess not.

I see the problem that it's not possible to know if that point will have WiFi/cell data before collection is complete.
Showing a dot first and then removing it in case it's not valid could be a possible solution, but it might be confusing.
And currently there is the glitch in `ObservedLocationsReceiver` that the retrieved `currentBundle` might actually belong to the observation before the `lastObservation`.

I'm interested in providing a patch for this, but I don't know what the desired solution should look like. Any other suggestions?

garvankeeley:
My preferred solution is that we build up our test framework so that we can
add bugs like this to the test cases, and then fix them afterwards. These
fringe cases get fixed, and then broken again, because we aren't testing
for them regularly.

PR #998, opened 2014-10-14
Reduce bandwidth at low zoom level: 'trick' the map into showing lower zoom levels.

garvankeeley:
For example, if the map is at zoom level x, show the tiles for x - 4. I don't do this exactly.

This code has a low/high zoom level switch, at
the high zoom level, behave as before this commit; at the low zoom
level, trick the map into showing very low resolution tiles, so only a
few tiles need loading to cover the map area.


garvankeeley:
Video demonstrating this feature: 
https://drive.google.com/file/d/0B1sYPwjczMsQWGtFSUdLTTZKS0U

First is high-res mode, it shows the lower resolution coverage map when zoomed out, and a higher res coverage when zoomed in. Next is low-res mode, showing two levels of low res map tiles, depending on the zoom level.  Notice the lower right label on the map is the status.

android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapActivity.java
```diff
@@ -74,16 +78,19 @@
     private static final int DEFAULT_ZOOM_AFTER_FIX = 16;
     private static final String LAT_KEY = "latitude";
     private static final String LON_KEY = "longitude";
+    private static final int HIGH_LOW_ZOOM_THRESHOLD = 14;
```
crankycoder: Rename to HIGH_ZOOM_THRESHOLD to be consistent with https://github.com/mozilla/MozStumbler/pull/998/files#diff-ecc6477a0a55adf1f37c8d8de23f5f85R284

garvankeeley: Nevermind, I see, change the name to HIGH_ZOOM_THRESHOLD

android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapActivity.java
```diff
@@ -239,38 +264,77 @@ public void run() {
...
+    }
+
+    //
+    // This determines which level of detail of tile layer is shown.
+    //
+    private boolean isHighZoom(int zoomLevel) {
+        return zoomLevel > HIGH_LOW_ZOOM_THRESHOLD;
+    }
+
+    private void updateOverlayBaseLayer(int zoomLevel) {
```
crankycoder, line 289: Can you put a comment in this method similar to `updateOverlayCoverageLayer` so that it's clear what we're trying to do here.

garvankeeley:
Argh. I used git-cola to push, and 9a2c66e has the wrong commit message. Let me fix...

Actually, no it is ok, the main thing is "Rename MapActivity to MapFragment", so it is correct.

Issue #997, opened 2014-10-14
NPE while starting the PreferencesScreen


crankycoder:
Crash on line: 

```
org.mozilla.mozstumbler.client.subactivities.PreferencesScreen.onCreate(PreferencesScreen.java:55)
```

https://github.com/mozilla/MozStumbler/blob/v0.40.1.0/android/src/main/java/org/mozilla/mozstumbler/client/subactivities/PreferencesScreen.java#L55

```
java.lang.RuntimeException: Unable to start activity ComponentInfo{org.mozilla.mozstumbler/org.mozilla.mozstumbler.client.subactivities.PreferencesScreen}: java.lang.NullPointerException
at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2394)
at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2446)
at android.app.ActivityThread.access$600(ActivityThread.java:165)
at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1373)
at android.os.Handler.dispatchMessage(Handler.java:107)
at android.os.Looper.loop(Looper.java:194)
at android.app.ActivityThread.main(ActivityThread.java:5434)
at java.lang.reflect.Method.invokeNative(Native Method)
at java.lang.reflect.Method.invoke(Method.java:525)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:834)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:601)
at dalvik.system.NativeStart.main(Native Method)
Caused by: java.lang.NullPointerException
at org.mozilla.mozstumbler.client.subactivities.PreferencesScreen.onCreate(PreferencesScreen.java:55)
at android.app.Activity.performCreate(Activity.java:5121)
at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1146)
at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2358)
... 11 more
java.lang.NullPointerException
at org.mozilla.mozstumbler.client.subactivities.PreferencesScreen.onCreate(PreferencesScreen.java:55)
at android.app.Activity.performCreate(Activity.java:5121)
at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1146)
at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2358)
at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2446)
[...]

PR #1001, opened 2014-10-14
Fix NPE, make sure ClientPrefs exists at all times

garvankeeley:
https://github.com/mozilla/MozStumbler/issues/997

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/ClientPrefs.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/ClientPrefs.java
@@ -31,8 +31,8 @@ public class ClientPrefs extends Prefs {
     }
 
     public static synchronized ClientPrefs getInstance() {
-        if (sInstance == null || !sInstance.getClass().isInstance(ClientPrefs.class)) {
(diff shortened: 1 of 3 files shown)


android/src/main/java/org/mozilla/mozstumbler/client/ClientPrefs.java
```diff
@@ -21,16 +23,17 @@ protected ClientPrefs(Context context) {
...
-        return sInstance;
+        return (ClientPrefs)sInstance;
     }
 
     public static synchronized ClientPrefs getInstance() {
-        assert(sInstance != null);
-        assert(sInstance.getClass().isInstance(ClientPrefs.class));
+        if (AppGlobals.isDebug && sInstance != null) {
+            assert (sInstance.getClass().isInstance(ClientPrefs.class));
+        }
```
crankycoder, line 36: These asserts will only work on debug builds, can we put in a different check?

garvankeeley, line 36: updated to an exception

PR #1005, opened 2014-10-15
Changed the dependency for SimpleKML to use ivy

crankycoder:
We have an ivy repository for mozilla-services thanks to @st3fan now.

woot!


garvankeeley:
tested, works, stamp it

Issue #975, opened 2014-10-09
the leaderboard doesn't jump to your nickname anymore


Djfe:
the leaderboard doesn't jump to your nickname anymore since it got integrated into the app using the webview component
either you are not forwarding the anchor or the webview component doesn't render it correctly

garvankeeley:
fixed

PR #1009, opened 2014-10-15
975 - fixed jumping to anchor for a nick

garvankeeley:
but if no nick stem we need this ichnaea bug fixed:
https://github.com/mozilla/ichnaea/issues/327
to jump to the start of the leaderboard.

For issue #975


cascheberg:
Do you want to keep `setJavaScriptEnabled(true)` in the `LeaderboardActivity` WebView?
Android Studio notes that

> Using setJavaScriptEnabled can introduce XSS vulnerabilities into your application

Although that might not even be a problem in this app,  I was just wondering.

garvankeeley:
No we should remove that

PR #1017, opened 2014-10-16
fix some hints found by Android Studio code inspection

cascheberg:


cascheberg:
Ok, fixed erroneous change

Issue #1023, opened 2014-10-17
MainApp::onStop crashes (if battery <15%, app crash on startup)


crankycoder:
Reference to the StumblerService in MainApp::onStop doesn't check for null properly.

```
java.lang.RuntimeException: Error receiving broadcast Intent { act=org.mozilla.mozstumbler.turnMeOff flg=0x10 } in org.mozilla.mozstumbler.client.MainApp$ServiceBroadcastReceiver@41818b20
at android.app.LoadedApk$ReceiverDispatcher$Args.run(LoadedApk.java:798)
at android.os.Handler.handleCallback(Handler.java:800)
at android.os.Handler.dispatchMessage(Handler.java:100)
at android.os.Looper.loop(Looper.java:194)
at android.app.ActivityThread.main(ActivityThread.java:5392)
at java.lang.reflect.Method.invokeNative(Native Method)
at java.lang.reflect.Method.invoke(Method.java:525)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:833)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:600)
at dalvik.system.NativeStart.main(Native Method)
Caused by: java.lang.NullPointerException
at org.mozilla.mozstumbler.client.MainApp.stopScanning(MainApp.java:240)
at org.mozilla.mozstumbler.client.MainApp$ServiceBroadcastReceiver.onReceive(MainApp.java:309)
at android.app.LoadedApk$ReceiverDispatcher$Args.run(LoadedApk.java:788)
... 9 more
java.lang.NullPointerException
at org.mozilla.mozstumbler.client.MainApp.stopScanning(MainApp.java:240)
at org.mozilla.mozstumbler.client.MainApp$ServiceBroadcastReceiver.onReceive(MainApp.java:309)
at android.app.LoadedApk$ReceiverDispatcher$Args.run(LoadedApk.java:788)
at android.os.Handler.handleCallback(Handler.java:800)
at android.os.Handler.dispatchMessage(Handler.java:100)
at android.os.Looper.loop(Looper.java:194)
at android.app.ActivityThread.main(ActivityThread.java:5392)
at java.lang.reflect.Method.invokeNative(Native Method)
at java.lang.reflect.Method.invoke(Method.java:525)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:833)
[...]

PR #1026, opened 2014-10-17
1023 - Failing to check for null service onStart and onStop

garvankeeley:
When the app starts with <15% battery, it immediately gets notified and
then calls onStop before the service is created, and crashed on a null
pointer.

Issue #1023


garvankeeley:
To test the low battery intent, I normally send it manually through ADB. To try repro this case, I ran my battery down to <15%, but I couldn't get the crash to occur. Regardless, this code will defend against a crash, I am just not clear on how it was caused.

PR #1031, opened 2014-10-17
Update Reporter to send flushed bundles to ObservedLocationsReceiver

cascheberg:
This fixes the case that green dots are drawn on the map when there are no cell/wifi observations, see #996.
I think it also cleans up the data flow and fixes another glitch.
This way `ObservedLocationsReceiver` will get the location that is currently being observed from the GPS, but only adds observations to the map when they are validated and flushed by the `Reporter`.

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/ObservedLocationsReceiver.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/ObservedLocationsReceiver.java
@@ -20,8 +20,6 @@ import org.mozilla.mozstumbler.client.mapview.ObservationPoint;
 import org.mozilla.mozstumbler.service.core.logging.Log;
 import org.mozilla.mozstumbler.service.stumblerthread.Reporter;
 import org.mozilla.mozstumbler.service.stumblerthread.datahandling.StumblerBundle;
-import org.mozilla.mozstumbler.service.stumblerthread.scanners.GPSScanner;
-import org.mozilla.mozstumbler.service.utils.CoordinateUtils;
 import org.osmdroid.util.GeoPoint;
 
 import java.lang.ref.WeakReference;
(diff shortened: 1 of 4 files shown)


garvankeeley:
Hi, can you provide steps to reproduce the bug you are fixing? Is this in airplane mode?

cascheberg:
Yes, the easiest way to reproduce is in airplane mode.
It also fixes the glitches that the session observations do not match (e.g. are one ahead of) the actual queue, and that the associated MLS query point might in fact belong to the previous observation.

garvankeeley:
I have tested a number of times in airplane mode, and not seen any
collection taking place, are you sure there isn't more steps to reproduce
this?

cascheberg:
It's not in the queue, it just shows up on the map.
Reading the code, the `ObservedLocationsReceiver` currently just adds every new location that it receives as an intend from the `GPSScanner` to its `mCollectionPoints`, and those are displayed on the map, without checking for observed networks.
Right now I have no idea why this can not be reproduced.

Screenshot taken after reinstalling the app, then started while in airplane mode, GPS enabled, on Nexus 4.
screenshot_2014-10-18-01-40-08

garvankeeley:
When I was testing, I would switch airplane mode on while stumbling. My
current location would update but no observation points. Do you have to
start the app in airplane mode maybe? I'll try that.

cascheberg:
The MLS layer needs to be disabled to reproduce this.

garvankeeley:
Cool, got the bug to happen, I like your approach of hooking up to the Reporter's bundles, I'll look at the PR now.

cascheberg:
Ok, I think I fixed all review comments (so far).

garvankeeley:
Looks good, thanks!

Issue #1002, opened 2014-10-14
Map behaves poorly when there are lots of scan points marked on the map


kbrosnan:

kbrosnan:
When there are a lot of observations on the map it will start to flicker on panning and eventually will stop displaying new points.

http://people.mozilla.org/~kbrosnan/tmp/mozstumbler-1002/mozStumbler-flicker.mp4

garvankeeley:
Nice stumbling! We set our bar low, which was, keep running when the observation count gets super-high. We are getting close to finding cycles to do a proper fix, not sure if we can get it in for 1.0.
You should see that the new observations draw first, and older ones will flicker, that is expected (albeit ugly) behaviour for now.

Dev Note:
- For 1.0, instead of stopping storage at the point limit, clear out old points and show new ones. 
- the future plan is to draw observation points to the saved rasters directly, instead of rendering as a vector layer
- if ^^ is problematic, store batches of points in R-tree, as we are not currently considering the viewport. This is tricky to combine with MRU concept that we have now.

garvankeeley:
Hopefully fixed in v0.41.0.0

PR #1040, opened 2014-10-20
1002: Point draw performance. Use hash of x and y to reduce number of points drawn

garvankeeley:
For #1002 

This uses a hash to simulate a 2D grid, by hashing on the x and y values combined.
It reduces the number of points drawn at a given zoom level. 
Recalculate each time the zoom level changes. The hash value is based on the x and y pixel, but scaled down to reduce the detail level (and increase collisions and reduce the data). If a particular hash slot is occupied already, then don't draw the point. 
_Note_, I added a bit of logic here, so that if the observation point is "better" [i.e. cell+wifi is better than cell], then it takes the hash slot.

Tested on Nexus 5 4.4.4 and Samsung Duos 2.3.6.

Testing is performed with: 
4800 point kml file
https://drive.google.com/file/d/0B1sYPwjczMsQM1VDdXdacGpDdEE/view?usp=sharing

Put it in: sdcard/Android/data/org.mozilla.mozstumbler/files

The go to load KML in the app.

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/ObservedLocationsReceiver.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/ObservedLocationsReceiver.java
@@ -9,6 +9,7 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.location.Location;
+import android.os.Build;
 import android.os.Handler;
 import android.os.Looper;
 import android.support.v4.content.LocalBroadcastManager;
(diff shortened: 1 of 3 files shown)


garvankeeley:
Here is a video on 2.3.6 device with 4800 points (note the code limits devices less than Android 4.x to 2500 points). It shows panning and zooming working well. The points are rendering newest to oldest, and the point drawing stops due to timing out, which is correct. The test data is not such that each line is newer/older than the next, but each point in each line from east to west is newer (east is newer than west).
https://drive.google.com/file/d/0B1sYPwjczMsQclZZQVFRR1lwRmM/view?usp=sharing

garvankeeley:
Based on chat with @cranckycoder, he is rubber stamping as we need lots of hands-on feedback, and based on that we will review the approach further.

PR #1041, opened 2014-10-20
Fix observations display

cascheberg:
This fixes a bug where the number of observed locations displayed in the map was reset to 0 after device rotation and only updated when a new location was observed.


garvankeeley:
Thanks looks good.

Issue #1039, opened 2014-10-20
Reduce collection frequency on stumbler


crankycoder:
We need to reduce the collection frequency on the stumbler down to ~30m.  We don't want ichnaea to keel over on initial google playstore release.

cc/ @hannosch

cascheberg:
A note here: currently WiFi is set to only scan every 5 seconds. That is ~42m at 30km/h.

hannosch:
I think there are two thresholds. We only collect data if you have at least moved X meters (currently 10m, proposed to be increased to 30m). And additionally only collect data at most every X seconds (which I thought was currently 1s).

garvankeeley:
GPSScanner.java currently has 10m and 1sec

cascheberg:
I was referring to `WifiScanner` that has a scheduler to run `getWifiManager().startScan()` every 5 seconds only. When that is finished, it sends a `ACTION_WIFIS_SCANNED` intent to the `Reporter`. That means that if you travel fast, currently only every ~4th observation contains WiFi data at all.
The cell scanner is scanning every 1 second.

garvankeeley:
The android wifi layer actually seems to scan closer to every minute. That
5 second scan does not trigger the hardware to scan any faster.

cascheberg:
I really do not know anything about Android (hardware) internals.
But: when I change the MozStumbler WiFi scanner to scan every second, it happens that the number of WiFis displayed in MozStumbler even changes within that time, and the native Android WiFi listing then also updates at that frequency (on my Nexus 4).

Djfe:
even though this got closed:
in 0.40.4.0 I recognized that if I ran/drove too fast then I often got points without wifi data (cell only report: blue square)

crankycoder:
Marking this as a duplicate of #1042

PR #1042, opened 2014-10-20
1039 reduce collection rate, changed 10m->30m 1s->2s

garvankeeley:
#1039


Djfe:
isn't it better to choose only 20ms?
we will keep a little better accuracy and would double both constants

crankycoder:
@djfe This is to reduce collection from hitting the server side collection too hard when the Google Play Store and lots of users come on for only a day or two.

Djfe:
one issue here:
users using the older versions get more points on the (not weekly) leaderboard...

PR #1046, opened 2014-10-20
Fix hardcoded strings and remove some unused resources

### LINKED ISSUE #1018: Strings in layout file
Why is there a string missplaced in a layout file?
It is supposed to be in the strings.xml that is translatable.

https://github.com/mozilla/MozStumbler/blob/dev/android/src/main/res/layout/fragment_kml.xml

cascheberg:
This fixes #1018


cascheberg:
I think the `onClick_ViewMapboxAttribution` could be removed because of the `autoLink`, and then some more code could be removed, but I am not quite sure...

garvankeeley:
Looks good thanks (I agree that onClick_ViewMapboxAttribution is not used)

Djfe:
@cascheberg thx for fixing this ;)
I was waiting for something like this to update the German translation of MozStumbler

gonna do that now :)

Issue #1045, opened 2014-10-20
Error deserializing status from JSON: org.json.JSONException: No value for status


cpeterso:
I saw JSON errors like these logged every ~10 seconds. The MozStumbler app was running in the foreground, but I was not actively stumbling.

```
10-20 11:54:09.927: E/Stumbler:MLSLocationGetter(23764): Error deserializing status from JSON
10-20 11:54:09.927: E/Stumbler:MLSLocationGetter(23764): org.json.JSONException: No value for status
10-20 11:54:09.927: E/Stumbler:MLSLocationGetter(23764): at org.json.JSONObject.get(JSONObject.java:354)
10-20 11:54:09.927: E/Stumbler:MLSLocationGetter(23764): at org.json.JSONObject.getString(JSONObject.java:514)
10-20 11:54:09.927: E/Stumbler:MLSLocationGetter(23764): at org.mozilla.mozstumbler.client.mapview.MLSLocationGetter.doInBackground$44e62fb4(MLSLocationGetter.java:71)
10-20 11:54:09.927: E/Stumbler:MLSLocationGetter(23764): at org.mozilla.mozstumbler.client.mapview.MLSLocationGetter.doInBackground(MLSLocationGetter.java:26)
10-20 11:54:09.927: E/Stumbler:MLSLocationGetter(23764): at android.os.AsyncTask$2.call(AsyncTask.java:287)
10-20 11:54:09.927: E/Stumbler:MLSLocationGetter(23764): at java.util.concurrent.FutureTask.run(FutureTask.java:234)
10-20 11:54:09.927: E/Stumbler:MLSLocationGetter(23764): at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:230)
10-20 11:54:09.927: E/Stumbler:MLSLocationGetter(23764): at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1080)
10-20 11:54:09.927: E/Stumbler:MLSLocationGetter(23764): at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:573)
10-20 11:54:09.927: E/Stumbler:MLSLocationGetter(23764): at java.lang.Thread.run(Thread.java:841)
[...]

PR #1050, opened 2014-10-20
1045 - not handling http response code when searching for mls

garvankeeley:
#1045


android/src/main/java/org/mozilla/mozstumbler/client/mapview/MLSLocationGetter.java
```diff
@@ -58,9 +58,13 @@ public Location doInBackground(String... params) {
             Log.i(LOG_TAG, "Error processing search request");
             return null;
         }
-        // int bytesSent = resp.bytesSent();
 
-        JSONObject response = null;
+        if (!resp.isSuccessCode2XX()) {
+            //TODO detect malformed request, and clear out mlsrequest on observation point
```
crankycoder, line 63: I don't think we have to clear out the MLSRequest - is this comment correct?   A failed request is going to pull the request out of the queue anyway.

If we do need to clear it out, let's do that now.

Issue #973, opened 2014-10-09
no fixed/visible satellites counter anymore and no evidence of an established GPS connection


Djfe:
since 0.40.0 it is not possible anymore to see the number of fixed as well as visible satellites anymore

and it is impossible to see whether the device lost the GPS connection without moving and taking a look at the map/log at the same time

Djfe:
to point this out further:
just having a number of 3 satellites there isn't enough, the GPS needs to be established and can be lost again, but if I loose my GPS connection then the old number (for example 8 satellites) still stands next to the gps icon (the label's text is still "8")

garvankeeley:
We could fit a fractional view there of fixes / visible

PR #1051, opened 2014-10-20
Display visible GPS satellites

cascheberg:
This is a suggestion for a partly fix of #973. I put the "/" to the strings.xml for RTL compatibility.
gps

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/GPSScanner.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/GPSScanner.java
@@ -29,9 +29,9 @@ public class GPSScanner implements LocationListener {
     public static final String NEW_STATUS_ARG_FIXES = "fixes";
     public static final String NEW_STATUS_ARG_SATS = "sats";
     public static final String NEW_LOCATION_ARG_LOCATION = "location";
+    public static final int MIN_SAT_USED_IN_FIX = 3;
 
     private static final String LOG_TAG = AppGlobals.LOG_PREFIX + GPSScanner.class.getSimpleName();
-    private static final int MIN_SAT_USED_IN_FIX = 3;
     private static final long ACTIVE_MODE_GPS_MIN_UPDATE_TIME_MS = 2000;
     private static final float ACTIVE_MODE_GPS_MIN_UPDATE_DISTANCE_M = 30;
     private static final long PASSIVE_GPS_MIN_UPDATE_FREQ_MS = 3000;
--- a/android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/GPSScanner.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/GPSScanner.java
@@ -29,7 +29,7 @@ public class GPSScanner implements LocationListener {
     public static final String NEW_STATUS_ARG_FIXES = "fixes";
     public static final String NEW_STATUS_ARG_SATS = "sats";
     public static final String NEW_LOCATION_ARG_LOCATION = "location";
-    public static final int MIN_SAT_USED_IN_FIX = 4;
+    public static final int MIN_SAT_USED_IN_FIX = 3;
 
     private static final String LOG_TAG = AppGlobals.LOG_PREFIX + GPSScanner.class.getSimpleName();
     private static final long ACTIVE_MODE_GPS_MIN_UPDATE_TIME_MS = 1000;
--- a/android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/GPSScanner.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/GPSScanner.java
@@ -29,9 +29,9 @@ public class GPSScanner implements LocationListener {
     public static final String NEW_STATUS_ARG_FIXES = "fixes";
     public static final String NEW_STATUS_ARG_SATS = "sats";
     public static final String NEW_LOCATION_ARG_LOCATION = "location";
+    public static final int MIN_SAT_USED_IN_FIX = 4;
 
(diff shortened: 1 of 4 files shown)


Djfe:
design looks good :+1:

garvankeeley:
Thanks, I like this approach. 
However, the spacing doesn't look right:
https://www.dropbox.com/s/wgh82guniy3pba8/Screenshot%202014-10-20%2018.16.39.png?dl=0

cascheberg:
Indeed. Does the updated version look good on your device?

garvankeeley:
The gap between the number and the GPS icon is too large:
https://www.dropbox.com/s/68zsssr4ew9x1g6/Screenshot%202014-10-20%2018.38.42.png?dl=0

cascheberg:
Let's try "wrap_content" then. Contra: now all other icons move a little when number of fixed satellites changes from 1 to 2 digits. I think it's not too noticeable though.
I don't know any other way right now, still learning by doing.

cascheberg:
Hm maybe put the GPS part in another nested block?

cascheberg:
Done, looks very promising to me

garvankeeley:
A few points:
1) The moving would need to be fixed.
2) Our base device for UI design is 2.3.6 320 × 480, and the text is cutoff (this is only using 9999 as a test value, the original layout works with a 19999 test value): https://www.dropbox.com/s/ce3mpbzde4w5wp9/2.3.6-gps-2-values.png?dl=0
3) The numbers for the cell, wifi and report count are all set up to show 5 digits (actually 19999 as a test value), MapFragment initTextView() has some code to help with that (where it does the getTextBounds of "00000").
That is based on having to show 5 digits for the wifi or report counts, and the cell uses the same for an even spacing. Arguably we could remove space from the cell counts to make room for the GPS sat.
4) The numbers are all 3-4px spaced from their icon

Djfe:
>9999 cells is unlikely (-> count of reports would rather be above 99999 then)
10000-99999 wifi is more likely

the number of cells could be more close to the cell icon, too to save some space ;)

garvankeeley:
The spacing is designed to be equal between the 3 counts

garvankeeley:
We could try reducing the width of the cell counts, but I would want @ryanfeeley, our UX reviewer to chime in.

cascheberg:
Thank you for the feedback!

The values are only the number of networks that were visible during the last scan. That means that the value is even lower than or equal to the number of networks reported for the current single observation.
I would think that it is not highly likely to have more than 999 visible cell or wifi networks at one single point (of course correct me if I am wrong). The `Reporter` only keeps a maximum of 50 cells (`MAX_CELLS_PER_LOCATION`) and 200 wifis (`MAX_WIFIS_PER_LOCATION`) per observation.

I fixed the moving by putting all GPS related text views into a fixed width linear layout.
I now also reduced the width of the cell and wifi fields to 32dp.
gps
Screenshot from Android Virtual Device

cascheberg:
I found out that `measureText()` returns higher and apparently more reliable values than `getTextBounds()`

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/GPSScanner.java
```diff
@@ -29,9 +29,9 @@
     public static final String NEW_STATUS_ARG_FIXES = "fixes";
     public static final String NEW_STATUS_ARG_SATS = "sats";
     public static final String NEW_LOCATION_ARG_LOCATION = "location";
+    public static final int MIN_SAT_USED_IN_FIX = 4;
```
garvankeeley: this should be a separate PR if we are changing this. I'd like to know more about why.

garvankeeley:
This is looking really good, tried it on a few devices. 

Also, good point about the number of digits in the cell and wifi, they are no where as large as observations, so no need to space them similarly

garvankeeley:
The response from UX is #1060, which I agree with. But, we are trying to finish 1.0 today, so I am inclined to accept this PR (1051) for 1.0 and leave further UI work for 1.1.

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/GPSScanner.java
cascheberg: GPS needs at least 4 satellites to compute x, y, z and time, you can find this on [Wikipedia](http://en.wikipedia.org/wiki/Global_Positioning_System#Navigation_equations) and other sources.
I can revert this rather unrelated change here.
P.S.: apparently, it is not sufficient to check this value on Android. The first few answers to this [stackoverflow question](http://stackoverflow.com/questions/2021176/how-can-i-check-the-current-status-of-the-gps-receiver) are really interesting. According to the second one (and some other sources), there is a `android.location.GPS_FIX_CHANGE` intent (which I could not find in the API).

garvankeeley: Sounds good, please file as a separate PR, in case we have to revert either one of the PRs.

Djfe: Android keeps on trying to calculate a GPS position even with 3 Satellites and less.
This variable is not about the first fix, the first fix occurs with more GPS satellites.
Once Android has a fix it is able to keep the connection to the satellites.

I don't think this has ever caused issues before, has it?

garvankeeley: For me no. I usually see 4 or higher. A position is available with 3
satellites, as the A-GPS can eliminate non-possible positions on the earth,
I haven't tested though to see what accuracy level this shows. I would
assume that if the sats drops below 3, the accuracy becomes poor enough
that we shouldn't be using it.

garvankeeley:
Can you rebase/merge and re-push

PR #1056, opened 2014-10-21
Delete stumbles on 400 (bad request) errors only

garvankeeley:
In theory other errors, particularly 404 can be temporary (and incorrectly) be returned by proxies.
Due to complaints about data disappearing I am trying to minimize the possibilities of stumbles being disposed.


Djfe:
But the API doesn't return 400, it just returns 204 even if it is a bad report
There is no reason to delete reports because of a bad connection.
A bad connection doesn't indicate a wrong report.

PR #1059, opened 2014-10-21
updated de translation, fixed en strings

Djfe:
Updated German translation and fixed some small bits in the English strings.

### DIFF
--- a/android/src/main/res/values-de/strings.xml
+++ b/android/src/main/res/values-de/strings.xml
@@ -1,23 +1,20 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
+
     <string name="app_name">MozStumbler</string>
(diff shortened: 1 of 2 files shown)


garvankeeley:
Thanks!

cascheberg:
This update breaks the metrics layout in German, because metrics assume that the string for wifi networks is the longest.

Djfe:
I cannot compile it myself, can you make a screenshot pls?
the reason why the cells text is longer now is that the wording got changed
from ap/tower -> to network (and ap in German is not a known abbreviation so it had to be written out: "Zugangspunkt" which means access point)
I wrote "Mobilfunk-Netzwerk" instead of "Funk-Netzwerk" because I thought there would be enough space for it

PR #1065, opened 2014-10-22
Round KB to nearest whole number instead of showing fractional metrics like "123.4 KB"

cpeterso:
I think the fractional KB metrics clutter the UI. We don't need that much precision. :)


garvankeeley:
This would need merge/rebase.

But I am partial to the existing display. If in a country that is all cell scans (so each scan is only 1 cell tower), one could easily have 100-300 byte reports, and it wouldn't look like anything much is happening.

cpeterso:
OK. Maybe I'll revisit this PR with a friendly version of file sizes (e.g. "X bytes", "Y KB", or "Z MB") like OCPsoft's PrettyTime does for time.

PR #1070, opened 2014-10-22
Change keep_screen_on pref default to false

cascheberg:
This pref was introduced when the map was not the default screen.
Now this pref is basically always in effect when the app is in the foreground. This is not a good default because users will be unaware of it (at first), so if they do not turn off the screen manually, they might only notice it when the battery is almost empty. On the other hand, users that want to have this feature will look for it in the menu.


garvankeeley:
I know we had a discussion about defaulting this to true, but I am inclined to agree it is best to be safe and default  to off.

Issue #1067, opened 2014-10-22
Cell count on map view shows 1, when many have been scanned


garvankeeley:
My cell count always shows 1, but I can see in the log, the current scan actually had many. I am not sure the cell count has ever shown anything other than 1 on the map view.

garvankeeley:
as @cascheberg noted, now the counts just increase

PR #1071, opened 2014-10-23
1067:Switching to show all cells scanned, 

garvankeeley:
the count we are using isn't showing all the cells that are seen at the current time
#1067

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/StumblerService.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/StumblerService.java
@@ -98,10 +98,6 @@ public class StumblerService extends PersistentIntentService
         return mScanManager.getCellInfoCount();
     }
 
-//    public synchronized int getCurrentCellInfoCount() {
-//        return mScanManager.getCurrentCellInfoCount();
-//    }
-
     // Previously this was done in onCreate(). Moved out of that so that in the passive standalone service
     // use (i.e. Fennec), init() can be called from this class's dedicated thread.
     // Safe to call more than once, ensure added code complies with that intent.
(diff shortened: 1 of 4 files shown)


android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/StumblerService.java
```diff
@@ -98,9 +98,9 @@ public synchronized int getCellInfoCount() {
         return mScanManager.getCellInfoCount();
     }
 
-    public synchronized int getCurrentCellInfoCount() {
-        return mScanManager.getCurrentCellInfoCount();
-    }
+//    public synchronized int getCurrentCellInfoCount() {
```
crankycoder: Just delete this code instead of commenting it out.

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/ScanManager.java
```diff
@@ -158,9 +158,9 @@ public int getCellInfoCount() {
         return (mCellScanner == null)? 0 :mCellScanner.getCellInfoCount();
     }
 
-    public int getCurrentCellInfoCount() {
-        return (mCellScanner == null)? 0 :mCellScanner.getCurrentCellInfoCount();
-    }
+//    public int getCurrentCellInfoCount() {
```
crankycoder: Just delete the code instead of commenting it out.

crankycoder:
r+ after removing the dead code.

cascheberg:
Now the values for cells and wifis mean different things. The number of cells is increasing steadily, while the wifi count is not, it still only shows the number of currently visible wifis, which is not nice. (Changing this for wifis might break the layout.)
Is there anything wrong with `getCurrentCellInfoCount` that can be fixed? I see currently visible cell counts greater than 1 (not always though).

garvankeeley:
The getCurrentCellInfoCount was always reporting 1 on my device, and then my bundle would 5 or 6 cells. The other number was giving me correct results, I didn't realize it was steadily increasing (need to move more). I don't know why the counts were set up to behave so differently. Being that the current number appears correct during a bundle, can't we just reset it to zero when the Reporter flushes a bundle to disk?

Issue #1069, opened 2014-10-22
Disable "keep screen on while in map" when in low power situations


kbrosnan:
When keep screen on while in map is selected and the device is under 15% battery the display never turns off and since it is the foreground application the current location is still requested. Allowing the screen to sleep when in low power mode seems to disable the location requests.

garvankeeley:
Good catch. Under 15% we should shut off the screen.

garvankeeley:
Changed this pref to 'Keep screen on while scanning', no reason to keep it on otherwise.

PR #1074, opened 2014-10-23
1069 - Keeping screen on pref only takes effect when scanning.

garvankeeley:
#1069


android/src/main/java/org/mozilla/mozstumbler/client/navdrawer/MainDrawerActivity.java
```diff
@@ -155,6 +153,12 @@ private void updateStartStopMenuItemState() {
         }
         boolean isScanning = svc.isScanning();
 
+        if (isScanning) {
+            keepScreenOnChanged(ClientPrefs.getInstance().getKeepScreenOn());
```
crankycoder: Nitpicking, but can we rename this method to `keepScreenOn` instead of `keepScreenOnChanged` as the argument we're passing is is the actual state of the screen on flag.

crankycoder:
r+ after method rename.

PR #1085, opened 2014-10-23
1067: Better fix: Listen for Reporter to flush bundle, then clear the cell count on next scan.

garvankeeley:
#1067
#1071


cascheberg:
Seems like a good idea. Since I probably cannot reproduce this, I wonder, what does the result look like then? Does the counter increase every second and then drop to one when cleared or does it instantly show the correct number?
Side note: better use the old function name?

garvankeeley:
The counter doesn't clear until the next scan has started, to prevent the number from going to zero briefly

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/cellscanner/CellScanner.java
```diff
@@ -77,6 +86,18 @@ public void start(final ActiveOrPassiveStumbling stumblingMode) {
             return;
         }
 
+        LocalBroadcastManager.getInstance(mContext).registerReceiver(mReportFlushedReceiver,
+                new IntentFilter(Reporter.ACTION_NEW_BUNDLE));
+
+        // This is to ensure the broadcast happens from the same thread the CellScanner start() is on
```
crankycoder, line 92: Let's reword this to make it clear that we need this to run on the UI thread. "Make sure that we send the broadcast on the UI thread."

garvankeeley, line 92: It is the UI thread, but only in Mozilla Stumbler. On fennec this is the service thread. 
Side note: After 1.0 goes out, I want to switch the code over to run _exactly_ like fennec does.

garvankeeley, line 92: p.s. It would be good to review our use of TimerTasks in the code, not just for threading issues, but to set them up so they do as little work as possible in the TimerTask, and message back to the calling thread.

PR #1091, opened 2014-10-23
Use table layout in metrics drawer, improves l12y

cascheberg:
This improves localizability, because the table columns automatically align at the widest title string. This fixes the problem mentioned in #1059


garvankeeley:
Great idea! Will take a look now...

garvankeeley:
much better

Issue #1092, opened 2014-10-23
OOM errors after applying #1075


crankycoder:
OOM problem with bitmaps

```
  90 E/Stumbler:BitmapTileSourceBase( 3632): OutOfMemoryError loading bitmap: /storage/emulated/0/Android/data/org.mozilla.mozstumbler/cache/osmdroid/tiles/MozStumbler Tile Store/18/73229/95670.png32     .tile
  91 E/Stumbler:BitmapTileSourceBase( 3632): java.lang.OutOfMemoryError
  92 E/Stumbler:BitmapTileSourceBase( 3632): >---at android.graphics.BitmapFactory.nativeDecodeStream(Native Method)
  93 E/Stumbler:BitmapTileSourceBase( 3632): >---at android.graphics.BitmapFactory.decodeStreamInternal(BitmapFactory.java:620)
  94 E/Stumbler:BitmapTileSourceBase( 3632): >---at android.graphics.BitmapFactory.decodeStream(BitmapFactory.java:596)
  95 E/Stumbler:BitmapTileSourceBase( 3632): >---at android.graphics.BitmapFactory.decodeFile(BitmapFactory.java:376)
  96 E/Stumbler:BitmapTileSourceBase( 3632): >---at org.mozilla.osmdroid.tileprovider.tilesource.BitmapTileSourceBase.getDrawable(BitmapTileSourceBase.java:110)
  97 E/Stumbler:BitmapTileSourceBase( 3632): >---at org.mozilla.osmdroid.tileprovider.modules.SmartFSProvider$TileLoader.loadTile(SmartFSProvider.java:192)
  98 E/Stumbler:BitmapTileSourceBase( 3632): >---at org.mozilla.osmdroid.tileprovider.modules.MapTileModuleProviderBase$TileLoader.run(MapTileModuleProviderBase.java:289)
  99 E/Stumbler:BitmapTileSourceBase( 3632): >---at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
 100 E/Stumbler:BitmapTileSourceBase( 3632): >---at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
 101 E/Stumbler:BitmapTileSourceBase( 3632): >---at java.lang.Thread.run(Thread.java:841)

```

PR #1093, opened 2014-10-23
1092 reduce heap usage to prevent OOM

crankycoder:
This is a fix for #1092 

This lowers the LRUCache size for the drawable pool down to it's original size as well as subsampling the images before they are loaded into heap.

On my MotoG before the changes I could easily get heap usage to exceed 85MB causing OOM errors.

Lowering the LRU size reduces the heap consumption to around 60MB on my device.  Subsampling cuts the heap down to ~45MB usage.

Subsampling makes road labels a bit hard to read, so we should increase the font size on the road mapstyling on our mapbox account.


garvankeeley:
I don't disagree, but we need the MapBox map style updated for this to be usable.
And for MapQuest tiles, they are unreadable on my device.

crankycoder:
Mozilla has custom map tiles generated and restyling the font size for roads doesn't see possible.

We're going to punt on this until post 1.0.

crankycoder:
I've pushed a change to revert the subsampling, so this patch will just reduce the LRUCache size.

garvankeeley:
What was effect/benefit of this value (100 vs 0), in terms of visible effect?

crankycoder:
No visible change when going from 100 to 0.

Issue #1089, opened 2014-10-23
Max zoom out level should be dependant on screen size


Djfe:
why did the lowest zoom level get reduced so much? I cannot scroll through countries efficiently anymore now (takes too long)

garvankeeley:
It loads too much tile data, consuming network bandwidth. The app isn't intended for general viewing of worldwide coverage (the website is better for that, maybe we should link to it in some way), only for the local area where the user will be stumbling.

Djfe:
ok, will close this now

Djfe:
Actually I think you should still leave this to the user
The user ist still able to disable the map

And it consumes less bandwidth the further you drive
I'm riding a train at the moment and the zoom level is far to small to see the complete route and going through all recorded points takes far more bandwidth than on another zoomlevel where i can see more... (only a few big tiles need less bandwidth than a lot of small ones, it would only take more bandwidth if I would switch the zooming level all the time...)

garvankeeley:
I see the bug. We should configure this based on the screen size.  On the Nexus 5 zoom out shows 25x25 km, on a small screen device (the Samsung Duos 320 x 480 I have) it shows 5x5 km.

VolMi:
On my S4 mini (540 x 960) and using version 0.99.0.0, my max Field of View is at most about 5x5 km. This is really unpleasant. I used to peek nation- continent- and even worldwide quite regularly and liked it.

garvankeeley:
Did you notice _any_ difference with the new build? Maybe we aren't checking screen size in a portable way.

Djfe:
People complaining about too much data consumption should know that scrolling through the map needs to load new tiles...

That's like people in the ios app store that complain in their reviews about swiftkey that it needs full access to what you type in.
I was like WTF when I read that :D , I mean it was like this forever on Android and I never heard anyone complaining because it just makes sense that it needs to know what you type in to give you proposals what you might want to type in. And now that the app came out on IOS this is what new users are complaining about lol.
(sorry for a topic but it's a fitting example)

As a dev you should at least expect users that have common sense... (especially for such a specific app like this one)
you already have tiles that are only 25% as big as before. Pls let us the choice what we want and what not.
I don't have always the time to check the online map to see what the world looks like...
And I have wifi to it doesn't consume data to load an endless number of tiles for me.

[...]

PR #1116, opened 2014-10-24
1089: For screens less than 500px wide, allow higher zooming on the map. 

garvankeeley:
On the 320px screen I tested, full zoom out was 20x20km, previously it was 5x5km. 
Screens over 500px wide are unchanged.
#1089

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/mapview/tiles/AbstractMapOverlay.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/mapview/tiles/AbstractMapOverlay.java
@@ -9,6 +9,10 @@ import android.graphics.Canvas;
 import android.graphics.Point;
 import android.graphics.Rect;
 import android.graphics.drawable.Drawable;
+import android.os.Build;
+import android.util.DisplayMetrics;
+import android.view.Display;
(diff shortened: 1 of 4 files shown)


android/src/main/java/org/mozilla/mozstumbler/client/mapview/tiles/AbstractMapOverlay.java
```diff
@@ -35,8 +39,25 @@
...
+        super(new BetterTileProvider(context), new DefaultResourceProxyImpl(context));
+    }
+
+    public static void setDisplayBasedMinimumZoomLevel(Context c) {
+        WindowManager wm = (WindowManager) c.getSystemService(Context.WINDOW_SERVICE);
+        Display display = wm.getDefaultDisplay();
+        DisplayMetrics outMetrics = new DisplayMetrics();
+        display.getMetrics(outMetrics);
+
+        sMinZoomLevelOfMapDisplaySizeBased = (outMetrics.widthPixels < 500)? 11 : 13;
```
crankycoder: 11 and 13 should be pulled out as constants. Maybe as SMALL_SCREEN_MIN_ZOOM and LARGE_SCREEN_MIN_ZOOM?

Issue #1121, opened 2014-10-27
Acra report: NPE ObservationPointsOverlay.java:85


garvankeeley:
java.lang.NullPointerException
at org.mozilla.osmdroid.views.Projection.toPixels(Projection.java:102)
at org.mozilla.mozstumbler.client.mapview.ObservationPointsOverlay.update(ObservationPointsOverlay.java:85)
at org.mozilla.mozstumbler.client.mapview.MapFragment.newMLSPoint(MapFragment.java:683)
at org.mozilla.mozstumbler.client.ObservedLocationsReceiver$1.run(ObservedLocationsReceiver.java:89)

PR #1122, opened 2014-10-27
1121 - NPE due to calling newMLSPoint with a null point

garvankeeley:
#1121

This happens due to 400 errors in requesting MLS point. Which leads to the next question as to _why_ we are getting 400 errors here. These should be logged already, so we should investigate.


android/src/main/java/org/mozilla/mozstumbler/client/mapview/ObservationPointsOverlay.java
```diff
@@ -82,6 +83,10 @@
     void update(ObservationPoint obsPoint, MapView mapView, boolean isMlsPointUpdate) {
         final Projection pj = mapView.getProjection();
         GeoPoint geoPoint = (isMlsPointUpdate)? obsPoint.pointMLS : obsPoint.pointGPS;
+        if (geoPoint == null) {
+            Log.i(LOG_TAG, "Caller error: geoPoint is null");
```
crankycoder: Should this be a warning and not an info level message?  It seems like it shouldn't happen.

PR #1123, opened 2014-10-27
Bump android tools gradle plugin to 0.12.+, and use the robolectric-gradle-plugin 0.12.+

keverets:
There was an issue with packaging MozStumbler in fdroid due to build.gradle using 0.12.2
and android/build.gradle using 0.10.

After applying this change (and doing a "git clean -dfx"), the unittest target and the build targets work correctly.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -11,11 +11,9 @@ buildscript {
     }
 
     dependencies {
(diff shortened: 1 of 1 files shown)


crankycoder:
Thanks @keverets!

Issue #1129, opened 2014-10-27
Map not updating with scanned locations


kbrosnan:
Using v0.42.0.0 on a Galaxy S5

I have seen times when the map displays scan locations spottily. Locking the phone and unlocking the phone displays the scan locations correctly.
![screenshot_2014-10-25-16-19-39](https://cloud.githubusercontent.com/assets/250273/4799280/3c7931e6-5e1b-11e4-9790-22fe4f1a31e7.png)
![screenshot_2014-10-25-16-19-46](https://cloud.githubusercontent.com/assets/250273/4799281/3c796c06-5e1b-11e4-8ef5-4b3116b45f93.png)
![screenshot_2014-10-27-09-22-47](https://cloud.githubusercontent.com/assets/250273/4799282/3c7aeae0-5e1b-11e4-984e-5fb9a68cdb17.png)
![screenshot_2014-10-27-09-22-56](https://cloud.githubusercontent.com/assets/250273/4799279/3c78fe9c-5e1b-11e4-845b-f82a17b2a700.png)

garvankeeley:
I assume zooming in and out didn't trigger a redraw?

kbrosnan:
I did not try that.

garvankeeley:
Zooming triggers a full point redraw, or panning so the points aren't shown and then panning back. Actually those are two different drawing paths, so it would be cool to know if that behaves differently, but I am guessing this is tricky to repro.

crankycoder:
I'm still getting this on the dev branch after #1135 was merged.  I'm seeing exactly the same behavior as @RexWheeler describes in https://github.com/mozilla/MozStumbler/issues/1087#issuecomment-60713970 

When I'm centered, the map moves with my position remaining centered, but no new dots are drawn.

Zoom in and zoom out do not seem to affect dots from showing up until I hit a threshold where the hashing of multiple dots comes into play.

crankycoder:
~~I'm pretty sure that this bug and #1087 are the same bug.~~

Nope, this ended up being a pixel grid hash bug.

garvankeeley:
@kbrosnan I think you'll see this fixed now, looks ok?

garvankeeley:
I'll be optimistic and close for now, please re-open if it happens again

PR #1135, opened 2014-10-28
1129 - add map.postInvalidate() on resume to force a redraw of the points

### LINKED ISSUE #1120: OutOfMemory errors on 0.42
I'm still getting OutOfMemory errors with v0.42.0.0

I don't think #1075 caused the OOM errors, this seems like a deep bug in osmdroid.

On investigation, the LRUMapTileCache is not a proper LRU, it doesn't implement ensureCapacity correctly and object eviction is incorrectly implemented.

I've got a patch in queue here https://github.com/crankycoder/MozStumbler/commit/db29ac57baab47e84e99ea90eb9687743da0ee72 which seems to keep my memory usage extremely low at ~37MB (compared to ~60-80MB).  I have not seen the memory usage _ever_ go above 40MB with this patch in place.

The OOM shows up mostly when tiles are switched over from low res to high res mode when going from cell to wifi data, but can also show up after pathologically viewing lots of high res map tiles.  Try scrolling across Canada for a test. 

@garvankeeley, this patch will need to also force garbage collection at the point in time when the tile layers are swapped between low and high resolution mode.

garvankeeley:
I can't repro, but this would make sense to fix #1120

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java
@@ -612,6 +612,7 @@ public final class MapFragment extends android.support.v4.app.Fragment
         setShowMLS(prefs.getOnMapShowMLS());
 
         mObservationPointsOverlay.zoomChanged(mMap);
+        mMap.postInvalidate();
     }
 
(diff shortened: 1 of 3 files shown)


crankycoder:
Do you mean #1129 ?

Issue #1142, opened 2014-10-28
Slowly pinch zooming out causes the viewport to not redraw properly


crankycoder:
![r3iq4j_jvtsvwrsxa_ertagd7rhtb9goutekmza11pc](https://cloud.githubusercontent.com/assets/4829/4813960/00eb9a9e-5ecd-11e4-929c-963bbe9a55d4.png)

PR #1143, opened 2014-10-28
1142 - fix partial zoom

crankycoder:
Fix for #1142 so that the viewport redraws on all zoom changes.


crankycoder:
There's comments in this commit which discuss how the snapping zoom level behavior has always existed, but was less obvious. https://github.com/mozilla/MozStumbler/commit/d57d43f90c9aa78081ca01b0726365e80f4dfd2b

android/src/main/java/org/mozilla/osmdroid/views/MapView.java
```diff
@@ -345,8 +345,7 @@ int setZoomLevel(final int aZoomLevel) {
                 IGeoPoint geoPoint = pj.fromPixels(snapPoint.x, snapPoint.y, null);
                 getController().animateTo(geoPoint);
             }
-
-            mTileProvider.rescaleCache(pj, newZoomLevel, curZoomLevel, getScreenRect(null));
+            invalidate();
```
crankycoder, line 348: This is where the magic happens.

garvankeeley:
Tested, can't repro the bug anymore.
(As an aside, I do hope we can get back in the removed code at some point, for rescaling the tiles until the new ones come in.)

PR #1146, opened 2014-10-29
update to 0.102

garvankeeley:


crankycoder:
The android/build.gradle should be updated to 0, 102, 0, 0

garvankeeley:
This is just for triggering an update to the public. The build .gradle should probably be at 1.0.0

Edit: set b.g to 0.102

PR #1161, opened 2014-10-29
Use scripts to get the versionCode and versionName from the git tag

keverets:
The version-code.sh script could be made more straightforward if a dependency on "sed" or "cut" is allowed, but a straight /bin/sh version works fine.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,8 +1,5 @@
-// each of the version numbers must be 0-99
-def versionMajor = 0
-def versionMinor = 42
-def versionPatch = 0
-def versionBuild = 2 // this is non-zero for any dev build
+def buildCode = file("../version-code.sh").toString().execute().text.trim().toInteger()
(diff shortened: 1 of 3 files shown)


crankycoder:
I've got a separate PR #1164 to check the version numbers are well formed (0-99 per section). 

I'd rather have the versionCode and versionName somewhere in source control other than the tag so that people can grab a tarball of the source and just replicate the binary for a given version.

We're almost certainly going to extend the build system to use the build flavor to target github, Google Play or F-Droid and stamp the release type in the versionCode.

I think it might make sense to just grab the versionName from the tag though, probably something like this: `git describe --tags --dirty --abbrev=12` and just stuff it verbatim into the versionName at build time.

So with all that, I think we'll try something different than what's in this PR.

keverets:
No worries.

It was just an option. As long as there's a single place to determine build details (that doesn't lead to inconsistencies), I'm happy.

PR #1173, opened 2014-10-31
fix Play Store feature filtering

### LINKED ISSUE #1150: nexus 7 2013 lte: can't install play store, make android.hardware.telephony required=false
I have a nexus 7 2013 lte.  The Updated Stumbler app will not install.  Build issue?  or is there another reason?

cascheberg:
This should fix #1150.
@garvankeeley Just saw that you found the same cause.

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -15,9 +15,9 @@
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
 
     <uses-feature android:name="android.hardware.location" />
-    <uses-feature android:name="android.hardware.location.gps" />
-    <uses-feature android:name="android.hardware.telephony" />
-    <uses-feature android:name="android.hardware.wifi" />
+    <uses-feature android:name="android.hardware.location.gps" android:required="true" />
+    <uses-feature android:name="android.hardware.telephony" android:required="false" />
+    <uses-feature android:name="android.hardware.wifi" android:required="false" />
 
     <application
         android:name=".client.MainApp"
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -15,9 +15,9 @@
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
 
     <uses-feature android:name="android.hardware.location" />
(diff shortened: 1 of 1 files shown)


android/src/main/AndroidManifest.xml
```diff
@@ -15,9 +15,9 @@
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
 
     <uses-feature android:name="android.hardware.location" />
-    <uses-feature android:name="android.hardware.location.gps" />
-    <uses-feature android:name="android.hardware.telephony" />
-    <uses-feature android:name="android.hardware.wifi" />
+    <uses-feature android:name="android.hardware.location.gps" android:required="true" />
+    <uses-feature android:name="android.hardware.telephony" android:required="false" />
+    <uses-feature android:name="android.hardware.wifi" android:required="false" />
```
garvankeeley: no wifi? I can't imagine we support a no wifi device

cascheberg: I don't know which devices those might be, but wouldn't it be the same as scanning with wifi disabled? Or is there a problem that will crash the app?

garvankeeley: Let's leave that one out since we are unsure of the implications.

android/src/main/AndroidManifest.xml
```diff
@@ -15,9 +15,9 @@
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
 
     <uses-feature android:name="android.hardware.location" />
-    <uses-feature android:name="android.hardware.location.gps" />
-    <uses-feature android:name="android.hardware.telephony" />
-    <uses-feature android:name="android.hardware.wifi" />
+    <uses-feature android:name="android.hardware.location.gps" android:required="true" />
```
garvankeeley: to keep changes to a minimum, revert this line, that is the default, thanks

android/src/main/AndroidManifest.xml
cascheberg: It would crash... could be fixed of course in case that is needed at some point.
Updated push to minimum changes.

garvankeeley:
Looks good!

Djfe:
@cascheberg is Wi-Fi now required or not?

garvankeeley:
wifi is required

PR #1179, opened 2014-11-01
Always update metrics last upload time

cascheberg:
The last upload time shown in metrics is not an absolute value anymore, so it needs to be updated continuously.

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/navdrawer/MetricsView.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/navdrawer/MetricsView.java
@@ -53,9 +53,13 @@ public class MetricsView {
 
     private WeakReference<IMapLayerToggleListener> mMapLayerToggleListener = new WeakReference<IMapLayerToggleListener>(null);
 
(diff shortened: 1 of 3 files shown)


garvankeeley:
Is there a solution that doesn't require reading the sync stats from disk. This updating of the last upload time could be separate from updateSyncStats() function entirely, as is isn't related to sync, it is purely a UI thing, updating a label based only on the existing value of the label and the current time.

cascheberg:
I see. I'm sure I can improve this then. How about using `final long bytesUploadedThisSession = AsyncUploader.sTotalBytesUploadedThisSession.get();` to check for updates (as before) and then caching the `lastUploadTime`?
I think this method is called whenever metrics are updated, so this includes updates on cell and wifi scans. That could also be improved if still necessary.

garvankeeley:
If the scanning isn't running, the label should still update. How about a posting an event every 10 seconds (handler.postDelayed) to update the label, updateLastUploadedLabel() which will update based on mLastUploadTime

cascheberg:
I think I got it, PR updated

garvankeeley:
Nice! thanks

Issue #1169, opened 2014-10-30
ACRA Report: NPE WifiScanner.java:98


garvankeeley:
The code is:
  ArrayList<ScanResult> scanResults = new ArrayList<ScanResult>();
  for (ScanResult scanResult : getScanResults()) {

We can add if scanResults == null return; 
Note this code is in response to a WifiManager.SCAN_RESULTS_AVAILABLE_ACTION, so by definition the scan results shouldn't be null.

java.lang.NullPointerException at org.mozilla.mozstumbler.service.stumblerthread.scanners.WifiScanner.onReceive(WifiScanner.java:98)
at android.app.LoadedApk$ReceiverDispatcher$Args.run(LoadedApk.java:709)
at android.os.Handler.handleCallback(Handler.java:587)
at android.os.Handler.dispatchMessage(Handler.java:92)
at android.os.Looper.loop(Looper.java:130)
at android.app.ActivityThread.main(ActivityThread.java:3844)

garvankeeley:
This a 2.3.4 android, running 1.0.0.0
VERSION {CODENAME:REL, RELEASE:2.3.4, INCREMENTAL:110922, SDK_INT:10, SDK:10, SYSTEM:Blur_Version.45.31.0.M...

PR #1192, opened 2014-11-03
1169: ACRA Report: NPE WifiScanner.java:98

garvankeeley:
#1169

Added null pointer checks to getting wifi manager and scan result. In theory neither of these should have ever been null.


crankycoder:
This bug amuses me.  Oh Android.

Djfe:
Lol :)
Probably more like, oh OEMs. ^^

Issue #1170, opened 2014-10-30
ACRA report: provider doesn't exist: gps


garvankeeley:
We could catch this I think

java.lang.IllegalArgumentException: provider doesn't exist: gps
at android.os.Parcel.readException(Parcel.java:1469)
at android.os.Parcel.readException(Parcel.java:1419)
at android.location.ILocationManager$Stub$Proxy.requestLocationUpdates(ILocationManager.java:540)
at android.location.LocationManager.requestLocationUpdates(LocationManager.java:860)

Sony Xperia S
{CODENAME:REL, RELEASE:4.4.4, INCREMENTAL:37, SDK_INT:19, RESOURCES_SDK_INT:19, SDK:19}

PR #1193, opened 2014-11-03
1170 ACRA report: provider doesn't exist: gps

garvankeeley:
#1170

Wrap the startScanning in a catch for IllegalArgumentException (which is thrown when no GPS is available).


crankycoder:
Catching an exception here doesn't make much sense to me - we shouldn't even provide the ability to start the stumbling if no GPS is available.

People installing from the Play store are already filtered out if they don't have GPS, let's just add a warning message on startup if we don't detect a GPS.  If the app crashes and burns after that, I think that's actually acceptable behavior if the user has been warned "This application does not function without a GPS."

garvankeeley:
adding test case

Issue #1185, opened 2014-11-03
ACRA report: NPE Reporter.startup:64


garvankeeley:
line is:
TelephonyManager tm = (TelephonyManager) mContext.getSystemService(Context.TELEPHONY_SERVICE);

PR #1194, opened 2014-11-03
1185 NPE Reporter.startup (no telephony manager)

garvankeeley:
#1185

If TelephonyManager is null, just return.


cascheberg:
Why not just set `mPhoneType = TelephonyManager.PHONE_TYPE_NONE` to still allow wifi scanning?

garvankeeley:
Good point, will change.

Issue #1190, opened 2014-11-03
ACRA: NPE FirstRunFragment$1.onClick(FirstRunFragment.java:50)


crankycoder:
Samsung GT-S5570 running 2.3.6.  Stumbler v1.0.0

```
java.lang.NullPointerException
at org.mozilla.mozstumbler.client.subactivities.FirstRunFragment$1.onClick(FirstRunFragment.java:50)
at android.view.View.performClick(View.java:2485)
at android.view.View$PerformClick.run(View.java:9080)
at android.os.Handler.handleCallback(Handler.java:587)
at android.os.Handler.dispatchMessage(Handler.java:92)
at android.os.Looper.loop(Looper.java:130)
at android.app.ActivityThread.main(ActivityThread.java:3687)
at java.lang.reflect.Method.invokeNative(Native Method)
at java.lang.reflect.Method.invoke(Method.java:507)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:867)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:625)
at dalvik.system.NativeStart.main(Native Method)
at android.app.ActivityThread.access$1300(ActivityThread.java:141)
at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1316)
at android.os.Handler.dispatchMessage(Handler.java:99)
at android.os.Looper.loop(Looper.java:137)
at android.app.ActivityThread.main(ActivityThread.java:5116)
at java.lang.reflect.Method.invokeNative(Native Method)
at java.lang.reflect.Method.invoke(Method.java:525)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:787)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:600)
at dalvik.system.NativeStart.main(Native Method)
at android.os.Parcel.readException(Parcel.java:1431)
at android.os.Parcel.readException(Parcel.java:1385)
at android.app.ActivityManagerProxy.registerReceiver(ActivityManagerNative.java:2181)
at android.app.ContextImpl.registerReceiverInternal(ContextImpl.java:1361)
at android.app.ContextImpl.registerReceiver(ContextImpl.java:1329)
at android.app.ContextImpl.registerReceiver(ContextImpl.java:1323)
at android.content.ContextWrapper.registerReceiver(ContextWrapper.java:446)
[...]

PR #1196, opened 2014-11-03
1190 NPE FirstRunFragment:50

garvankeeley:
#1190

This line number is where getDialog() is assumed to return a non-null, fixed.


crankycoder:
There's a lot of stuff in this PR that seems unrelated to #1190. Can we get those removed?

garvankeeley:
oops.

Issue #893, opened 2014-09-29
Let's move the upload button as well as the settings


crankycoder:
I'd like to move the upload button and the settings to the bottom of the metrics.  

Like what Pocketcasts does:

![xxernkibiyjl2duxxqscgysm2stddjotlvcg0vnvbvi](https://cloud.githubusercontent.com/assets/4829/4443561/4a83d496-47e9-11e4-9b71-640d97144b7c.png)

garvankeeley:
@ryanfeeley is not a fan of the current upload button either. (nor am I, but leaving as-is for 1.0).

Djfe:
I fav this as well.
Upload + statistics in the left sidebar would look great and you can access them more easily from the map/main activity by just swiping from the left of the device.

rodmoreno:
@crankycoder maybe this should close.

cascheberg:
@crankycoder Reminder: This issue was fixed and should be closed.

PR #1197, opened 2014-11-04
 #893 Drawer: uploads+settings as non-scrolling footer row

santiago-hollmann:
# Description

The upload reports button was added to the bottom of the Drawer together with the preferences button.

This feature is working on portrait and landscape mode as well (in this case, the the report is scrollable). This bottom buttons on the drawer remain always visible. Below you can find screenshots of both modes.
# Implementation
### Layout design explanation

On the xml file (fragment_metrics_drawer.xml), the ScrollView was moved inside a LinearLayout with vertical orientation.

The only thing that changed on the ScrollView is that now it has a weight of 10.

Below of the ScrollView there is a RelativeLayout (aka "buttons container") which has the upload button and the preferences button. This RelativeLayout has a weight of 1 and its buttons are established one on the left side and the other on the right one according to the bug #893.

Also, following the AC of #893, the preferences menu item was removed.
### Java code explanation

Regarding to the code, the references to the preferences menu item were removed.

Also, the preferences button was added to the MetricsView with its click listener. Besides that, the click listener was removed of the drawer buttons container because when the drawer was open, if you touched the container, the event went directly to the map view which leaded on a wrong behavior.
# Screenshots

screenshot_2014-11-03-22-40-43
screenshot_2014-11-03-22-40-51

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/navdrawer/MainDrawerActivity.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/navdrawer/MainDrawerActivity.java
@@ -131,8 +131,6 @@ public class MainDrawerActivity
         }
         else {
             MenuItemCompat.setShowAsAction(mMenuItemStartStop, MenuItemCompat.SHOW_AS_ACTION_ALWAYS);
-            MenuItem item = menu.findItem(R.id.action_preferences);
-            MenuItemCompat.setShowAsAction(item, MenuItemCompat.SHOW_AS_ACTION_ALWAYS);
         }
 
         updateStartStopMenuItemState();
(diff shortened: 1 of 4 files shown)


crankycoder:
@shollmann this looks great!  Going to review this in the morning.

garvankeeley:
Awesome!
 Once https://github.com/mozilla/MozStumbler/pull/1199 lands, we can put the last uploaded time as a textview (with no label) beside the upload button.

crankycoder:
r+

@garvankeeley I think we can merge this without blocking on #1199

santiago-hollmann:
Thanks guys!

I also think we can merge this without blocking on #1199.  Regarding to #1199, it is almost done and plurals on android are easy to implement.

My only concern about adding the last upload time at bottom is that lack of space we have between both buttons. Maybe we should use a pretty date and text like this: 
- Last upload 2 m ago
- Last upload 1 h ago
- Last upload 3 d ago
- Last upload 2 months ago: This is the longest string but, we hope our users to come often, right? ;)

But for now, I would keep the last upload time right where it is.

garvankeeley:
Just the time next to the upload button (which should change to a sync icon)
[button] 3 hours ago [button]
I think we have an example app that has this, let me look...

garvankeeley:
I see, you already saw it on #893. That one has a label with the time, I wonder can we fit:
[button] Uploaded 10 minutes ago [button]

Regardless, merging, we can change that in a later PR.

PR #1198, opened 2014-11-04
Update translations

cascheberg:
This is a minor update regarding German translations (mostly).
I also added a new `keep_screen_on_summary` string, so that you do not necessarily need to find a perfect and short `keep_screen_on_title`, is that ok?


garvankeeley:
After applying this patch, I didn't see a difference in the Settings view, I was expecting a change keep_screen_on_summary to show up, no?
screenshot 2014-11-04 10 31 12

cascheberg:
Indeed... for me it looks like this:
screenshot_2014-11-04-16-52-17

cascheberg:
@garvankeeley I had also changed the english text for `enable_option_show_mls_on_map` (because in fact, it does not just enable some option but it enables the actual queries) which does not show up on your screenshot. Can you check this again?

crankycoder:
@garvankeeley I can see the text, so I'm guessing something is wrong with your build.

garvankeeley:
@crankycoder thanks for checking

Issue #1066, opened 2014-10-22
Last Upload Time displaying a time in zh-CN script in zh-TW locale (l12y?)


petercpg:
Hi, 
I just found the time in "Last Upload Time:" are not localizable so it displays values as "never", "片刻之前", "19 小时 前" (19 hours ago in Simplified Chinese) rather than strings fully localized in Traditional Chinese. Probably a library issue?

Screenshot: http://imgur.com/dUA1sHq

Tested in 0.41.0.0 on Asus Zenfone 5 LTE (Android 4.4.2 KOT49H) and Nexus 7 2013 (Android 5.0 LPX13D)

cascheberg:
That is a library named `PrettyTime`.
There is a problem with the German translation as well.
I prefer an accurate date to figure out the time diff myself, so I think one might just as well remove that formatting. @garvankeeley ?

![bug](https://cloud.githubusercontent.com/assets/3845150/4734865/f5267222-59d8-11e4-8274-0c63ab37eaa8.png)

garvankeeley:
I like the PrettyTime, and we should investigate the translation issue, both for missing translations and for  fitting the translation better. For now I am just going to do:

if (Locale.getDefault().getLanguage().equals("en")) {
                    lastUploadString = new PrettyTime().format(new Date(lastUploadTime));
 } else {
                    lastUploadString = DateTimeUtils.formatTimeForLocale(lastUploadTime); 
 }

petercpg:
I read the time messages used by PrettyTime at [Here](https://github.com/ocpsoft/prettytime/tree/master/core/src/main/java/org/ocpsoft/prettytime/i18n) and it seemed no problem. 
Could that be a locale fallback issue?

garvankeeley:
We should ensure we are using the latest version. The version we get is from the Maven repo, which is not always up-to-date.

cascheberg:
Wouldn't it be just a few lines of code to check for
- x seconds
- x minutes
- x hours
  else: display just the date

cascheberg:
@petercpg "Last Upload Time" strings should now be fully localizable. Please check the translation and close this issue.

petercpg:
Will close this issue as soon as #1373 is merged. :)

PR #1199, opened 2014-11-04
Remove ocpsoft PrettyTime dependency

cascheberg:
Implement custom `PrettyTime` replacement, see #1066 
@garvankeeley What do you think about this?


hannosch:
It looks like you are providing a singular / plural message for each of the "pretty times". Unfortunately different languages have very different rules on how to change a text based on the number of items they refer to. Many languages have multiple plural forms, with complex rules for when to apply them.

If you want to get into the details, the gettext suite and their po/mo format solves a lot of these. See for example https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html. I'm not really familiar with Android development, so I'm not sure if there's a way to encode these sort of complex plural forms into the translation system.

hannosch:
And of course a simple google search reveals: http://developer.android.com/guide/topics/resources/string-resource.html#Plurals

cascheberg:
Thanks for the information!
I was wondering if that could be a problem, but I checked and PrettyTime is not supporting this either (or I just totally missed it).
Update: That should be easy to implement if the push is ok in general.

garvankeeley:
+1 to this.
Can you add plurals.xml (or whatever it is called) for english so others will see how to add plurals for other languages.

cascheberg:
PR updated to use Android quantity strings for plural forms.
Note that the feature is still only enabled for `en` locales.

garvankeeley:
Tested, works great. Can you merge/rebase and I'll land it? Thanks.

cascheberg:
Merged

PR #1206, opened 2014-11-04
1170 ACRA report: provider doesn't exist: gps

garvankeeley:
#1170

Check if the provider is available, if not, log  message to the GUI log.

Replacing pr #1193 with this one


garvankeeley:
@crankycoder updated

PR #853, opened 2014-09-17
Features/767 remove scan wifi

crankycoder:
Fix for #767 to drop the Wifi Scan Always preference.

PR #1218, opened 2014-11-05
Fix major WifiScanner bugs

### LINKED ISSUE #1216: bug wifi always on
Wi-Fi always on doesn't seem to work anymore...
As far as I know Mozilla Stumbler should try to scan and scanning should work with that option on but wifi on.

It shows 0 wifis if wifi is deactivated on app start
if I switch it on for a short term only then it shows a number of wifis, but that number only goes down after that (as if the system or the app caches the wifis)

cascheberg:
@Djfe good catch!
This fixes #1216, the feature was removed in #853

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiScanner.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiScanner.java
@@ -82,7 +82,9 @@ public class WifiScanner extends BroadcastReceiver {
             activatePeriodicScan(stumblingMode);
         }
 
-        IntentFilter i = new IntentFilter(WifiManager.SCAN_RESULTS_AVAILABLE_ACTION);
+        IntentFilter i = new IntentFilter();
+        i.addAction(WifiManager.WIFI_STATE_CHANGED_ACTION);
+        i.addAction(WifiManager.SCAN_RESULTS_AVAILABLE_ACTION);
         mContext.registerReceiver(this, i);
     }
 
--- a/android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiScanner.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiScanner.java
@@ -11,6 +11,7 @@ import android.content.IntentFilter;
 import android.net.wifi.ScanResult;
 import android.net.wifi.WifiManager;
 import android.net.wifi.WifiManager.WifiLock;
+import android.os.Build;
 import android.support.v4.content.LocalBroadcastManager;
 import android.util.Log;
 
@@ -53,8 +54,13 @@ public class WifiScanner extends BroadcastReceiver {
         mContext = c;
     }
 
-    private boolean isWifiEnabled() {
-        return getWifiManager().isWifiEnabled();
+    private boolean isScanEnabled() {
+        WifiManager manager = getWifiManager();
+        boolean scanEnabled = manager.isWifiEnabled();
+        if (Build.VERSION.SDK_INT >= 18) {
+            scanEnabled |= manager.isScanAlwaysAvailable();
+        }
+        return scanEnabled;
     }
 
     private List<ScanResult> getScanResults() {
@@ -72,7 +78,7 @@ public class WifiScanner extends BroadcastReceiver {
         }
         mStarted = true;
 
-        if (isWifiEnabled()) {
+        if (isScanEnabled()) {
             activatePeriodicScan(stumblingMode);
         }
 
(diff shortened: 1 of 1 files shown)


cascheberg:
Fixed another major bug.
The `WifiScanner` did not start `activatePeriodicScan` if the wifi status changed to enabled while the app was already running. Bad is that you do not even notice this in the GUI, because Android will report `WifiManager.SCAN_RESULTS_AVAILABLE_ACTION` at least once after that, which is received by the app.

Djfe:
did your second commit fix the issue I described here #974 in the first (starting) comment after the "Edit:"?

cascheberg:
@Djfe Yes it does. It probably would have been fixed earlier if you had filed a separate bug about it ;)
The code was there, it just wasn't hooked up properly. I even think this regressed at some point.

Djfe:
does mozilla stumbler listen to the mobile connection properly? (turn off airplane mode after scanning gets started)
I haven't tested it

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiScanner.java
```diff
@@ -53,8 +54,13 @@ public WifiScanner(Context c) {
         mContext = c;
     }
 
-    private boolean isWifiEnabled() {
-        return getWifiManager().isWifiEnabled();
+    private boolean isScanEnabled() {
+        WifiManager manager = getWifiManager();
+        boolean scanEnabled = manager.isWifiEnabled();
+        if (Build.VERSION.SDK_INT >= 18) {
```
garvankeeley, line 60: !scanEnabled && Build.VERSION.SDK_INT >= 18

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiScanner.java
```diff
@@ -53,8 +54,13 @@ public WifiScanner(Context c) {
...
+            scanEnabled |= manager.isScanAlwaysAvailable();
```
garvankeeley, line 61: scanEnabled =

garvankeeley:
The fix looks good, thanks!

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiScanner.java
```diff
@@ -72,11 +78,13 @@ public synchronized void start(final ActiveOrPassiveStumbling stumblingMode) {
...
 
-        if (isWifiEnabled()) {
+        if (isScanEnabled()) {
             activatePeriodicScan(stumblingMode);
         }
 
-        IntentFilter i = new IntentFilter(WifiManager.SCAN_RESULTS_AVAILABLE_ACTION);
+        IntentFilter i = new IntentFilter();
+        i.addAction(WifiManager.WIFI_STATE_CHANGED_ACTION);
+        i.addAction(WifiManager.SCAN_RESULTS_AVAILABLE_ACTION);
```
garvankeeley, line 87: I didn't understand your comment. Without listening for SCAN_RESULTS_AVAILABLE_ACTION, what happens?

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiScanner.java
cascheberg, line 60: If I do this, Android Studio does not recognize the SDK version check and warns about using the function, which isn't really nice. It seems to recognize it only if the version check is the only condition in the `if` statement.
Since this is not a hot function, I left it like this. Having two nested `if`s would also work, but does not look nice. Or just not care about Android Studio and use your suggestion anyway?
as

garvankeeley, line 60: Weird! Leave as-is then.
This code has to go through a second review when I put it into mobile Firefox, and they will question that code, at least I know why it is written that way :)

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiScanner.java
cascheberg, line 87: The actual problem was not listening for `WIFI_STATE_CHANGED_ACTION`.
My comment was about the fact that the app will receive at least one `SCAN_RESULTS_AVAILABLE_ACTION` anyway and will update the wifi count display accordingly. As a result, it looks like the app has started scanning actively, while in fact it has not.

garvankeeley, line 87: Oh right! I didn't notice that in the diff, that we were _only_ listening for SCAN_RESULTS_AVAILABLE_ACTION and _not_ the state changed.

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiScanner.java
Djfe, line 61: nope "|=" is correct or else it will overwrite with false if manager.isScanAlwaysAvailable() is false but wifi is true...

garvankeeley, line 61: see above, it would go with: 
!scanEnabled && Build.VERSION.SDK_INT >= 18

Issue #1019, opened 2014-10-16
Show number of unique cells/wifis scanned during session in metrics


cascheberg:
Those numbers were previously displayed on the main screen and are not available anymore

Djfe:
They got removed to stop displaying unnecessary metrics (could cause confusion if there are two different numbers and you should follow DRY)

cascheberg:
Why did you decide to not hide the other values instead then? I myself found the unique numbers to be more interesting and helpful, and I think they are easier to understand in general.

Djfe:
I didn't decide that that were the devs, I'm just a user of MozStumbler that is interested in open source and wants to contribute later into the project like you do already. I'm still learning Java and I need to look into developing for Android still.

The reason is simple, the other (old) numbers were counting wifi/cell sights that were recorded without a valid/established GPS connection -> so they won't be uploaded

the users would wonder when their sent numbers and score in the leaderboard would be rising slower then the number of recorded wifis and cells

so it makes sense to only show the numbers that will be uploaded later on if there is an internet connection available

cascheberg:
I wasn't necessarily referring to you personally, but to everyone involved :) I was also asking because I could not find the specific relevant push for that change.

So maybe it's a good idead to keep in mind to fix that bug. I think it's not easily possible to compare the non-unique values to leaderboard points at all.

Djfe:
then the wifi/aps should be shown like 5/151 over the map next to the specific icon

took some time but I found the issue about it:
#917

cascheberg:
@garvankeeley I'd be willing to look into fixing this (#132 for unique values) if there is a chance for this to be added back to the UI.
Having different values based on different ways of counting is not great, but I still think the more intuitive and meaningful thing to keep (per default) is to count unique networks.

Djfe:
the numbers on the left are unique already (got changed some time ago from countings to unique numbers)
the only difference is that the left only counts saved wifi/cell networks
and the other numbers include wifi/cell networks that were recorded without a valid GPS signal, which would be too confusing IMO)

what about a devs option to enable showing them?

Djfe:
explained at #1027 that it actually makes sense to show them again because the numbers decrease if you upload during scanning so that ready to sent doesn't represent this session anymore

haven't thought about it before because I was only uploading over wifi -> there was no difference noticeable to me

PR #1221, opened 2014-11-06
Show unique number of networks in metrics

### LINKED ISSUE #132: We should only increase WiFi access point count when we have a GPS fix
This will prevent confusion like issue #131 where the number of WiFi access points appears large, but the number of locations is small or zero.

cascheberg:
This PR is a bit larger, I am looking forward to get your feedback.
I read a German article about the Mozilla Stumbler release, and the author wrote that he had stumbled over 900 wifis in a really small area, so I remembered #1019.

It's best to check each commit separately:
- first commit does some cleanup and renames some variables, no functional change.
- second commit makes the `Reporter` count the observations, instead of the `MainApp`
- third commit counts the unique networks in the `Reporter`, I guess it fixes #132
- fourth commits displays the unique network stats in the metrics (see first screenshot)
- last commit removes the other network counts (see final screenshot)

This was one of my tests to see that the code works. I did five scans on the same position (6 cells, 7 wifis visible), only two of the scans had wifi enabled:
screenshot_2014-11-07-00-15-45

This is the final version:
screenshot_2014-11-07-00-25-24

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/MainApp.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/MainApp.java
@@ -59,15 +59,13 @@ import java.util.concurrent.atomic.AtomicBoolean;
     formUriBasicAuthLogin = BuildConfig.ACRA_USER,
     formUriBasicAuthPassword = BuildConfig.ACRA_PASS)
 public class MainApp extends Application
-        implements ObservedLocationsReceiver.ICountObserver,
-        AsyncUploader.AsyncUploaderListener {
+        implements AsyncUploader.AsyncUploaderListener {
     public static final AtomicBoolean isUploading = new AtomicBoolean();
     private final String LOG_TAG = AppGlobals.LOG_PREFIX + MainApp.class.getSimpleName();
     private ClientStumblerService mStumblerService;
     private ServiceConnection mConnection;
     private ServiceBroadcastReceiver mReceiver;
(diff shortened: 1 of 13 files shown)


crankycoder:
Let's at least remove the last commit.  It's unrelated to showing unique wifi access points and clutters up the PR for review.

cascheberg:
Do you want to remove aacf77b942d41c67e84d37d86ca54018d0b99fde?
Can you please try to review each commit separately then? I paid attention to having small commits, and it would be really great if I did not have to revert things except for real problems.

crankycoder:
Yes.  I think it makes sense to cherry-pick that one commit and put it into a different branch for now.  This is mostly just so that we can have our UX people take a look at aacf77b and approve that change.

garvankeeley:
The fewer numbers the better, we need to make room for more interesting options for the user, such as power management, bandwidth management (the only option we have ATM is tiles default/low/high).
@ryanfeeley can you eyeball this screenshot

garvankeeley:
@cascheberg can you split out the "Previous Sessions" -> "Already Sent" change to a separate bug. I can then assign that to UX review, since they created those titles originally.
Other than that LGTM!

cascheberg:
Of course, thanks! Review comments addressed in d41782db2fbd55cec9378c908ecf8875e641f3fe

cascheberg:
Resolved merge conflicts

garvankeeley:
Thanks, looks good.

PR #1225, opened 2014-11-07
Always clear cells after flush

cascheberg:
If there are no cells in the current scan and there was a bundle flush, clear cells before returning early.
Otherwise number of cells will never be zero.


garvankeeley:
LGTM

PR #1226, opened 2014-11-07
Listen for GPSScanner.SUBJECT_LOCATION_LOST

cascheberg:
There is no listener for `GPSScanner.SUBJECT_LOCATION_LOST`. Is this ok?


garvankeeley:
Makes sense to do this. I am sure the code did this at one point, and this got unhooked when we moved code around

PR #1229, opened 2014-11-08
Build System: parallel install of different app versions

santiago-hollmann:
# Overview

By adding ".debug" as applicationIdSuffix to the build type debug, what we do is to change the app package name for the debug build so we can have installed the debug and production version at same time.

This is a quick fix until the beta program is defined.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -97,6 +97,10 @@ android {
 
     buildTypes {
         debug {
+            // We add this suffix to the pakage name to be able to have production and dev version
+            // at same time
+            applicationIdSuffix '.debug'
+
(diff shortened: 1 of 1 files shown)


Djfe:
this wouldn't be a quickfix unless the beta program uses a different package name, too.

Before beta updates had the same signing key and could be installed on top of the old versions without problems.

The reason it doesn't work probably is that the build system isn't setup for a different package name, you should probably leave this change to @crankycoder or @garvankeeley
they have set it all up and know what's the best way to do it

santiago-hollmann:
Thanks @Djfe I will :)

Issue #836, opened 2014-09-13
Auto-follow mode: indicate using icon that this is on/off


garvankeeley:
This is a follow-up item from #834.
Ideally, the icon would be a different style/color when auto-following the users GPS position.

Djfe:
This would definitely be better for a simple reason:
the GPS follow icon is not shown if there is no GPS signal
but also if the map follows the user's GPS position

if the GPS signal is there but the map doesn't follow the user's position then the button isn't shown

Djfe:
I tested Version 1.4.0 and I like the approach there:
an empty GPS icon looks pretty nice for the follow mode being turned off

cascheberg:
@garvankeeley This issue was fixed in #1248 and should be closed.

PR #1231, opened 2014-11-09
Only display the "Center Me" button when the view is not centered

keverets:
There's no need to have the button there obscuring the map when it would have
no effect to press it.


Djfe:
it would be better to display a lock icon instead so that users know that the map follows your position
#836

garvankeeley:
Thanks Kevin we can expand on this to do @djfe request where we change the icon when locked.

PR #1232, opened 2014-11-10
Update Marathi translation

vinitraje:
Added new strings and refined old ones :)


vinitraje:
Don't merge. Missed a line. :P

PR #1234, opened 2014-11-10
Update Hindi & Marathi translations

vinitraje:
Refined old strings and added missing strings


vinitraje:
Again missed a line. :(

PR #1237, opened 2014-11-11
Additional cleanup: Create isSupported() check on the cell scanner implementation

garvankeeley:
isSupported() checks the implementation to see if it usable. The wraps up some repeated checks the code was doing.


garvankeeley:
a backport from Fennec

Issue #1202, opened 2014-11-04
Capture buildhost in versionName


crankycoder:
We may be receiving crash reports from custom builds of MozStumbler. We should include the build host name in the versionName or somewhere else that we can inspect within an ACRA crash log.  

We should be able to do something like:

``` java
InetAddress.getLocalHost().getHostName();
```

and stuff it into the versionName.

aishraj:
I'd like to work on this. Is there a way to assign the issue to me ?
Also what sort of build version hostname combination are we looking? Is it simply something like MozStumbler Version X.X - hostname.domain.org ?

aishraj:
@crankycoder I've added the buildHostname in #1238 . Please let know if it requires any further changes.

PR #1238, opened 2014-11-11
Updating build.gradle as per  #1202

aishraj:
Making the change as per #1202 Capture buildhost in versionName
Please review it accordingly. From my understanding, the Updater which uses the versions seems to work fine. Kindly suggest if anything else might be under the scope of this change .

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -2,6 +2,7 @@
 def versionMajor = 1
 def versionMinor = 3 // minor feature releases
 def versionPatch = 0 // This should be bumped for hot fixes
+def buildHost = InetAddress.localHost.hostName
 
 // Double check the versioning
 for (versionPart in [versionPatch, versionMinor, versionMajor]) {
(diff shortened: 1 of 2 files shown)


crankycoder:
I think this needs some changes to the Updater.java code as it's naive and assumes integers in each part of the version name. 

The use of a raw hostname is probably unsafe as well, I think we want to restrict the characters with a regular expression to just [a-z] and maybe [_-].

aishraj:
@crankycoder I've added a few rudimentary checks for sanitizing the hostname check in both the Updater and the gradle config. Let me know if it looks good to you.

garvankeeley:
make debug now fails for me:
- Where:
  Build file '/Users/gkeeley/c/github/mz-passive-stumbler/MozStumbler/android/build.gradle' line: 5
- What went wrong:
  A problem occurred evaluating project ':android'.
  
  > gkeeley13311: gkeeley13311: nodename nor servname provided, or not known

aishraj:
I have not attempted to reproduce the behaviour yet. However it looks like it may be related to a bug while performing hostname lookup on OSX with JDK 7 [link here](https://coderwall.com/p/ktbkea/play-framework-2-error-nodename-nor-servname-provided-or-not-known) I will  nevertheless try to reproduce it on my Linux machine and provide an update on it later.

crankycoder:
Hotfix applied in https://github.com/mozilla/MozStumbler/commit/ec3f306b0ed8b18888c0290de54e775b58f5e1e6

PR #1239, opened 2014-11-11
Fix NPE startup crash

cascheberg:
@garvankeeley #1237 introduced a startup NPE, this PR restores the initialization of `mTelephonyManager`.


garvankeeley:
Thanks for the fix!

PR #1246, opened 2014-11-11
Fix untruncated log feature

cascheberg:
@garvankeeley The feature added in #1117 is not working because it is comparing a String to a char.


garvankeeley:
Thanks!

PR #1247, opened 2014-11-12
Add green dot in map leyend

### LINKED ISSUE #1166: Key is missing green dots
On my maps I see bright green dots that are not shown on the key. My guess is that they are cell + Wi-fi as I have not seen any wi-fi only (magenta) dots

rodmoreno:
Fix #1166


garvankeeley:
Thanks, I know this was confusing some users that we didn't have it.

Issue #1109, opened 2014-10-24
Battery Saving: minimum battery level setting (currently hard-coded to 15%)


garvankeeley:
Have an option to set the minimum level to any value, for me, it would be 50%. 
I charge my phone daily, and don't mind committing the first 50% of consumption to the app, but after that, I want the rest of the charge for other purposes.

Djfe:
Thx, was hoping for this!
Finally a feature that allows me to disable the battery limit ;)
0% -> it stops only while my device is shutting down

I stumble only actively
So this was a bit stupid if I was charging my device in my car while stumbling and my device was low on battery (after a day of work/studying/whatever)
-> stumbling stopped when charging stopped
Scenario:
I connect my device to the charging cable
I start the engine of my car, which briefly connects and disconnects the power source so thaf I have to start stumbling manually again (listener:is on battery && is battery% below 15%)

garvankeeley:
#1261 fixes this

PR #1250, opened 2014-11-12
Implement custom low battery check

cascheberg:
This implements a custom low battery check which hopefully can be used to implement #1109.
I added vibration to notify the user. It would also be nice to add a notification.


garvankeeley:
I think this is going to conflict with battery-related work on my branch. I'll look at this tomorrow.

garvankeeley:
The manifest-based low battery intent receiver was removed as it only notifies on a fixed value (15%):
https://github.com/mozilla/MozStumbler/pull/1261

I think vibrations are typically more actionable (answer a call, check a text or email) than this use case, I'm not sure what a user is to do when they get this vibration. Maybe file a bug and see if others like the idea more than me?

PR #1251, opened 2014-11-12
Move prefs for crash reporting and tile detail to main settings screen

garvankeeley:


garvankeeley:
screenshot 2014-11-12 17 16 40

garvankeeley:
screenshot 2014-11-12 17 18 15

garvankeeley:
Added preference categories (2.3 screenshot):
screenshot 2014-11-12 18 35 36

cascheberg:
Very nice update!

Some afterthoughts:
- Can you add a title for the crash reporting? On Android 4 the separating spaces are all black (see screenshot below).
- I was thinking that location queries probably also cost quite a bit of energy because it will send a single request every few seconds. Worth mentioning (and worth improving)?
- The "upload only on Wi-Fi" pref also affects MLS fetching, which is ok, but not obvious.

settings

garvankeeley:
- Can you add a title for the crash reporting? On Android 4 the separating spaces are all black (see screenshot below).

I didn't know what title to use, but yeah, it looks rather lonely sitting there.
- I was thinking that location queries probably also cost quite a bit of energy because it will send a single request every few seconds. Worth mentioning (and worth improving)?

This feature is pretty advanced, so I am not too concerned, but yes it would do this. Good point about aggregating these queries to reduce the cost of powering up the network. We should try get power measurements for this and see what the cost is. 
- The "upload only on Wi-Fi" pref also affects MLS fetching, which is ok, but not obvious.

Good point, any suggestion where to put that?

cascheberg:
> I didn't know what title to use, but yeah, it looks rather lonely sitting there.

Not sure either, just "Crash reports", or something in the direction of "Feedback" / "Support". Firefox has "Data Choices".

> Good point, any suggestion where to put that?

I think rewording the wifi pref is a bit of a hassle. How about this for the detailed info?

> At the cost of a bit more energy, memory and network usage, the MLS location is shown for each report. This is shown on the map as a red dot for the MLS location, with a line to its report. However, it will not query locations over mobile data connections if \"%1$s\" is enabled.

garvankeeley:
Cool, filed https://github.com/mozilla/MozStumbler/issues/1257 so we don't forget to make those changes

Issue #1176, opened 2014-10-31
Write to disk on low memory notification from OS


garvankeeley:
We aren't handling Application::onLowMemory() onTrimMemory(), and we should.

PR #1252, opened 2014-11-12
1176 Write stumbles to disk on low memory notification

garvankeeley:
#1176

Handle OS low memory notifications and call DataStorageManager saveCurrentReportsToDisk()


crankycoder:
Is there a way to test this at all using adb?

garvankeeley:
There doesn't appear to be a way to directly send low memory in adb .

SteveALee:
How about 'adb shell' and then run 'ulimit' ?

garvankeeley:
Great suggestion! will try...

PR #1253, opened 2014-11-12
Update translations

### LINKED ISSUE #1240: Change wording of title "Previous sessions" in metrics
The title "Previous sessions" is not quite accurate, since in fact it shows the number of reports uploaded, and you can already upload reports multiple times within the current session.
Changing the title to e.g. "Already sent" (or "Sent (total)") would give a clear distinction between current session data and uploads (sent/ready to send).
I can create a PR if the change is ok.
metrics
The metrics were updated in #1221

cascheberg:
This fixes #1240


l-hedgehog:
Hi @cascheberg, thx for fixing my mistakes in zh-CN strings!

PR #1254, opened 2014-11-13
Update Hindi & Marathi translations

vinitraje:
Updated string "Previous Sessions" to "Already Sent"


crankycoder:
thanks @vinitraje !

Issue #1255, opened 2014-11-13
Add option to move stumble files into a user defined directory.


crankycoder:
While the stumbler is active, we currently save our stumble records as gzipped JSON data into persistent storage.

After we upload, we delete those files.
#3 is a request for a local database so that users can export their data.
#979 is a bug that's real intent is to improve debugging of the stumbler in real world situations.

I propose that instead of deleting the stumble logs, we just provide a user defined directory where stumble files can be **moved** to instead of deleted.  That way we can replay the StumbleBundles through the simulation mode of #1191 satisfying the requirement of #979.

Exporting data becomes much simpler as we then only have to implement transform functions on the JSON data.

scruss:
Yes, please ... Don't gracenote us!

Ideally, I'd want website access to all data since it's open, but later ...

crankycoder:
You can currently download the aggregated cellular data from https://location.services.mozilla.com/downloads

We don't currently publish any wifi data yet as we haven't resolved the privacy issues surrounding that.  More information is available at that downloads link.

PR #1256, opened 2014-11-13
Enable stumble simulation

crankycoder:
This adds preliminary hooks to enable simulated stumbles.

Overview:

There's a new developer option checkbox to enable simulation mode.

When this is enabled, toggling the stumbler to 'on' will generate a context wrapper (SimulationContext) around the application context that is used by the GPSScanner, WifiScanner and CellScanner.

The SimulationContext provides hooks to emit CellInfo and ScanResult to spoof cell scans and wifi scans.  The GPS scans are spoofed using a MOCK_LOCATION permission and we inject new locations into the LocationManager.

Currently, the spoofed locations will start from your last known 'real' GPS location and will simply increment north west by roughly 5 m every second.  Cell tower and wifi spoofing is hardcoded to simply use a single tower and 2 wifi BSSIDs.

The next step to getting this into something more useful will be to land #1255 and then have the ability to read back the JSON logs and replay a full stumble.

I've renamed a few of the Context instances so that it's clear that we're actually passing the main application context. This is important as that context is wrapped by SimulationContext in the case of a simulation.

The AsyncUploader also checks the simulation preference to see if uploads should be real or just be mocked out with a spoofed HTTP 200 success.

Along the way, I've upgraded robolectric to use revision 2.4.


crankycoder:
This needs to properly mock out the visible GPS on the GUI and we need to catch an exception if the phone does not have the developer option of "Allow Mock Locations" is enabled.

android/src/main/java/org/mozilla/mozstumbler/client/mapview/GPSListener.java
```diff
@@ -69,8 +70,14 @@ void removeListener() {
...
     @Override
     public void onLocationChanged(Location location) {
+        if (location == null) {
+            return;
+        }
+
         if (mMapActivity != null) {
             mMapActivity.get().setUserPositionAt(location);
+            ClientPrefs.getInstance().setSimulationLat((float) location.getLatitude());
+            ClientPrefs.getInstance().setSimulationLon((float) location.getLongitude());
```
garvankeeley: Do we have to do this on every location? Isn't this just to seed it one time? Writing to the prefs heavily is something I would like to avoid. 
Is there a Build.Debug flag we can wrap this code in?

android/src/main/java/org/mozilla/mozstumbler/service/Prefs.java
```diff
@@ -26,6 +26,12 @@
     private static final String LAST_ATTEMPTED_UPLOAD_TIME = "last_attempted_upload_time";
     protected static final String PREFS_FILE = Prefs.class.getSimpleName();
 
+    // Simulation prefs
+    private static final String SIMULATE_STUMBLE = "simulate_stumble";
+    private static final String SIMULATION_LAT = "simulate_lat";
+    private static final String SIMULATION_LON = "simulate_lon";
+
+
```
garvankeeley: nit extra CR

android/src/main/java/org/mozilla/mozstumbler/service/Prefs.java
```diff
@@ -159,6 +165,18 @@ protected void setBoolPref(String key, Boolean state) {
         apply(editor);
     }
 
+    protected float getFloatPrefWithDefault(String key, float def) {
+        return getPrefs().getFloat(key, def);
+    }
+
+
```
garvankeeley: nit extra CR

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/ScanManager.java
```diff
@@ -93,24 +99,35 @@ public synchronized void setPassiveMode(boolean on) {
...
+        if (mAppContext == null) {
             Log.w(LOG_TAG, "No app context available.");
             return;
         }
-        if (mGPSScanner == null) {
-            mGPSScanner = new GPSScanner(context, this);
-            mWifiScanner = new WifiScanner(context);
-            mCellScanner = new CellScanner(context);
+
+        Prefs prefs = Prefs.getInstance();
```
garvankeeley: Would love this whole block wrapping in Build.Debug or something

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/ScanManager.java
```diff
@@ -93,24 +99,35 @@ public synchronized void setPassiveMode(boolean on) {
...
         }
 
-        if (AppGlobals.isDebug) {
-            Log.d(LOG_TAG, "Scanning started...");
+        mGPSScanner = new GPSScanner(mAppContext, this);
+        mWifiScanner = new WifiScanner(mAppContext);
+        mCellScanner = new CellScanner(mAppContext);
+
+        if (prefs != null) {
+            Log.d(LOG_TAG, "Simulation is set to: " + prefs.isSimulateStumble());
```
garvankeeley: this could move to after line 121?

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/ScanManager.java
```diff
@@ -128,6 +145,12 @@ public synchronized boolean stopScanning() {
             return false;
         }
 
+        if (mAppContext instanceof SimulationContext) {
```
garvankeeley, line 155: I like this check, if we can do more wraps like this, it separates the flow of test code from non-test code. And prevents any performance nits.

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiManagerProxy.java
```diff
@@ -0,0 +1,121 @@
...
+           return result;
+       } else {
+           WifiManager manager = getWifiManager();
+           if (manager == null) {
+               return null;
+           }
+           return getWifiManager().getScanResults();
+       }
+
+
```
garvankeeley: oh yes, my meaningful review comment: extra CRs

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiManagerProxy.java
```diff
@@ -0,0 +1,121 @@
...
+        public boolean isWifiEnabled() {
+       return getWifiManager().isWifiEnabled();
+   }
+
+    private WifiManager getWifiManager() {
+        return (WifiManager) mAppContext.getSystemService(Context.WIFI_SERVICE);
+    }
+
+
```
garvankeeley: extra CR. see aren't I helpful?

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/cellscanner/CellScanner.java
```diff
@@ -34,31 +38,26 @@
...
-        List<CellInfo> getCellInfo();
-    }
+    private final ISimpleCellScanner mSimpleCellScanner;
 
-    public CellScanner(Context context) {
-        mContext = context;
-        mCellScannerImplementation = new CellScannerImplementation(context);
+    public CellScanner(Context appCtx) {
+        mAppContext = appCtx;
+        if ((Prefs.getInstance() != null) && Prefs.getInstance().isSimulateStumble()) {
```
garvankeeley: we don't normal check if Prefs are null, it is an assumed precondition that the app created this on startup. You could do a Prefs.createGlobalInstance(appCtx) here if you are concerned. Also, this code could be wrapped in a build-time check for debugging build.

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/cellscanner/ISimpleCellScanner.java
```diff
@@ -0,0 +1,14 @@
+package org.mozilla.mozstumbler.service.stumblerthread.scanners.cellscanner;
+
+import java.util.List;
+
+/**
+ * Created by victorng on 14-11-10.
+ */
```
garvankeeley: Remove this, and add an MPL

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/cellscanner/MockSimpleCellScanner.java
```diff
@@ -0,0 +1,61 @@
...
+import org.mozilla.mozstumbler.service.core.logging.Log;
+import org.mozilla.mozstumbler.service.stumblerthread.scanners.SimulationContext;
+
+import java.util.LinkedList;
+import java.util.List;
+
+/**
+ * Created by victorng on 14-11-11.
+ */
+public class MockSimpleCellScanner implements ISimpleCellScanner {
```
garvankeeley, line 17: remove, add MPL on file

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/cellscanner/SimpleCellScannerImplementation.java
```diff
@@ -92,6 +92,21 @@ public synchronized void start() {
...
+            }
+            mPhoneType = mTelephonyManager.getPhoneType();
+            if (mPhoneType == TelephonyManager.PHONE_TYPE_NONE) {
+                // This is almost certainly a tablet or some other wifi only device.
+                return;
+            }
+
+            if (mPhoneType != TelephonyManager.PHONE_TYPE_GSM
+                && mPhoneType != TelephonyManager.PHONE_TYPE_CDMA) {
+                throw new UnsupportedOperationException("Unexpected Phone Type: " + mPhoneType);
```
garvankeeley: Were we always throwing in this condition?

android/src/main/java/org/mozilla/mozstumbler/service/uploadthread/AsyncUploader.java
```diff
@@ -124,8 +128,19 @@ private void uploadReports(AsyncUploadParam param) {
...
             headers.put(MLS.NICKNAME_HEADER, param.nickname);
 
+            Prefs prefs = Prefs.getInstance();
             while (batch != null) {
-               IResponse result = mls.submit(batch.data, headers, true);
+                IResponse result;
+                if (prefs != null && prefs.isSimulateStumble()) {
+
+                   result = new HTTPResponse(200,
+                                                       new HashMap<String, List<String>>(),
```
garvankeeley, line 137: fix indentation

garvankeeley:
Other than my nits, and a few questions, this is working great for me on my device

android/src/main/java/org/mozilla/mozstumbler/client/mapview/GPSListener.java
crankycoder: I'll change this to expose the lat/long in the developer options and have it set once.  If the default is overwritten once, I'll skip the preference write.

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/ScanManager.java
crankycoder: Ah, this was just stuffed in during debugging of the feature. I'm just removing this log line as it provides no value to us.

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/ScanManager.java
crankycoder, line 155: This check is mandatory actually - the simulation expects that the app context has been wrapped.  The context acts as a central source of all simulated data.  Checking the type was the only way I knew to safely get at that data.

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiManagerProxy.java
crankycoder: dang it.  I'm sure there's a way to tell Intellij to do this automatically when reformatting code.. but TOO MANY DAMN OPTIONS.

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/cellscanner/SimpleCellScannerImplementation.java
crankycoder: Yeah. I'm not sure why github is saying this bit is new.  The phone type comparison was done way way back in:

https://github.com/garvankeeley/MozStumbler/blob/723f7727507c872d4ff3459866dbc56de8547126/src/org/mozilla/mozstumbler/service/scanners/cellscanner/CellScannerNoWCDMA.java#L73

There's really only 4 phone types anyway, (NONE, GSM, CDMA and SIP).  I'll modify this so that we treat NONE and SIP the same, and drop the if block on 106-109.

crankycoder:
Crap. 3-way merge left to fix up, but I think I've addressed the comments you made.

crankycoder:
I want to double check the merge conflicts one more time to make sure I got the low power battery mode stuff merged properly.

garvankeeley:
+1 to the isDebug changes to ensure these code paths are debug build only. Yeah this looks ready to land.

crankycoder:
I've double checked the 3-way merge, this should be ok now.

crankycoder:
I'm finishing up adding a testcase to this.  Closing in the meantime.

PR #1259, opened 2014-11-13
Update translations

### LINKED ISSUE #1257: Minor settings UI items: Crash Reports title and change to MLS detailed info
- Title for crash reports section:
  *\* just "Crash reports", or something in the direction of "Feedback" / "Support". Firefox has "Data Choices".
- detailed info for MLS location:
  "At the cost of a bit more energy, memory and network usage, the MLS location is shown for each report. This is shown on the map as a red dot for the MLS location, with a line to its report. However, it will not query locations over mobile data connections if \"%1$s\" is enabled."

cascheberg:
This fixes #1257 and updates German translations

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/subactivities/PreferencesScreen.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/subactivities/PreferencesScreen.java
@@ -5,9 +5,7 @@
 package org.mozilla.mozstumbler.client.subactivities;
 
 import android.app.AlertDialog;
-import android.app.Dialog;
 import android.content.Context;
-import android.content.DialogInterface;
 import android.content.Intent;
 import android.os.Build;
 import android.os.Bundle;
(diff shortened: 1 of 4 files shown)


Djfe:
Ich fand Auflösung besser
Ansonsten: "Mehr Details" und "Weniger Details"

Es gibt keine hohen oder niedrigen Details, das klingt einfach falsch ;)

cascheberg:
OK, changed back to "Auflösung".
I wasn't sure about the wording either, but "Hohe Details" seems to be quite common when talking about computer game settings. I think "Mehr Details" is also not quite perfect.

Djfe:
wording is always complex, but everybody should understand "Auflösung"
"Mehr/Weniger Details" might not suggest that it is about the resolution
it could also be something like showing less street names and stuff like that (which of course is the case, but it's because of the lower zoom level that is being used to get the lower resolution)

garvankeeley:
Looks good

PR #1261, opened 2014-11-14
Low battery mode: make client-side configurable

garvankeeley:
Moved the low battery check in active scanning to the client side, and added a developer setting to set the min battery level.


cascheberg:
So I will close #1250 unmerged and file a new issue about user notification.

Some comments about this PR (vs. #1250):
- the user should be able to ignore low battery and restart scans, so the app should sent the turn off intent only once
- you could send a dedicated low battery intent instead of the `INTENT_TURN_OFF` to prepare for user notifications, I used the `notificationDrawerEventReceiver` for that
- in my version I hooked up to the `ACTION_BATTERY_CHANGED` intent instead of creating my own runnable
- I kept the `TurnOffReceiver` because I thought it would help with testing

garvankeeley:
Thanks for the review, good points made, I am updating my PR to address most of these,

garvankeeley:
@cascheberg Updated based on your feedback
@crankycoder this needs an automated test. But I don't think we have the framework in place to do the testing I want, yet.

Djfe:
Some of my ideas:
It should allow 0% battery level as the stopping point -> stop scanning when the device is shutting down, to save the scanned data before it gets lost

Is there a check at the first scan start? it shouldn't start scanning automatically on app start on low battery. Or is it doing that already because of a general check?

EDIT:
what do you think about using a picker for the battery percentage instead?
http://developer.android.com/design/building-blocks/pickers.html

garvankeeley:
I can add 0% .

The checks begin after scanning starts when the battery service reports a change.

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/ScanManager.java
```diff
@@ -32,28 +32,33 @@
...
+    private enum PassiveModeBatteryState { OK, LOW, IGNORE_BATTERY_STATE }
+    private PassiveModeBatteryState mPassiveModeBatteryState;
+
+    private BatteryCheckReceiver.BatteryCheckCallback mBatteryCheckCallback = new BatteryCheckReceiver.BatteryCheckCallback() {
+        @Override
+        public void batteryCheckCallback(BatteryCheckReceiver receiver) {
+            if (mPassiveModeBatteryState == PassiveModeBatteryState.IGNORE_BATTERY_STATE) {
+                return;
+            }
+            final int kMinBatteryPct = 15;
```
crankycoder, line 45: Shouldn't this use the battery preference settings instead of just 15?

garvankeeley, line 45: This is hard coded for out-of-the-box Fennec use.

cascheberg:
There have been lots of fixes/changes since the last release, so it would be nice to have another official pre-release. Do you have any plans regarding that?

garvankeeley:
Yeah, hoping to do a github release tomorrow. Then maybe a week later do an f-droid and playstore release if that build is looking healthy.

PR #1265, opened 2014-11-15
Only auto-scroll log if at bottom

### LINKED ISSUE #1263: Freeze log
I would like to have a button in the log screen that allows me to freeze the log
->New entries like a GPS fix normally "refresh" the log, so that the view gets reset to the bottom of the screen, which is bugging me for some time already

cascheberg:
Only auto-scroll log if it is already scrolled to the bottom. This fixes #1263


garvankeeley:
Much better, thanks!

crankycoder:
Hi @cascheberg !  Thanks for this.  We're going to start getting a bit more strict about having tests added for features like this.  I've got a branch setup over here https://github.com/crankycoder/MozStumbler/tree/features/1263-autoscroll-log that extends robolectric a bit so that we can see the enableScroll behavior actually work properly in a unittest.

Issue #476, opened 2014-03-08
Dynamic statistics in notification?


goldfndr:
Currently, the notification shows only the `service_scanning` text (Scanning for Wi-Fi and Cellular networks). It'd be nice to have this dynamically updated like the WiGLE stumbler, as I don't always have it in the foreground. For example,

453 locations, 1/12 Cells, 4/23 Wi-Fi

with a prefix of "Off; " if geofenced.

Djfe:
A button for start/stop scanning would be nice as well

cascheberg:
@goldfndr This feature is now implemented, so I think this issue should be closed.

PR #1270, opened 2014-11-18
Battery-saving mode (by pausing GPS and scanners) on unchanging location.

garvankeeley:
Uses motion detection to wake the scanning up after being paused. Note, on Android 4.3 and higher, the significant motion detector is used, which does a great job of detection user movement. On lower Androids, the acceleration sensor is used (with a low threshold), which will wake on small movements of the device.

Uses 2-part detection:
1) DetectUnchangingLocation is used to determine if the user has not
moved in 2 minutes, then the ScanManager pauses (this is a low power
state)
2) When the MotionSensor detects movement (which uses very low power
sensors to do so), start scanning again.

Also worth noting, the notification text changes when paused, and the
map view shows a paused message also.

@crankycoder this branched was based of my previous battery change PR, so I'll clean that one up, land it first, and merge this one. However changes I make there won't affect this in any major way, so if you feel like starting on a review, go nuts.


android/src/main/res/layout/fragment_developer_options.xml
```diff
@@ -17,5 +17,135 @@
...
+                android:layout_alignParentLeft="true"
+                android:layout_alignParentStart="true" />
+
+            <Spinner
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:layout_toEndOf="@+id/batteryLabel"
+                android:layout_alignBaseline="@+id/batteryLabel"
+                android:layout_toRightOf="@+id/batteryLabel"
+                android:entries="@array/battery_percent_options"
```
crankycoder: entries seems wrong here.  I think this needs to be in seconds not percentages.

android/src/main/res/layout/fragment_developer_options.xml
```diff
@@ -17,5 +17,135 @@
...
```
crankycoder: entries seems wrong here. I think this needs to be in meters not percentages.

android/src/main/res/layout/fragment_developer_options.xml
garvankeeley: Sorry, this is other stuff that is mashed into this PR, I need to clean it up. I just wanted to PR something before I did the cleanup, to show as early as possible what I am working on.

android/src/main/res/layout/fragment_developer_options.xml
garvankeeley: Ideas I am playing with, will be removed from this PR

garvankeeley:
I will fix this PR after #1261 lands.

garvankeeley:
Should really add developer settings for distance required to move for disabling idle and for the time window for this movement to occur.

cascheberg:
I'd like to fix #476 because it is quite nice to have on Android 5 lock screen, and the `NotificationUtil` in this PR seems quite helpful for that. Are you going to merge this soon, or can you merge that part in advance?

garvankeeley:
This is ready to land IMHO.
Added developer option to control the motion detection params.
The default is 50 meters in 120 seconds, this screenshot I have changed the numbers for debugging
screenshot_2014-11-20-19-06-09

garvankeeley:
@crankycoder is it ok we we only land this  and #1279 tomorrow morning, push a release to github, then land the other stuff. Otherwise, this patch is conflicting with the pending PRs, which will delay getting a release build up.

crankycoder:
Yeah, let's get the battery stuff in quickly. I'll check this in the
morning and we'll move to cut a release.

garvankeeley:
@cascheberg Updating the notification item with some stats would be great. 
Keep in mind this code recreates the notification every time the message changes, which has the effect in the UI of the row deleting (with animation) and creating a new row, so the update should be rate limited. Maybe there is a better way to change the message on the notification item that doesn't require deletion and re-creation?

Issue #1264, opened 2014-11-14
The osm contributors string needs a translation, too


Djfe:
This is about the osm string in the bottom-right corner of the map

cascheberg:
@Djfe This is really easy to fix. So if you are interested in fixing this yourself and need some help, I could give you some hints.

Djfe:
ok the pull request is up at #1273 ^^

PR #1273, opened 2014-11-18
Added OSM copyright string #1264

Djfe:
Moved strings from code to the English and German strings.xml files.
Added a comment to the layout file of the map about where to find the
code that sets the strings.

Was quite a hassle to find that string...
Should have used the windows search from the beginning instead of looking through each .xml file ^^
(ended up being hidden in the mapfragment)

Pls the check the code, since I never wrote apps myself for Android it is likely that I made a mistake.
(I'm not sure whether I can set the @string stuff in the mapfragment and whether it is good style and I'm not sure whether I can use a "-" in the string's name (for F-Droid))

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java
@@ -524,9 +524,9 @@ public final class MapFragment extends android.support.v4.app.Fragment
     private void showCopyright() {
         TextView copyrightArea = (TextView) mRootView.findViewById(R.id.copyright_area);
         if (BuildConfig.TILE_SERVER_URL == null) {
-            copyrightArea.setText("Tiles Courtesy of MapQuest\n© OpenStreetMap contributors");
+            copyrightArea.setText(getActivity().getString(R.string.map_copyright_fdroid));
         } else {
-            copyrightArea.setText("© MapBox © OpenStreetMap contributors");
+            copyrightArea.setText(getActivity().getString(R.string.map_copyright_moz));
         }
     }
 
--- a/android/src/main/res/layout/activity_map.xml
+++ b/android/src/main/res/layout/activity_map.xml
@@ -169,6 +169,7 @@
         android:layout_alignParentRight="true"
         android:layout_alignParentEnd="true" />
 
+	<!-- The copyright text is set at java\org\mozilla\mozstumbler\client\mapview\MapFragment.java showCopyright() -->
     <TextView
         android:id="@+id/copyright_area"
         android:textColor="@android:color/black"
--- a/android/src/main/res/values-de/strings.xml
+++ b/android/src/main/res/values-de/strings.xml
@@ -86,6 +86,8 @@
     <string name="first_run_welcome_to_mozstumbler">Willkommen bei Mozilla Stumbler</string>
     <string name="map_unavailable">Die Karte kann nicht angezeigt werden, da kein SD-Kartenspeicher vorhanden ist.\nSie können aber immer noch Daten sammeln.</string>
     <string name="map_offline_mode">Die Karte kann ohne eine verfügbare Internetverbindung nicht aktualisiert werden.\nSie können aber immer noch Daten sammeln.</string>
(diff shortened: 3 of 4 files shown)


cascheberg:
OK, first of all, you should **really** install and setup [Android Studio](https://developer.android.com/sdk/installing/studio.html). Try it and you will know why.
Examples: you found the relevant code parts, but the IDE can easily tell you were to find the string, and it would also tell you that your PR is not valid code.

Once you installed Android Studio:

> File > Import Project... > [select cloned git folder]

_One more thing: you should not close the issue before the corresponding PR gets merged._

Djfe:
Well now I have this:

```
private void showCopyright() {
        TextView copyrightArea = (TextView) mRootView.findViewById(R.id.copyright_area);
        if (BuildConfig.TILE_SERVER_URL == null) {
            copyrightArea.setText(context.getResources().getString(R.string.map_copyright_fdroid));
        } else {
            copyrightArea.setText(context.getResources().getString(R.string.map_copyright_moz));
        }
    }
```

I don't know how to get a context. (where to get it from)

garvankeeley:
I think it is ok to do translations without the IDE and let travis catch the failures, and Github interface is pretty handy for making simple edits.

Djfe:
I think he tries to get me into contributing further stuff to this projects (apart from translations and comments on features/bugs).
So nice way of decoying me in :)

Another reason is that the IDE would have pointed me to the string -> I could have found it easier.
And the IDE points out code errors fairly well so that you can correct them. (Notepad++ doesn't of course ;) )

Of course the GitHub webinterface works perfectly fine for small edits.
For this pull request I used the Windows Client and Notepad++ though.
This makes much more sense since you cannot make one commit with several file changes in the webinterface. It ends up being one commit per file and you cannot rebase them online.

Have tried out Android Studio now.
Gonna read about contexts tomorrow so that I can update this pull request ^^

But I want to learn writing Android apps with a book or if I find someone that wants to teach me then with him/her.

cascheberg:
I did not want to force you or anything ;)
I myself never developed on Android before this, and my experience was that installing the IDE helped me to understand the structure of the project. Another feature is that the IDE gives you a preview of the layouts, and from there, right click > "Find Usages" can take you to relevant code parts (for example to find where the strings are set).

@Djfe In the same file you can find this example: `mTextViewIsLowResMap.setText(getActivity().getString(R.string.low_resolution_map));`. You can do the same. But in fact, `setText()` also accepts a resource ID such as `R.string.low_resolution_map` as a parameter directly (as the IDE can tell you).

@garvankeeley Are there any strict rules about the exact wording of those strings? If it's not enough to translate those to the best of one's knowledge, then parts of it could stay hard-coded outside the string resources.

garvankeeley:
@cascheberg I don't understand your question, rules about wordings of strings?

cascheberg:
I was wondering if copyright strings are predefined by a third party (OSM / MapQuest) and thus must not be changed.

garvankeeley:
Oh ok, I have no idea. We can just follow the convention that other things use, can anyone comment on whether they typically see copyright notices translated? It would make sense that they are.

Djfe:
I updated my pull request now :)
Does this look better?

I just imported the cloned GitHub folder, installed the necessary sdks and gradle and maven got installed automatically
still it reports sync errors with gradle and in all every .java file that I open it reports stuff that is not defined like R. or buildconfig

what do I do wrong?

...and the build failed again...

cascheberg:
Travis tells you that there is one closing parenthesis missing. And the indentation in the .xml files appears to be wrong.

Can you post those sync errors?

What you need to have installed is this (can be done from within Android Studio):
- **Tools**
  - Android SDK Tools
  - Android SDK Platform-tools
  - Android SDK Build-tools 19.1
- **Android 4.4.2 (API 19)**
  - SDK Platform 19
- **Extras**
  - Android Support Repository
  - Android Support Library

Djfe:
android studio

cascheberg:
Go to

> Tools > Android > SDK Manager

and install what I wrote in the previous comment.

Djfe:
I missed the Android Repo support, it works now :)

Djfe:
Ok I found the issue I guess, it was a missing bracket ^^
it builds now

tested it running a debug build in an emulator, works fine

PR #1274, opened 2014-11-19
Large Text is on: show only GPS fixes label, hide visible

cascheberg:
@garvankeeley There is a layout problem regarding the display of available GPS satellites if `Accessibility > Large text` is enabled.
This PR fully fixes the problem for #used satellites <= 9. When more satellites are in use, the number of available satellites is now cut off. I could not find another fix without breaking what was discussed in #1051.
Do you think this is an improvement? Any other ideas?


garvankeeley:
Is there a way to know that Large Text is on, or the text is overflowing? If so, I would not show the available satellites, just the fixes. Or do a font size reduction on both the visible and fixed satellites in order to fit the available space.

cascheberg:
PR updated to always hide available satellites if large text accessibility is enabled.

PR #1286, opened 2014-11-21
Fix for MotoG (XT1032) accelerometer usage

crankycoder:
The Moto G is a 2013 device and runs KitKat, but will only respond to API 3 of the accelerometer sensor API.


Djfe:
what about the 1.4.0 commit? I think it was included accidently

PR #1287, opened 2014-11-22
Add metrics to notification

cascheberg:
This adresses #476. `Notification.BigTextStyle` was only added in API level 16. I think the notification text can only be one line though, which is a bit too short.

@garvankeeley 

> Keep in mind this code recreates the notification every time the message changes, which has the effect in the UI of the row deleting (with animation) and creating a new row, so the update should be rate limited.

There is no animation during an update on Android 5. I did not investiage yet how other devices (>= API 16) do this though.

Pulled down notification on Android 5 lock screen:
notification

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -187,7 +187,7 @@ dependencies {
     // libraries, because they support a wide range of Android versions and
     // provide APIs for recommended user interface patterns.
 
-    compile "com.android.support:support-v4:19.1.+"
+    compile "com.android.support:support-v4:21.0.+"
     compile 'com.android.support:appcompat-v7:19.1.0'
 
     compile('org.simpleframework:simple-xml:2.7.1') {
--- a/android/src/main/java/org/mozilla/mozstumbler/client/MainApp.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/MainApp.java
@@ -219,8 +219,7 @@ public class MainApp extends Application
             return;
         }
         NotificationUtil nm = new NotificationUtil(this.getApplicationContext());
-        Notification notification = nm.buildNotification(getText(R.string.service_name).toString(),
-                getText(R.string.service_scanning).toString(), getString(R.string.stop_scanning).toString());
+        Notification notification = nm.buildNotification(getString(R.string.service_name), getString(R.string.stop_scanning));
         mStumblerService.startForeground(NotificationUtil.NOTIFICATION_ID, notification);
         mStumblerService.startScanning();
         if (mMainActivity.get() != null) {
@@ -415,11 +414,7 @@ public class MainApp extends Application
         AppGlobals.guiLogInfo("Is motionless: " + mIsScanningPausedDueToNoMotion);
 
         NotificationUtil util = new NotificationUtil(this.getApplicationContext());
-        if (mIsScanningPausedDueToNoMotion) {
-            util.updateSubtitle(getString(R.string.map_scanning_paused_no_motion));
-        } else {
-            util.updateSubtitle(getString(R.string.service_scanning));
-        }
+        util.setPaused(mIsScanningPausedDueToNoMotion);
 
         if (mMainActivity.get() != null) {
             mMainActivity.get().isPausedDueToNoMotion(mIsScanningPausedDueToNoMotion);
--- a/android/src/main/java/org/mozilla/mozstumbler/client/navdrawer/MainDrawerActivity.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/navdrawer/MainDrawerActivity.java
@@ -264,7 +264,8 @@ public class MainDrawerActivity
 
         int observationCount = service.getObservationCount();
         mMapFragment.formatTextView(R.id.text_observation_count, "%d", observationCount);
-        mMetricsView.setObservationCount(observationCount, service.getUniqueCellCount(), service.getUniqueAPCount());
+        mMetricsView.setObservationCount(observationCount, service.getUniqueCellCount(),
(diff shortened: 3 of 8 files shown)


Djfe:
looks pretty cool

cascheberg:
I tested this on a virtual device running API level 16, notification updates did not cause animations there.

garvankeeley:
I updated my phone to L, so I can't check this too easily. But yes, you can't change the notification message without destroying it and recreating it. On 4.4 for me, this would cause the notification to be removed from the list using animation, then added like a new notification again with animation. Doing this while you are looking at the notification list makes it rather difficult to read (or interact with the notification item)

garvankeeley:
@cascheberg can I sell you on something more compact for this? 
Like:
Mozilla Stumbler
Scanning. Reports: xxx, Wi-Fis: xxx, Cells: xxx

The swipe down could still have the last uploaded info and the X to stop scanning.

EDIT: trying now on a 4.2  device, I see you did something similar. Cool.
Unfortunately, the deletion and recreation makes this method unusable, if you expand it, on update it gets collapsed.
Perhaps updating once a minute might be ok.

Djfe:
isn't there a way to prevent the animation?

does that much text even fit into one notification? (is there a maximum string length for the compact notification?)

cascheberg:
This updated version does not use `setSubText` anymore, which is only supported on API > 16. Instead it includes the metrics in the notification text in a very short form for all devices (as you suggested):

> Scanning. (Metrics: 1, 6, 23)

There is not enough space to add the labels there, but the order (reports, cells, wifi) is the same as in the `MetricsView`, and usually wifis should be a distincly greater number than cells.

It also restricts the update frequency to once a minute for all versions before Android 5.

cascheberg:
@garvankeeley Do you have any new feedback or plans regarding this PR?

garvankeeley:
Sorry, I missed that this PR was updated, will look at it now.

garvankeeley:
This seems to fit ok, what do you think?
metrics_notification_text %1$s Reports:%2$d Cells:%3$d Wi-Fis:%4$d 
screenshot 2014-12-09 09 32 24

garvankeeley:
I find this takes up most of my notification screen:

screenshot_2014-12-09-09-21-51

garvankeeley:
How about this:

screenshot_2014-12-09-09-38-48

crankycoder:
Can we get icons in here like the MapView?  I'd prefer to have one visual language that is consistent between different UI bits of the stumbler.  

I have a feeling text layouts will breakdown with translations.  Things won't line up.  I think it will be better to have one language across the mapview, the notification and the widget in #1316

cascheberg:
The notification should only be that big after you expand it manually. And I think it will become small again automatically, or if there are several notifications. (I can't test this right now, and I think that might depend on the Android version.)
Regarding the short form: did you try 4 to 5 digits each on a small screen? And localization might be difficult (e.g. the correct German word for "Cells" is "Funkzellen").

garvankeeley:
The translations can have newlines, we don't need to have added space for the english one. Or use some other compact form.

cascheberg:
You can't have newlines in the basic notification content text. And I don't think you can have icons within the notification text either.

garvankeeley:
I think the plan is to move away from icons, we have a bug for that. Also, I think icons would create too much visual clutter.

cascheberg:
If this needs bigger modifications to be acceptable: it is possible to create custom notifications using `RemoteViews`.

garvankeeley:
Just updating the strings is the final piece of the puzzle.
I'll try to summarize where we are:

Collapsed Notification:
1) Show: Metrics (x, y, z)
- I don't think this is clear enough for average users

2) Show: Reports:x Cells:y Wi-Fis:z
- Translation problem
- May be too long in english with large numbers

Expanded Notification:
1) https://github.com/mozilla/MozStumbler/pull/1287#issuecomment-66292190
- Can this be more compactly written? Notifications area is a shared space, even though it is expanded, we should only use the space we have to.

garvankeeley:
A minor item, when I apply this patch I get:

Error: uses-sdk:minSdkVersion 10 cannot be smaller than version L declared in library com.android.support:support-v4:21.0.0-rc1

garvankeeley:
Collapsed Notification ideas:
1) Show only Scanning. Reports: x

or

2) Change the title to Mozilla Stumbler (paused) if paused, and have the detail label as
Reports:x Cells:y Wi-Fis:z

cascheberg:
One more idea: use `setSubText` again as in the first commit (API >= 16 only, adds a second line, makes text smaller)

I did not see the SDK problem last time. It would be great to have the Gradle version updated to 1.0 for easier testing ;) (I created a PR for that.)

garvankeeley:
I just noticed that notification titles are an action sentence, not just a app title.
If the title was
- "Mozilla Stumbler is scanning"

that would actually be more correct, and provide more space on the 2nd line.

garvankeeley:
Here is a patch that can be applied on top of this one to get more compact notifications:
https://gist.github.com/garvankeeley/128c88e7bb2d93f31c7b

@cascheberg can I get your feedback on that?

I still don't know how to handle the build.gradle issue though.

crankycoder:
@garvankeeley I don't get the SDK problem, but I'm running over a dev branch that has the newer gradle versions installed.

This patch looks like what we discusses earlier today - lgtm.

cascheberg:
From looking at the code I think that the idea is OK.

But I cannot test any of this until next week, and I might be unavailable until then.

garvankeeley:
Been running the app with this notification on, it is great, thanks @cascheberg!
Closing this, will land as PR #1341.

Issue #1295, opened 2014-11-24
Port significant motion sensor code from open-sensor project


crankycoder:
There's some code at the open sensor projects licensed under the Apache license that should allow us to implement proper signficant motion detection using the standard accelerometers for devices that do not support the significant motion detector.

https://github.com/sensorplatforms/open-sensor-platform/tree/master/embedded/common/alg

PR #1298, opened 2014-11-24
Features/1297 filter gravity

### LINKED ISSUE #1297: TYPE_ACCELEROMETER readings include gravity
We have to filter out gravity from the LegacyMotionSensor readings when we use Sensor.TYPE_ACCELEROMETER.

The TYPE_LINEAR_ACCELEROMETER already does this automatically. 

One thing to note is that filtering gravity out will require convergence on the calculation, so we'll need to skip the first couple dozen iterations of the value before we get something that is usable.

crankycoder:
This fixes #1297 by adding a filter to extract gravity from the regular accelerometer readings.

This patch also includes a test that checks for convergence using raw values that I extracted from my device on startup.

The patch includes a testcase using robolectric.  The most important detail in the test is that robolectric isn't magic.  You still can't construct a SensorEvent because of private constructors and final instance attributes (thanks Android!), so I ended up passing a flattened SensorEvent back from the event listener to the LegacyMotionSensor instance.

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/LegacyMotionSensor.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/LegacyMotionSensor.java
@@ -1,0 +1,122 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+package org.mozilla.mozstumbler.service.stumblerthread.motiondetection;
+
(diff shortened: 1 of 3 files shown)


Djfe:
Will this make the legacy motion detector usable? 
I tested it today and it seems to even report motion if the device is not moved...

Further debug log Infos could be helpful to determine the reason

Maybe MozStumbler could compare the GPS movement speed with the legacy sensor so that we could see in the log whether the sensor is just too sensitive.

crankycoder:
@Djfe sort of.

This will make the LegacyMotionSensor work as expected - which is that it will trigger after 1m/s^2 in any direction is detected.

This won't give you significant motion detection, that won't land until #1295 is ported, implemented and tested.

PR #1302, opened 2014-11-24
Update metrics on Reporter.ACTION_NEW_BUNDLE, not when cell/wifi counts change

cascheberg:
I noticed that UI updates always update all metrics. The UI is updated whenever a `ACTION_WIFIS_SCANNED` or `ACTION_CELLS_SCANNED` intent occurs, which is about once per second.
The current call stack looks like this:
- update the stats displayed in the map in `MainDrawerActivity.updateNumberDisplay()`
- always call `MetricsView.update()` _(this was improved)_
- call `MetricsView.updateQueuedStats()`
- call `DataStorageManager.getQueuedCounts()`

The last method for example calls `Zipper.zipData()` and has this comment:

> Some data is calculated on-demand, don't abuse this function

This PR changes the code to only update metrics stats if a bundle was flushed.
The UI updates could be differentiated a bit more even.

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/IMainActivity.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/IMainActivity.java
@@ -6,7 +6,7 @@ package org.mozilla.mozstumbler.client;
 
 public interface IMainActivity {
     // Thread-safe way to update the main UI.
-    public void updateUiOnMainThread();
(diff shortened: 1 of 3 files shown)


garvankeeley:
@crankycoder I'll review this one today

garvankeeley:
Agreed, this is much more efficient now, thanks!

PR #1306, opened 2014-11-25
Update the ACRA login

### LINKED ISSUE #1249: Rev the ACRA login
We're getting crash reports from 0.40.4 that were fixed a long ago.  Update the ACRA login to bounce the 0.40.4 crash logs.

crankycoder:
Updates the ACRA login to close #1249 so that we can shut down the errors from the old 0.40.4 release.


garvankeeley:
This is great, but when is a good time to merge this?

crankycoder:
You can merge it now. Both passwords are live.  I figure we can shut down the old login when the playstore release comes out.

PR #1307, opened 2014-11-25
Add dialog for testing motion detection.

garvankeeley:
This is accessible through long press on the developer screen title.

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/DetectUnchangingLocation.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/DetectUnchangingLocation.java
@@ -12,6 +12,17 @@ import org.mozilla.mozstumbler.service.AppGlobals;
 import org.mozilla.mozstumbler.service.Prefs;
 import org.mozilla.mozstumbler.service.stumblerthread.scanners.GPSScanner;
 
+// This class is a bit confusing because of 2 checks that need to take place.
+// 1) One check happens when a gps event arrives, to see if the user moved x meters in t seconds.
+// 2) The other is a timeout in case no gps event arrives during time t.
+//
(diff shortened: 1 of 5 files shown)


garvankeeley:
screenshot_2014-11-25-15-22-17

garvankeeley:
@crankycoder ready for your kind perusal.

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/DetectUnchangingLocation.java
```diff
@@ -111,11 +134,11 @@ private void scheduleTimeoutCheck() {
         mHandler.postDelayed(mCheckTimeout, mPrefMotionChangeTimeWindowMs + addedDelay);
```
crankycoder, line 134: Just a question so that I don't forget to ask tomorrow. It's not clear to me _why_ we have this addedDelay.  The raw mPrefMotionChangeTimeWindowMs seems like it should be good enough.  There doesn't seem to be any case where we use the raw value that the user has set without adding this additional 2 second padding.

We'll sort this out in IRC tomorrow - I'll codify the reasoning in a test.

garvankeeley, line 134: I'd like to clean this up after we have testing in. I don't like it.

Given a timeout time of _t_ ms, if I postDelayed for _t_ ms, the timeout check happens before _t_ ms (because postDelayed is an approximate delay). The timeout check has a confirmation that _t_ ms has passed (I can't recall why I confirm this, need to ponder more), if not, check again in _t_ ms. If there is no addedDelay, the timeout check has to run twice instead of once to check. This was annoying for debugging, the only negative effect is that the timeout check has to run a second time to see that _t_ has passed.

PR #1308, opened 2014-11-25
Update translations

cascheberg:
Make strings localizable
Update German translations


Djfe:
so you turn on prettytime again since they implemented my pull request? ("Gerade eben" instead of "Vor einem Augenblick")

for the german translations: except the stuff I commented on, r+

cascheberg:
@Djfe Review comments fixed. And no, it's not PrettyTime, it's a custom implementation, see #1199

garvankeeley:
LGTM, thanks!
I have avoided making work for localizers with the strings on the developer screen, as those strings I expect to change/disappear. 

The plan being to introduce new options/experiments on the developer screen, and if they are good enough, move them to settings, or even the metrics screen now that we have some space. Although, I don't see why that would necessarily require the strings to change, so maybe I have no point to make here :).

Issue #1311, opened 2014-11-27
ACRA crash: MapLocationListener.java:124 Network provider not available


garvankeeley:
Acra report:
Java.lang.IllegalArgumentException: provider doesn't exist: network at org.mozilla.mozstumbler.client.mapview.MapLocationListener.enableLocationListener(MapLocationListener.java:124)

cascheberg:
I got this error (reproducibly) when I ran the app in an Android virtual device API level 16. But I don't know if that was logged to the server, and of course I don't know if you got more reports from other people as well.

garvankeeley:
this should be fixed with #1312

PR #1312, opened 2014-11-28
MapLocationListener: wrap requestLocUpdates to catch and ignore exceptions

garvankeeley:
Arguably, one can check if particular location providers are available (instead of catching the exception as I did here). 
This is only useful if the code then takes further useful actions based on this info. We have no interest in handling this case beyond not crashing and putting an error in the log. Why is the network provider unavailable anyway? What is the mojo on the device required to reproduce this? I tried airplane mode, restricting geolocation to GPS only.  Neither triggered this.
Issue #1311


crankycoder:
This looks like something to do with the CM Privacy Guard.

Issue #983, opened 2014-10-10
widget support


Djfe:
widgets should implemented, I think
there are probably lots of people that use them still.

we could show stats

this is similar to #476 but would work on older android versions as well ;)

crankycoder:
Once #1498 lands, this should make homescreen widgets a lot easier to implement.

Widget docs: http://developer.android.com/guide/topics/appwidgets/index.html

PR #1316, opened 2014-11-30
Homescreen widget support

lomash:
This addresses #983 and to some extent #476. 

This commit would allow the addition of a simple widget to the user's home screen that currently displays the number of WiFi access points and cell towers it sees. These text views can be easily modified to display the user's rank on the leader board or the number of reports pending. 

Clicking on the icon toggles the service on and off.

When the service is on

When the service is off

Has been tested on Android 5.0 and Android 2.3.6


crankycoder:
Thanks @lomash !  Do you think you can recreate this patch against the current dev branch?

lomash:
I'm sorry but I'm new to this. What do you mean by that?

crankycoder:
We generally like to see fewer commits rather than more.  We don't really have too many hard rules, but the multiple merges in the commit history make this patch difficult to understand.

I've cleaned up the branch to show you something closer to what we want over here:

https://github.com/crankycoder/MozStumbler/tree/features/1316-recreated

Since your code compiled clean in Travis, I really just wanted 
To clean up that tree, I switched to your branch, then did : 
`git diff dev --binary > /tmp/widget.patch`

That got me a clean diff that I can apply to the dev tree with `git apply /tmp/widget.patch`.

After that I just cleaned up the patch a bit to remove changes that were just whitespace and some translations which look like they got mangled during your merges:

https://github.com/crankycoder/MozStumbler/commit/6471d90cac07814fa795d9f795e8f391f9ebe0a6

If you can - please drop by the irc channel #geo on irc.mozilla.org.  Instructions: https://wiki.mozilla.org/IRC

Now, if you go grab the branch I cleaned up, there's a couple problems I see.

The widget doesn't seem to enforce a minimum size so I get clipping of the widget pretty easily.

2014-12-08_21_31_50_png

So I think we need to do two things here. 

1) Move the numbers so that they appear horizontally.  I'd shrink the logos down for wifi and cell tower.
2) Enforce a minimum and maximum widget height and width so that the widget always looks 'good'.

This is a fantastic start though and a nice alternative to having the full map view for stumbling.

crankycoder:
I think we should probably just turn the icons red like the main map view when the antennas are disabled.  That avoids the problem of translating 'disabled' into N different languages.

lomash:
So, I redid the widget layout. It is now horizontal, the disabled text was removed. Here's how it looks right now.

stumbler_on

stumbler_off

Also, I set minimum heights for the layouts. I was unable to recreate the clipping on my devices. Can you check if this works for you?

https://github.com/lomash/MozStumbler/tree/1316_reworked

crankycoder:
This seems to work for me using the default launcher in Cyanogenmod 11 (Trebuchet) and the Google Now.

On Nova Launcher, I can force resize the widget which seems to break things but I think that's just Nova Launcher being crazy.  

In anycase, I think this is good enough for now.  We can modify the layout and unify the UI look and feel with the notification system later.

Can you merge dev into your branch - the only difference appears to be the AndroidManifest.xml where it's been simplified a bit so that we don't need to do substitution of the application name or version anymore.

lomash:
I'm getting a lot more conflicts than that. Even some with the Polish and Spanish strings.

https://github.com/mozilla/MozStumbler/compare/lomash:1316_reworked...mozilla:dev

crankycoder:
It looks like you're working on the same branch as this current pull request.  I've created a separate branch over at https://github.com/crankycoder/MozStumbler/tree/features/1316-recreated

Do you know how to pull from other branches using git?  If you're not comfortable on the command line, you can try using SourceTree from : http://www.sourcetreeapp.com/

lomash:
I merged it with the current version of dev. 

https://github.com/lomash/mozstumbler/tree/1316_reworked

crankycoder:
Great!  Can you close this pull request and open a new one with your new branch?  I'll check the diff again to see if there's anything outstanding that should be fixed.  @garvankeeley mentioned he'd like to see the number of reports in the widget - can we add that in there?

lomash:
made a new request #1353. I'll try to add that asap but it might still take a few days as I'm busy with my finals atm.

crankycoder:
Hey @lomash - #1498 should land shortly which ought to make widgets easier to implement without causing performance degradation.

I've updated #983 to point to more recent documentation for the developer docs on doing app widgets.

PR #1319, opened 2014-12-02
Fix MapLocationListener pausing GPS

### LINKED ISSUE #1317: 1.5.2 drains the battery by leaving the GPS active after stop stumbling
After upgrading MozStumbler into 1.5.2 in my Nexus One, and after the following steps:
1. Stumble a while.
2. Got home, submit report, press the stop button
3. Leave the application

I was using an external battery, so when I stop stumbling the phone was fully charged.  Next day morning, the battery was ~25% being just inactive.  The power consumption used to be ~11% overnight, not ~%75%.

Then I noticed the GPS was still active, by looking the icon in the notification bar.  After that, I rebooted my phone, I tried running MozStumbler again, stopped it and wait if the GPS became inactive, nothing happened.

In previous versions, every time I pressed the stop button, the GPS icon became inactive after a couple of seconds.

Now, the only way to get the GPS inactive after pressing the stop button is by force closing the application (via "_Manage apps_").

I also see this behavior on my Nexus 5, but because this has better battery, it did not become apparent earlier.

According to the Changelog, from 1.5.0 it should use less battery consumption if one does not move, then MozStumble stop stumbling.  But it turns out that it consume more battery if I don't kill the app explicitly after stop using it (after the screen, the GPS consumes more battery, then the wifi).

It seems to me that a fair compromise would be:
1. If the user does not press the stop button, then try to minimize the stumbling by checking any movement.
2. If the user explicitly press the stop button, then stop using the GPS.

cascheberg:
@garvankeeley This might fix #1317

How to reproduce the bug:
- start using the app
- at any point: do not move so that the scanning is paused
- **make sure that the app is not visible when you start moving and scanning is resumed**
  - e.g. enabled lockscreen (I suppose this happens a lot), use another app, press home button
- after scanning was resumed: stop scanning, press home button

Result:
- GPS is still active

I think what happens here is this:
- hiding app: `MapFragment.onPause` calls `mMapLocationListener.removeListener`
- moving after pause: `MapFragment.showPausedDueToNoMotionMessage(false)` re-enables GPS
- app becomes visible: `MapFragment.onResume` creates a new `MapLocationListener`, the reference to the old `MapLocationListener` with the active GPS is lost

This PR makes sure that `pauseGpsUpdates` does not have any effect after `removeListener` was called.

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapLocationListener.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapLocationListener.java
@@ -142,6 +142,7 @@ class MapLocationListener  {
         mLocationManager.removeGpsStatusListener(mSatelliteListener);
         enableLocationListener(false, mGpsLocationListener);
         enableLocationListener(false, mNetworkLocationListener);
+        mLocationManager = null;
     }
 
     public void pauseGpsUpdates(boolean isGpsOff) {
--- a/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapLocationListener.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapLocationListener.java
@@ -142,7 +142,6 @@ class MapLocationListener  {
         mLocationManager.removeGpsStatusListener(mSatelliteListener);
         enableLocationListener(false, mGpsLocationListener);
         enableLocationListener(false, mNetworkLocationListener);
-        mLocationManager = null;
     }
 
     public void pauseGpsUpdates(boolean isGpsOff) {
--- a/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java
@@ -644,7 +644,10 @@ public final class MapFragment extends android.support.v4.app.Fragment
         Log.d(LOG_TAG, "onPause");
         saveStateToPrefs();
 
-        mMapLocationListener.removeListener();
+        if (mMapLocationListener != null) {
+            mMapLocationListener.removeListener();
+            mMapLocationListener = null;
+        }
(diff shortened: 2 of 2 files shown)


cascheberg:
@garvankeeley I removed one unrelated code change, it's now in #1321. (Previous commit was 8f2aed1.)

The `mLocationManager = null` in `MapLocationListener` alone fixes the bug, and I thought it'd be good to have in case the method will be used somewhere else.
My idea of adding the `mMapLocationListener = null` in `MapFragment` was to make it more obvious that the object is in a kind of unusable state.

android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapLocationListener.java
```diff
@@ -142,6 +142,7 @@ public void removeListener() {
         mLocationManager.removeGpsStatusListener(mSatelliteListener);
         enableLocationListener(false, mGpsLocationListener);
         enableLocationListener(false, mNetworkLocationListener);
+        mLocationManager = null;
```
garvankeeley: In my branch I have mLocationManager as final in this class, so I was planning to merge that. I don't understand setting the location manager to null, is there a reason?

@cascheberg I apologize for being to be so slow with reviews, @crankycoder and I haven't been available much this week. Stumbler does need an update with this fix, and also we need to figure out why certain users are reporting stumbler pausing and not starting.

cascheberg: When this is null, then any following calls to `pauseGpsUpdates` will not have any effect, so the bug is fixed. But the changes in `MapFragment` alone are sufficient to fix the bug, I just felt like this would be less likely to regress because it does not depend on code outside the class.
I can push the updated version later.

cascheberg:
PR updated

garvankeeley:
Thanks!

PR #1321, opened 2014-12-03
Minor improvement regarding enabling GPS in MapLocationListener

cascheberg:
To be improved: you can see the GPS icon blinking once when scanning is paused and the app comes to the foreground (`MapFragment.onResume`).

In `MapLocationListener` there is this code:

```
enableLocationListener(true, mNetworkLocationListener);
enableLocationListener(true, mGpsLocationListener);

if (mapFragment.getApplication().isIsScanningPausedDueToNoMotion()) {
    pauseGpsUpdates(true);
}
```

If scanning is paused, the result of what this does is this:

```
enableLocationListener(true, mNetworkLocationListener);
enableLocationListener(true, mGpsLocationListener);
// then in pauseGpsUpdates:
enableLocationListener(false, mGpsLocationListener);
enableLocationListener(true, mNetworkLocationListener);
```

The updated code in this PR only runs one of the two paths.

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapLocationListener.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapLocationListener.java
@@ -108,11 +108,11 @@ class MapLocationListener  {
             mMapActivity.get().setUserPositionAt(lastLoc);
         }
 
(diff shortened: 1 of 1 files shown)


garvankeeley:
LGTM!

Issue #1162, opened 2014-10-29
Automate installation of the android SDK to get a buildserver setup


crankycoder:
Setting up the Android enviroment is a pain.  Let's wrap that up in a script that also checks that all the paths and enviroment variables are correct.

crankycoder:
We basically need to scrape the output from: `android list sdk --all --extended`, grab the name of the package and then issue : `android update sdk -u -a -t <package>`

khernyo:
You can use https://github.com/JakeWharton/sdk-manager-plugin for automated SDK install/update.

crankycoder:
Thanks @khernyo !  I've finally gotten around to doing this.  I'm hoping to get more contributors in now that we have a one step build.

PR #1328, opened 2014-12-05
Android SDK automatic installation support added

crankycoder:
Fix for #1162 - Android SDK is automatically installed if you haven't configured Android SDK already.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,12 @@ MozStumbler
 
 # Building a debug version from command line #
 
+The build system is smart enough to automatically download and install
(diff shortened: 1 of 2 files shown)


crankycoder:
Testing this is pretty easy, just remove your Android SDK by moving it to a different directory.

Remove all PATH changes that include pointers into the ADT.  Remove your ANDROID_HOME variable from your shell profile.

Start a new terminal window that has no awareness of an android sdk.

Run `make test`

That should do a full installation of all Android SDK bits, do a githubUnittest build, run the testsuite and you should get a passing test suite.

PR #1333, opened 2014-12-08
Motion Detect: more logging

garvankeeley:
- log activity shows isTimeWindowForMovementExceeded logic
- removed some log spam from the tile code


garvankeeley:
screenshot 2014-12-08 12 14 12

PR #1337, opened 2014-12-09
Update Gradle to version 1.0.0

### LINKED ISSUE #1327: Update Gradle version
The recently released Android Studio 1.0 RC requires a new Gradle version.

> Error:The project is using an unsupported version of the Android Gradle plug-in (0.12.2). The recommended version is 1.0.0.

The `build.gradle` file needs some adjustments and should use:

> classpath 'com.android.tools.build:gradle:1.0.0'
> classpath 'org.robolectric:robolectric-gradle-plugin:0.14.+'

And in `gradle-wrapper.properties`:

> distributionUrl=https\://services.gradle.org/distributions/gradle-2.2.1-all.zip

cascheberg:
@crankycoder This fixes #1327 
The `versionCode` and `versionName` are set automatically from the `build.gradle` file, so I removed the custom code.
Regarding the `PackageUtils` change: for testing I modified the version and `applicationId` and checked if the app version was still displayed correctly. Everything seems fine.
Do you see any problems with this?

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -17,8 +17,8 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:0.12.+'
-        classpath 'org.robolectric:robolectric-gradle-plugin:0.12.+'
+        classpath 'com.android.tools.build:gradle:1.0.0'
+        classpath 'org.robolectric:robolectric-gradle-plugin:0.14.+'
         classpath 'com.jakewharton.sdkmanager:gradle-plugin:0.12.+'
     }
 }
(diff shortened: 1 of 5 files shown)


crankycoder:
Thanks!  This looks great and works for me.

I ended up having to remove all my preferences for Android Studio.

To safely upgrade to 1.0 on my OSX machine, I had to remove:

```
rm -rf ~/Library/Caches/AndroidStudio*
rm -rf ~/Library/Application\ Support/AndroidStudio*
rm -rf ~/Library/Preferences/AndroidStudio*
```

@garvankeeley - blowing away all those preferences doesn't make you lose any of the project specific preferences.

Issue #1294, opened 2014-11-24
Add tests around Motion Sensor code


crankycoder:
We need to add some tests around the motion sensor code.

Initially, I'd like to see code that exercises the significant motion sensor, and verifies the state of all the scanners.

We should have a test that checks the notification drawer to check that it has been updated during pause or motion states.

Start/Stop of scanning behaviour should be verified for the pause and movement states.

We should really have a some documentation or diagram that describes the expected state changes for the system.

PR #1338, opened 2014-12-09
Features/1294 rebased sensor tests

crankycoder:
This closes off #1294 to add basic tests around the sensor code.

This adds basic roboelectric test coverage to the CellScanner and WifiScanner classes.

The primary goal of this is to get some concrete examples of using roboelectric and custom shadow classes to exercise the stumbler code.

Some preliminary documentation has been added into docs/testing.md, but it's still best to look at the example code to see how to write tests. 

We still need to get robotium tests for things where roboelectric's mocking system isn't sufficient and we need a real Android VM to test on.


garvankeeley:
In future try to separate bulk name changes to a single commit. Less to review, as I don't need to use my brain to review a name change commit (or any similar trivial refactoring commit). I just skim those.

garvankeeley:
This is good, all my comments are nothing more than nits. I don't want to leave this outstanding because there is a lot of code changes.

PR #1339, opened 2014-12-09
Use different application ID for debug builds

### LINKED ISSUE #1201: Build System: parallel install of different app versions
I'd like to keep the Play Store version installed in any case, so it would be great if Github and local builds could be installed in parallel.
Whenever I built the app locally, I have to remove the Play Store version and all its data before I can run it on my device. Is there a way around that? Am I doing it wrong?
And it's nice to have one permanent stable version and one dev/debug/reference version installed.
problem

cascheberg:
@crankycoder This fixes #1201 (or at least this is what I had in mind when I filed that issue).
It allows to install clean debug app versions (including version downgrades) while keeping the release version and all its data separately.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -98,6 +98,8 @@ android {
     buildTypes {
         debug {
             jniDebuggable true
+            applicationIdSuffix ".debug"
+            versionNameSuffix "-debug"
 
             // set this to false if you want your debug builds to have
             // standard unlabelled tiles.


crankycoder:
Cool.  Can you add an XML blob

```
<?xml version="1.0" encoding="utf-8"?>
<resources>
    <string name="app_name">Debug MozStumbler</string>
</resources>
```

to `android/src/debug/res/values/strings.xml` so we can also get a proper application name on build.

cascheberg:
Good idea. I tried that, but for some reason it did not work. But I added a debug versionNameSuffix.

I cannot work on this any further until next week, and I might be unavailable until then.

PR #1340, opened 2014-12-10
#1201 use different application ID for debug builds

crankycoder:
This is a minor update to #1339.

This patch will change the applicationId for debug builds as well as using "Debug Stumbler" as the app_name so that it's easy to identify release vs debug builds of the stumbler on your launcher.

I've removed the change to the versionName as I don't see any benefit to messing with versions when the applicationId is completely different anyway.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -99,7 +99,6 @@ android {
         debug {
             jniDebuggable true
             applicationIdSuffix ".debug"
-            versionNameSuffix "-debug"
 
             // set this to false if you want your debug builds to have
             // standard unlabelled tiles.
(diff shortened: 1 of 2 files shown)


cascheberg:
@crankycoder I do not see "Debug Stumbler" app name anywhere in debug builds. Is this working for you?

cascheberg:
@crankycoder Update: now I noticed that the app name is only set to "Debug Stumbler" for debug builds if the system language is set to English...

crankycoder:
dammit i10n.  i never thought of that.  Hmm.. we'll have to figure out how to localize the app_name strings.  Oh gradle.  I hate gradle.  So much.

Djfe:
I thought gradle worked well? Are there so mayn issues with it?

PR #1345, opened 2014-12-11
Copies stumble logs into the sdcard instead of deleting them

crankycoder:
This is an incomplete patch for #1255, but I just wanted some code that merges cleanly.

This _mostly_ works, except that there is a set of logs that are stored only in memory.  Whenever the save logs to disk flag is enabled, we must flush those memory buffers to disk, then do the regular upload thing.


android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/datahandling/ClientDataStorageManager.java
```diff
@@ -0,0 +1,159 @@
...
+    @Override
+    public synchronized void deleteAll() {
+        Log.i(LOG_TAG, "deleteAll() called");
+
+        if (mFileList.mFiles == null) {
+            return;
+        }
+
+        for (File f : mFileList.mFiles) {
+            delete(f.getName());
```
garvankeeley: r-, deleteAll is a way to bulk delete without calling heavy operations on every single file delete. I should have documented that.

garvankeeley:
I am merging this patch locally so I can try it out, will send you the merge.

garvankeeley:
After taking your patch here and patching onto dev, my patch can be applied on top:
https://gist.github.com/garvankeeley/b05ce2f2c2055266af1f

Sorry, this probably wasn't the best way to go about this :)

My patch will take care of the merge conflicts and has some minor changes.

crankycoder:
Let's talk about this one - I can't get your gist to merge with my code.  Pretty sure we can resolve any remaining issues pretty quick.

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/datahandling/ClientDataStorageManager.java
crankycoder: I've removed the deleteAll() code in the ClientDataStorageManager.  The base DataStorageManager also doesn't try to call the single file delete() method anymore so no attempt will be made to move the files into the sdcard.

Added some comments that deleteAll() is a special case.

Changes are in : 35c3da45ee31959d27f72002de417fd180671f72

PR #1346, opened 2014-12-11
Changed prefs to getInstance(Context) and the less-safe getInstanceWithoutContext()

garvankeeley:
Where possible getInstance(Context) should be used. The internal utility classes can use Prefs.getInstanceWithoutContext(), but must check the instance is not null. As part of the app instantiation, the Prefs instance is created, so there is no known reason that Prefs.getInstanceWithoutContext() would return null.


garvankeeley:
vng already reviewed this, landing.
https://bugzilla.mozilla.org/show_bug.cgi?id=1106584

Issue #1191, opened 2014-11-03
Add a simulation mode


crankycoder:
We need to have a simulation mode where the stumbler will receive a fake walking path, fake GPS, fake Wifi and fake Cell data.

We should be able to instrument this using a file as input so that we can later use the actual report logs or KML files to supply simulation data to the stumbler.

PR #1351, opened 2014-12-12
1191 Simulation Mode

crankycoder:
This adds a simulation mode to satisfy #1191.

This needs a testcase to verify that the application context being passed from the ScanManager down to GPSScanner, WifiScanner and CellScanner gets properly wrapped with the SimulationContext, but this branch should merge cleanly to dev.

Let's just merge it in and add more tests along the way.


android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/ScanManager.java
```diff
@@ -162,9 +189,21 @@ public synchronized boolean stopScanning() {
...
+
+        if (mWifiScanner != null) {
         mWifiScanner.stop();
+        }
+
+        if (mCellScanner != null) {
         mCellScanner.stop();
+        }
+
+        mGPSScanner = null;
```
garvankeeley, line 204: How is the null assignment used? I assume simulation needs this?

android/build.gradle
```diff
@@ -1,7 +1,7 @@
 // each of the version numbers must be 0-99
 def versionMajor = 1
 def versionMinor = 5 // minor feature releases
-def versionPatch = 4 // This should be bumped for hot fixes
+def versionPatch = 99 // This should be bumped for hot fixes
```
garvankeeley, line 4: is 99 ok here?

android/build.gradle
```diff
@@ -117,6 +117,51 @@ android {
             buildConfigField "boolean", "ROBOLECTRIC", "true"
         }
 
+        applicationVariants.all { variant ->
+            // I hate you gradle. The outputs list should always be of
+            // length 1.  But who knows what darkness lies in the
+            // heart of gradle.  The shadow does.
+            assert(variant.outputs.size == 1);
```
garvankeeley, line 124: gradely gradely gradle

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/ScanManager.java
```diff
@@ -162,9 +189,21 @@ public synchronized boolean stopScanning() {
         mLocationChangeSensor.stop();
         mMotionSensor.stop();
 
+        if (mGPSScanner != null) {
         mGPSScanner.stop();
+        }
```
garvankeeley, line 194: missing tab

garvankeeley:
Using it, working well so far. The wifi and cell counts don't increment, it would be nice if those did, so the simulation would be even more realistic. I am referring to the total counts, for instance the current cells is always 1, so with each report, the cell total would increment by one.

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/ScanManager.java
crankycoder, line 204: Yes.   We need to clear GPSScanner, WifiScanner and CellScanner and set them to be null values on stop of the ScanManager.  The ScanManager::startScanning method grabs the application context and wraps it up with a SimulationContext so that there is a place to obtain the simulation data.

Now that I write this, we don't really have to do this anymore since I've added the service locator code.  We could just inject something like the SimulationContext into the service locator.

android/build.gradle
crankycoder, line 4: I just set it to 99 for my own testing.  We should pull this out.

crankycoder, line 4: I'm actually going to just leave this as it's not clear what a 'good' value should be on the dev branch anyway.

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/ScanManager.java
```diff
@@ -121,24 +130,36 @@ public synchronized void startScanning(Context context) {
...
+            mMotionSensor = new MotionSensor(mAppContext, mDetectMotionReceiver);
         }
 
         if (mGPSScanner == null) {
-            mGPSScanner = new GPSScanner(context, this);
-            mWifiScanner = new WifiScanner(context);
-            mCellScanner = new CellScanner(context);
+            mGPSScanner = new GPSScanner(mAppContext, this);
+            mWifiScanner = new WifiScanner(mAppContext);
+            mCellScanner = new CellScanner(mAppContext);
```
cascheberg, line 144: @crankycoder these are always reassigned below. Does that make sense?

crankycoder, line 144: Looks like a bad merge.  The null check is not necessary before we
instantiate and assign.

More generally,  this is a good example of why we should pull out the mock
location generation from the SimulationContext and use an injected
simulation service.

Ugh.  I'll take a better look at this on Monday

cascheberg, line 144: @crankycoder just a reminder that you wanted to take a look at removing this presumably obsolete code block

PR #1353, opened 2014-12-16
 Homescreen widget support cont. #1316

lomash:
In reference to #1316, this is for the addition of home screen widgets to the application.


lomash:
I took a look at the code for report access. As far I can see, to get those numbers, we would require frequent calls to getQueuedCounts in DataStorageManager which isn't recommended according to the comments. 

Is there any other way that I can access the number of pending reports?

crankycoder:
One option might be to rework the data storage manager to emit the report
count using an asynchronous local broadcast as reports get queued up.

Djfe:
Maybe the code of notification could be refused somehow (or modified)

cascheberg:
If you want to get the session's reports, you can just use `StumblerService.getObservationCount()` (or `Reporter.getObservationCount()`). This is also what is displayed in the notification.
But if you really want the number of pending reports (i.e. those that have not been sent to the server yet), then you are right with what you wrote previously. But obviously this value can decrease during one session.

Djfe:
The notification should show the number of recorded observations of the current session only IMO (the same number as the one next to the binoculars) and use one of the getter functions that cascheberg mentioned

why are there two btw.? Is it, because of the structure of the app? (split up into client and service for fennec)

crankycoder:
@djfe - this is indeed confusing.   Let's just get this merged in for now and we should consolidate the different count methods.  I'd rather see just the current session.  The only reason we want to see the total of current session + past uploads is to have a visual reference point that uploads are working for the user.

@lomash - can you change the text size to 16dp?  When the counts become 2 digits, the right most digit gets truncated and is not visible.  After that, let's just merge this thing.

lomash:
@crankycoder Just pushed a new version with the text size changed to 16.

crankycoder:
Thanks @lomash!

There's a couple minor nits with layout, but I'll just apply those changes after merging.

PR #1356, opened 2014-12-16
Fix: notifications are never removed

cascheberg:
Sometimes notifications are never removed.

I think the problematic call stack is this:
- stopForeground and remove notifications
- stopScanning
- flush bundle
- update metrics and notification (scanner appears to be active still)
- scan manager only then sets scanning state to stopped


garvankeeley:
I haven't see this. The code change makes sense, lets go with it.

cascheberg:
This happens to me all the time. Maybe because on Android 5 the notification update rate is not limited? This is quite annoying because the notification persists even if you kill the app.

Issue #1362, opened 2014-12-19
Public help needed: Nexus 4 + 4.4 in India is missing Significant Motion sensor?


garvankeeley:
We have confirmed that Nexus 4 + 4.4 users in N.America have this sensor (which is actually a software sensor that depends a a number of hardware sensors).

In India, the report we have is that this device is reporting legacy motion sensor only. Obviously this is very weird. Can anyone with a Nexus 4 please check the first line in the Log screen in the app and let us know if it says 'Significant Motion sensor detected' or 'Legacy motion sensor'.

garvankeeley:
Custom 1.5.5 build for testing that removes the API check before trying to acquire the significant motion sensor.
https://drive.google.com/file/d/0B1sYPwjczMsQb2dtbkdWMGF6bVk

garvankeeley:
This doesn't make a difference, I will revert this change. 
The vast majority of inexpensive (<200 USD) 4.4 android models in India do not appear to have this sensor.

kbrosnan:
The Nexus 4 only has one [ROM](https://developers.google.com/android/nexus/images#occam) and two variants. One is a North American model and the other is a international model. My understanding is the difference is in the supported radio bands. I only have access to the NA model which has a significant motion sensor, maybe there is a European contributor that can check their device?

crankycoder:
This is unnecessary now.  We now use the signfiicant motion sensor if it's around, only white list known good devices (nexus and oneplus).  

Everything else uses standard accelerometer and hints from network location to wake to stumble.

PR #1363, opened 2014-12-19
India testing: ignore Build.VERSION.SDK_INT for Sig. Motion Sensor

garvankeeley:
My current guess is that Build.VERSION.SDK_INT doesn't seem to be working in Indian ROMs.
This code is for a test build, and likely will be reverted after 1.5.5 build

related to https://github.com/mozilla/MozStumbler/issues/1362


garvankeeley:
@crankycoder i am doing a 1.5.5 test build with this for Vinit in India to try out

Issue #1182, opened 2014-11-01
Add a "file a bug" button


crankycoder:
We frequently get bug reports where we inevitably ask for device and rom info.

We should really be able to file bugs from the device which captured this information for us.

Maybe implement this as a hook into ACRA?

SteveALee:
I tried to supply info from the log but could not cut and paste it.
I then tried a screen dump but after switching to desktop GitHub in Firefox for Android it reset back to an empty mobile form after I tried to insert the image.

Bugsbane:
Yes, please. I just accidentally filled what I thought was a Stumbler big on the ichnara project because I couldn't find a bug report link in Stumbler. Then I tried to insert a screenshot, to get the same form blanking bug in GitHub mobile, then I tried to copy and paste the log, but found that you can't select text in the log... just like the previous commenter.

SteveALee:
So glad it is not just me! :)

johnnyxh:
I've been working towards adding this feature today, mostly on the UI portions for a new bug reporting activity. I've been meaning to use ACRA for this by taking in bug title/description/reproducibility fields from my new bug reporting activity and stuffing it into the CUSTOM_DATA field, then I can manually send a "crash report" with just that data. Problem is I don't know where these reports are sent (or if I even have access to them) to see if they are working properly. I guess I'm not too sure how I'd test to see how this works without having to make my own destination for these ARCA reports.

crankycoder:
@johnnyxh have you tried installing the ACRA server on your own machine?  Standing up an Acralyzer server is not too bad.     Can you try the instructions over here?  https://github.com/ACRA/acralyzer/wiki/setup

johnnyxh:
@crankycoder I set up my own ACRA database on Cloudant along with Acralyzer, and tried it out. It works decently well, though at first glance these look like regular crash reports. The stack trace sets apart regular crashes from bug reports, since the stack trace entries contain "Requested by the developer". Not sure if this is a good enough distinction, but you can see the bug title/description/etc fields under the custom field along with the usual information sent in an ACRA report.

Bugsbane:
Sweeeeeeeeeeeeeet. :)

cascheberg:
This issue is fixed and should be closed.

PR #1369, opened 2014-12-24
Added "File a Bug" button

johnnyxh:
Attempting to address #1182 by adding a "File a Bug" button under the settings/preferences activity.

This spawns a new activity that prompts the user to fill in a title for the bug, the type of bug it is (from a preset list), description of the bug, and how to reproduce the bug. An optional name and email field are there just in case.

These reports appear as ACRA reports, though the stack trace will always read "Report requested by developer" and contain the user provided information in the CUSTOM field under the stack trace. Aside from that, all the other information in an ACRA report is available as normal. I hope that is a good enough distinction between manual bug reports and crashes.

Some screen shots of the bug report form: 

device-2014-12-23-211915

device-2014-12-23-211928

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -85,6 +85,15 @@
                 android:name="android.support.PARENT_ACTIVITY"
                 android:value=".client.navdrawer.MainDrawerActivity" />
         </activity>
+        <activity
+            android:name=".client.subactivities.FileBugActivity"
+            android:configChanges="orientation|screenSize|keyboardHidden"
+            android:parentActivityName=".client.subactivities.PreferencesScreen" >
+            <!-- Parent activity meta-data to support 4.0 and lower -->
+            <meta-data
+                android:name="android.support.PARENT_ACTIVITY"
+                android:value=".client.subactivities.PreferencesScreen" />
+        </activity>
 
         <service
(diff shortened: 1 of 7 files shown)


crankycoder:
Nice!  This looks really great!  I'm going to give this a spin on my local setup.

garvankeeley:
Um, this is a week old. We should try to merge this ASAP.

crankycoder:
Sweet.  This looks really good.  I'm seeing nice bug reports showup like this: 

acralyzer___myapp

crankycoder:
Thanks @johnnyxh !

johnnyxh:
I just realized the toast message and the field error messages are hardcoded in the FileBugActivity class rather than being pulled from the strings.xml file. I'm not near a computer right now, but I could make the changes tomorrow.

crankycoder:
No worries.  Just make a new pull request and we can review and approve the
patch.

Thanks!

VolMi:
I just started translating those strings to German. Then I noticed that getting German/Urdu/Chinese bug reports may only be of peripheral use... :-)
What's the attitude on this? Should we translate the strings but declare that bugs be reported in English? Or, since a few co-Germans seem to drift around here, is German an allowed language, maybe also French and Spanish, but others aren't? Or are all languages fine?

PR #1376, opened 2014-12-31
Do not log opt-out SSIDs

### LINKED ISSUE #149: Shouldn't "_nomap" mean no-log as well
Most likely I'm missing something here, so please do correct me if I'm wrong.

Looking at the class SSIDBlockList in https://github.com/dougt/MozStumbler/blob/master/src/org/mozilla/mozstumbler/SSIDBlockList.java 
it's a simple string filter that allow networks to opt-out of the mapping by appending "_nomap" to their SSID.
And that could be done on the basis of privacy concerns and countless other reasons, which are all irrelevant to the point I'm about to raise.

Then checking line 131 of WifiScanner. After it checks SSIDBlockList, it logs the scanResult any way. 
So regardless of the reason why that Network opted-out of the mapping, the application still logs that Network scan result.
See https://github.com/dougt/MozStumbler/blob/master/src/org/mozilla/mozstumbler/WifiScanner.java#L131

My point being: if a network opt-out by using the "_nomap" then the app should respect that at all levels and not log anything about that network.
It looks like the "_nomap" means: Hey, we won't plot your network in a map nor make it searchable for third-parties, but we'll still store data about it regarless of your opt-out instruction.
Surely verbose logging eases debugging, but the user's options should be respected, don't you think?

cascheberg:
Fixes #149

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiScanner.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiScanner.java
@@ -9,7 +9,6 @@ import android.content.Intent;
 import android.net.wifi.ScanResult;
 import android.net.wifi.WifiManager;
 import android.net.wifi.WifiManager.WifiLock;
-import android.os.Build;
 import android.support.v4.content.LocalBroadcastManager;
 import android.util.Log;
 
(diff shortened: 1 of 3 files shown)


android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiScanner.java
```diff
@@ -182,6 +181,10 @@ synchronized void deactivatePeriodicScan() {
     }
 
     public static boolean shouldLog(ScanResult scanResult) {
+        if (SSIDBlockList.isOptOut(scanResult)) {
+            Log.w(LOG_TAG, "Blocked opt-out SSID");
```
crankycoder: I don't think we need the warning level log here.  Let's change that to a Log.d for debugging.

cascheberg:
@crankycoder PR updated

garvankeeley:
Thanks, this looks good. 

ASIDE: I just noticed the existing method shouldLog should be called shouldReport. "Log" implies internal logging, not reporting publicly.

PR #1378, opened 2014-12-31
Added telemetry for Fennec.

garvankeeley:
Required using reflection to grab the telemetry class in Fennec due to the partitioning of the build system; stumbler has no access to gecko.


garvankeeley:
Corresponding Fennec bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1116810

PR #1383, opened 2015-01-02
Minor fixes

cascheberg:
Minor fixes


crankycoder:
I need to know what bugs this fixes and some accompanying tests would be nice.   There's mockito support for doing mocks now, that should at least help a bit.

cascheberg:
This is more of a cleanup based on IDE hints.
1. the critical error toast is never displayed
2. in `fragment_developer_options.xml` some string references are in a wrong format
3. zh-rTW plurals do not use quantity one
4. the other changes improve RTL support

garvankeeley:
Thanks for the cleanup work!

PR #1384, opened 2015-01-04
Hardcoded strings replaced with references to strings.xml

johnnyxh:
Small fix to #1369. There were two instances where I hard coded string messages for the user. Those are now being pulled from the strings.xml file.


crankycoder:
Awesome. Thanks!

Issue #1323, opened 2014-12-03
Coverage map tiles are still not expiring correctly.


crankycoder:
Coverage tiles are still not expiring correctly.

I've reproduced this problem by stumbling over a new area with one device, then viewing the coverage maps on the same device, along with a separate device.

After 24 hours, the second device correctly shows the stumble path while the first device - the one used to stumble which had a copy of the old coverage tiles continued to display the old coverage tiles.  The effect is that one device shows blue dots, the other shows the area as 'unstumbled'. 

Strategy to fixing this problem is :
- Wrap the osmdroid tile downloader code with unit tests first around the conditional GET caching.
- Force tiles to expire on a hard deadline of 6 hours.

cascheberg:
One thing I noticed in `TileDownloaderDelegate.isTileCurrent`: you request the URL and add the etag header, but  you throw away the response if the response code is not 304. But in that case, the response already contained the full requested tile, right? Wouldn't it be better to move that code to `downloadTile` and only send and check the etag there?

crankycoder:
Crap. That's probably correct.  I was doing an HTTP HEAD before but that wasn't working consistently - so I switched to using the etag method.  I'll take a look at your PR tonight.

cascheberg:
I tried to reproduce this bug, but so far I couldn't. Could it be that you get tiles from the `LRUMapTileCache` or so, without `isTileCurrent` ever being called? Did you kill the app and then check again?

crankycoder:
Hrm.  The LRUMapTileCache may very well be the problem.  That would be consistent with the behavior I saw.

Here's what I saw - it was while using two separate devices.  

I had a OnePlus One running CM11s (Android 4.4.4) and a Moto G 2013 (1st gen) running Android 4.4.3.

The Moto G had a previous set of coverage maps from a couple days earlier.  The One Plus was doing a new stumble and covered previously uncovered areas where the coverage tiles should have been out of date on the Moto G.

What I saw was the OnePlus and the Moto G had differing coverage tiles, which shouldn't have happened as the code as is should do a conditional GET on all tiles all the time, but just pass in the etag to get the new data.

In general, I think the LRUMapTileCache should probably just be removed, but I think it's tightly coupled with the way the tiles are rendered into the viewport.  I remember trying to remove it months ago, but I failed and the mapview stopped rendering.

[...]

PR #1388, opened 2015-01-07
Increase cache time of map tiles

cascheberg:
This PR increases the cache time of map tiles to 7 days, while leaving the cache time of the coverage overlay unchanged at 12 hours.
From what I read from the code, the cache is only cleared when it is exceeding `OSMConstants.TILE_MAX_CACHE_SIZE_BYTES` (which is at a pretty high value of 600MB).
So this should only influence how long cached map tiles are assumed to be current and thus reduce data usage.

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java
@@ -485,6 +485,7 @@ public class MapFragment extends android.support.v4.app.Fragment
                         new String[]{BuildConfig.TILE_SERVER_URL});
             }
             System.gc();
+            mHighResMapSource.setCacheTime(7 * 24 * 60 * 60 * 1000);
             mMap.setTileSource(mHighResMapSource);
         } else {
             if (mLowResMapOverlayHighZoom != null) {
--- a/android/src/main/java/org/mozilla/mozstumbler/client/mapview/tiles/CoverageOverlay.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/mapview/tiles/CoverageOverlay.java
@@ -32,6 +32,7 @@ public class CoverageOverlay extends AbstractMapOverlay {
                 ".png",
                 new String[] { coverageUrl });
         this.setLoadingBackgroundColor(Color.TRANSPARENT);
+        coverageTileSource.setCacheTime(12 * 60 * 60 * 1000);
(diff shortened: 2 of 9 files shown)


crankycoder:
I'm going to r- this mostly because OSMDroid is buggy as is, and I don't want to accept any patches that affect that section of the codebase without some tests that show that the behavior in code is what we expect it to be.  As it stands right now, the conditional GET code I implemented doesn't seem to do the right thing.  Tiles are _not_ expiring as expected (see: #1323).

@cascheberg - I'll follow up with you on email so that we can build some tests around this code.  It'll be a good test of the enhanced testing infrastructure that's available.

crankycoder:
This won't work - or at least it won't work reliably.

https://github.com/mozilla/MozStumbler/blob/dev/android/src/main/java/org/mozilla/osmdroid/tileprovider/modules/TileIOFacade.java#L215

trims the cache using on every single tile save but it only does so by checking the last modified timestamp.   This won't work on Android because mtime is never reliable and is marked as 'declined' as a bug in Android:

https://code.google.com/p/android/issues/detail?id=1699

I think it's best to just delete the entire cache cleanup code in OSMDroid as it's invoking a synchronized call on _every_ tile save.    We'd be better off to just keep an index of tile paths and file save times and just cull the tiles in MapFragment::onResume and MapFragment::onPause.

cascheberg:
@crankycoder This would at least reduce the amount of actual HTTP requests for map tiles in `TileDownloaderDelegate.downloadTile()` because of `if (systemClock.currentTimeMillis() < serializableTile.getCacheControl())`

crankycoder:
I really don't think the client should be overriding the server's cache time limits.  In the case where the network is completely unavailable, we already provide the tile from disk if it's there already.  

The cache control setting should be changed on the server side if the intent is to keep tiles around longer.

cascheberg:
Hmm good point, but currently this is always overridden and set to 12 hours, which is not much for map data: https://github.com/mozilla/MozStumbler/blame/dev/android/src/main/java/org/mozilla/osmdroid/tileprovider/modules/SerializableTile.java#L147

crankycoder:
Doh.  I did do that didn't I?  Crap.

Alright, let's refactor this a bit.  The networking code should really not tamper with the cache-control settings.  

I'd like to see a chain of responsibility pattern where we try to load the tile from disk with a longer cache period falling back to loading from the network.  This shouldn't be too bad as the MapTileRequest already contains the filepath of the tile.    The serializable tile also allows for arbitrary headers to be added, so we can add an extra one to store the overridden cache setting.

PR #1393, opened 2015-01-10
Blocklist additions for in-flight WiFi

JensErat:
For sources see commit messages.


cpeterso:
Looks good. Thanks, @JensErat!

PR #1398, opened 2015-01-12
Improve etag usage when downloading map tiles

cascheberg:
This removes the extra request that is done to only check the map tile etag status. Instead, the etag is checked when an actual tile download happens. This way, the response can be used in case of a non-matching etag as well. See https://github.com/mozilla/MozStumbler/issues/1323#issuecomment-69102515


crankycoder:
Thanks @cascheberg  - I'm going to merge the isTileCurrent and the downloadTile code as it doesn't make sense to separate the two.  I'll make those changes and we'll get this in for the 1.5.5 release.

crankycoder:
Closing as #1400 supercedes this and we've moved to not accepting patches to OSMDroid related code unless tests are included.

PR #1400, opened 2015-01-13
Conditional get rework

crankycoder:
This is an addendum to #1398. 

This removes the isTileCurrent check and just moves it together with the actual tile download.

The significant change over @cascheberg's original patch is that the TileIOFacade will now return a SerializableTile after saving the tile.  This can then be immediately used to load an instance of Drawable.

An HTTP 304 will also check that the tile can be properly loaded from disk in the case that the tile was somehow corrupted on disk.  If this is the case, the tile's etag is completely invalidated and the returned tile is set to NULL.


crankycoder:
This still needs a testcase.

android/src/main/java/org/mozilla/osmdroid/tileprovider/modules/TileDownloaderDelegate.java
```diff
@@ -52,80 +49,31 @@ public TileDownloaderDelegate(INetworkAvailablityCheck pNetworkAvailablityCheck,
...
-
-        final String tileURLString = tileSource.getTileURLString(tile);
-
-        if (tileURLString == null || tileURLString.length() == 0) {
-            return false;
-        }
-
-        if (urlIs404Cached(tileURLString)) {
-            return false;
+            return tileSource.getDrawable(serializableTile.getTileData());
```
cascheberg, line 82: @crankycoder Almost all `return null` in this method should be replaced by returning the cached tile, similar to this. Otherwise there is no map if offline etc.

crankycoder, line 82: I'd rather see that added in as a separate class and implemented as a proxy
to the delegate.

Separate the concerns and you can easily test that.  Load the image from
the tiledownloader,  if that fails,  try to load from disk.

crankycoder:
@garvankeeley ok, tests are in.  You'll find tests that cover the following cases:
- basic HTTP200
- HTTP 304 conditional GET with on disk tile 
- HTTP 304 conditional GET with a tile that has zero bytes - which should result in a NULL instead of a tile
- Network outage should return tile if it exists on disk
- Network outage should return null if no tile exists on disk

crankycoder:
One more note - I had to modify the build because gradle and android are terrible.  3461ff7 includes patches to android/build.gradle and the Groovy test runner that will manually copy over test resources so that you can use test fixture data.  

rocket_racoon

android/src/main/java/org/mozilla/osmdroid/tileprovider/modules/SmartFSTileLoader.java
```diff
@@ -0,0 +1,84 @@
...
+        ITileSource tileSource = smartFSProvider.getTileSource();
+
+        if (tileSource == null) {
+            return null;
+        }
+
+        final MapTile tile = pState.getMapTile();
+
+        // if there's no sdcard then don't do anything
+        if (!smartFSProvider.getSdCardAvailable()) {
```
garvankeeley, line 42: We don't depend on an SD card.

crankycoder, line 42: That's the same code that was in the TileLoader before.  I just moved the code and didn't want to add any functional changes in this patch.

PR #1403, opened 2015-01-14
Show MLS coverage overlay when offline

### LINKED ISSUE #1401: Full offline map support
Use case:
1. use the app, some map tiles will be cached, close the app
2. disable all internet connections
3. open the app again

Result:
- The OSM tiles are visible, but the MLS overlay is not
- If you enable internet access briefly, then all previously cached overlay tiles become visible

cascheberg:
This fixes #1401 

Initially no internet access available:
- `CoverageOverlay` will have URL = `null`, but yet it is able to display cached tiles

Internet access becomes available:
- `sCoverageUrl` will be updated
- coverage tiles are removed
- coverage tiles are then added back with the right URL

Intended changes:
- removed check for `mCoverageTilesOverlayLowZoom != null` is still done in `initOnMainThread()`
- just move `scanner` code so that `initOnMainThread()` is always called
- always remove coverage layers when `sCoverageUrl` is set, they are then added back in `initOnMainThread()`

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java
@@ -300,10 +300,9 @@ public class MapFragment extends android.support.v4.app.Fragment
         }
 
         void getUrlAndInit() {
-            if (isGetUrlAndInitCoverageRunning.get() || mCoverageTilesOverlayLowZoom != null) {
+            if (!isGetUrlAndInitCoverageRunning.compareAndSet(false, true)) {
                 return;
             }
-            isGetUrlAndInitCoverageRunning.set(true);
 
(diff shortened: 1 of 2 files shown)


garvankeeley:
Thanks for fixing this! This is ready to land with that added comment in the code.

cascheberg:
@garvankeeley PR updated

That exception was "hit as part of normal offline handling" before the change as well, but now it just does not `return;` anymore.

I noticed that I had to add another check in `XYTileSource` so that `TileDownloaderDelegate` does not try to request `null` URLs.

garvankeeley:
Looks good!

My point being, exception handling like this:
  catch (ex) { abort method in some way } 
is unlikely to be misunderstood, or refactored in later patches to break this behaviour,
  catch (ex) { eat exception } now do important work because of having eaten exception 
requires documentation. Later authors may not realize the value of having eaten that exception.

PR #1407, opened 2015-01-16
Reduce data: cellular map (mLowResMapOverlays) uses same cache as wi-fi (mHighResMapSource)

cascheberg:
Low-res and high-res map tiles are using the same URL, but not the same name for caching.
If the name is the same, then map tiles already downloaded in high-res mode can be used in low-res mode as well (and vice versa).
This is already the case if `!isMLSTileStore`. In that case, both are using the name "MapquestOSM".

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java
@@ -478,7 +478,7 @@ public class MapFragment extends android.support.v4.app.Fragment
             if (!isMLSTileStore) {
                 mHighResMapSource = TileSourceFactory.MAPQUESTOSM;
             } else {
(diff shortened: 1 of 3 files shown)


Djfe:
thx for working on it ;)
on Wi-Fi MozStumbler should replace low-res with high-res tiles though, I think.

cascheberg:
@Djfe I'll try to clarify this: this PR reduces data usage, but does not improve or change the visual appearance of any map mode.
The tiles that you see in **low-res mode at a high zoom level** are the very same tiles that you see in **high-res mode on a certain lower zoom level**. That is why they can share the cache.
The difference is that high-res mode loads specific high zoom tiles if you zoom in, while low-res mode just upscales the low zoom tiles.

garvankeeley:
Much appreciated fix!

Issue #1409, opened 2015-01-17
[Regression] MozStumbler's UI gets unresponsive after scanning a while (low end devices)


gpoo:
Since 1.5.4 (I think), my Nexus One takes several minutes to respond the UI after scanning for a while. Say for example:
- Start Mozstumbler and press Start Scanning.
- Walk for around 15 minutes
- See MozStumbler again, and try to interact with it ('stop' button, load the menu, get actual location), and it takes several minutes (possibly 5 or more) to show something.

In the meantime, Android display a dialog  to either "Force stop" or "Wait" the application.  I have to press "Wait" over and over, in spite that MosStumbler seems freeze.  But it is just slow, once it stops scanning, the UI becomes responsive again.

At the beginning I thought it could be too many points to display.  But later it happened with a clear screen, and only some dots (enough to see in a screen without panning).  I tried rebooting, nothing changed.  I tried only low resolution tiles, nothing changed.  Previously I had enabled saving data to a file", but when I disabled it, nothing changed.

No other applications are running in parallel. I use the phone only to Stumble with a different carrier.

It became responsive again once I downgraded it to 1.5.3.

cascheberg:
I have only scanned a few times for a short while since upgrading to 1.5.5 (Update: I did not notice this using 1.5.4), but I also noticed that responsiveness was worse. After scanning for a while, pressing the stop button and scrolling the map became laggy on my Nexus 4, Android 5.0.1.

gpoo:
@cascheberg I've heard many times that Android 5 is painfully slow (besides that drains the battery faster). Could it be that?

FWIW, in a Nexus 5, with Android 4.4.4 I don't see any issues, likely because it has better hardware (I won't upgrade to Android 5 yet :-)

cascheberg:
@crankycoder @lomash I did some regression testing and I am pretty sure that the widget (94c77b8 #1353) is causing performance problems on my device.
I always installed a clean build, put it on a desk where it had a stable GPS fix, then started the scan. The screen was on all the time, and I manually disabled the motion detection code so that it kept scanning. After waiting for a few minutes (~ 5-10) I scrolled the map and it was laggy. Stopping the scan fixed the problem. Starting it again immediately brought back the problem. Killing and restarting the app while the service kept scanning also fixed the problem. I did not actually set up the widget on my home screen.

@gpoo Android 5 is running perfectly fine on my Nexus 4. Also, that should not be the problem because it did not happen in previous stumbler versions.

[...]

PR #1413, opened 2015-01-19
Features/backout homescreen widget

crankycoder:
This completely backs out the homescreen widget due to performance regressions as noted in #1409.


crankycoder:
dang it.

PR #1414, opened 2015-01-20
496 - adds changelog on startup

crankycoder:
I've added a changelog on startup.

The changelog is not localized as we currently do not have a process to do localization prior to a new release.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,7 +1,7 @@
 // each of the version numbers must be 0-99
 def versionMajor = 1
 def versionMinor = 6 // minor feature releases
-def versionPatch = 1 // This should be bumped for hot fixes
+def versionPatch = 99 // This should be bumped for hot fixes
 
 // Double check the versioning
 for (versionPart in [versionPatch, versionMinor, versionMajor]) {
--- a/android/src/main/java/org/mozilla/mozstumbler/client/ClientPrefs.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/ClientPrefs.java
@@ -26,6 +26,16 @@ public class ClientPrefs extends Prefs {
     private static final String DEFAULT_SIMULATION_LAT_LONG = "default_simulation_lat_lon";
     private static final String MIN_BATTERY_PCT = "min_battery_pct";
     public static final int MIN_BATTERY_DEFAULT = 15;
+    public static final String LAST_VERSION = "last_version";
+
+    public long getLastVersion() {
+        return getPrefs().getLong(LAST_VERSION, 0);
+    }
+
+    public void setDontShowChangelog() {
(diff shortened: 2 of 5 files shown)


crankycoder:
@garvankeeley travis died on grabbing one of the dependencies. blargh.

garvankeeley:
The bulleted list can be done in strings.xml
For example:
`<string name="whats_new"><li>More conservative power saving options</li>`
`\n<li>Home screen widget</li>`
`\n<li>Wifi blocklist updates</li></string>`

Will look like this:
https://www.dropbox.com/s/6iv4npwu19rwaov/Screenshot%202015-01-21%2018.25.27.png

garvankeeley:
Needs to be wrapped in a scrollview, for scrolling

garvankeeley:
I am thinking we shouldn't show it on first run, rather than have 2 dialogs at startup. First-run users are less concerned with new features.

crankycoder:
I've disabled the firstrun changelog thing. Scrollview isn't needed as AlertDialog will automatically scroll if the text overflows vertically.  I think that's everything.

cascheberg:
I have some comments:
- you can also move the "OK" to the `strings.xml`
- I guess you can now completely remove the `dialog_whatsnew.xml`
- shouldn't it be "What's new" (instead of "Whats new")?

crankycoder:
@garvankeeley this should merge cleanly now.  You can test by using the diff @ version afa5c1a where I commented out the pref update to ensure that the changelog displays once and only once, but not on first run.

@cascheberg thanks for the review - layout XML file has been dropped and strings have been updated.

garvankeeley:
One more thing, if the strings.xml name="whats_new" is blank, the dialog shouldn't show. In case we don't want it to show. (Maybe it does this, I haven't looked)

crankycoder:
Added the check for whats_new text length in 48c5b9d.  

Dang it.  There are no tests for MainDrawActivity.  I'm going to add a test in a separate PR.

garvankeeley:
This didn't scroll for me when I tried, are you sure this is supposed to work? (I rotated the screen one the dialog was showing, and not all lines were shown any more, and I couldn't scroll)

crankycoder:
Hrm.  Here's two screenshots where I copy/pasted the sample text several times.  

The bottom screenshot is where I've scrolled down to the bottom - you can see the scrollbar on the right hand side.

screenshot_2015-01-27-08-34-12
screenshot_2015-01-27-08-34-23

crankycoder:
Bah. This works on both the OnePlus and Moto G over here.  I'll try with the Nexus 4 running Lollipop.

crankycoder:
Works for me on Lollipop on the Nexus 4.

PR #1416, opened 2015-01-21
Update German translation

cascheberg:
Do you want to have a hint added that bug reports are preferred / should be filed in English?


garvankeeley:
Assuming a non-english report is rare, I am sure we can get it translated

PR #1418, opened 2015-01-21
1417 - drop wifi blocklist

### LINKED ISSUE #1417: Drop wifi blocklists
Now that Firefox Android embeds the stumbler in a passive stumbling mode, the wifi blocklist does more harm than good for the server side.

We are pulling out the wifi blocklist from the stumbler so that filtering can occur on the server side.

crankycoder:
Fixes #1417 and removes all references to the wifi blocklist.  We still keep the opt-out '_nomap' postfix check.


garvankeeley:
Reminder, this needs a rebase

crankycoder:
I'm all about rebase, about rebase, no checkout.

This should be ok now.

PR #1419, opened 2015-01-21
Added power saving settings screen.

garvankeeley:
Added power saving section to drawer view.
Default power saving depends on presence of sig. motion sensor

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -95,6 +95,16 @@
         </activity>
 
         <activity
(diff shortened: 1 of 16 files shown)


garvankeeley:
BTW, that last commit won't fix the 'Settings' screen not showing a '<' in the upper-left. That one inherits from PreferenceActivity, which won't put an up button in the title bar (yes the "up button" is a left-arrow, because Android uses the term 'back button' for the hardware/software back button next to the home button). PreferenceFragment will do that, but requires a higher API level, so we can change that in future when we raise the API level.

PR #1421, opened 2015-01-22
Update translations

cascheberg:
Updated translations for #1419 

@garvankeeley It's not nice to make major string changes minutes before a major release ;)


garvankeeley:
Thanks! We'll update again soon.

PR #1432, opened 2015-01-24
Rate limit update checks (github builds)

### LINKED ISSUE #324: Only show the update notification once on app start

cascheberg:
This fixes #324 by only checking for updates once every 6 hours, and on app start, but still makes sure that updates are checked if there was no useable network connection previously. This only affects github builds.

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/Updater.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/Updater.java
@@ -32,6 +32,8 @@ public class Updater {
     private static final String LOG_TAG = AppGlobals.makeLogTag(Updater.class.getSimpleName());
     private static final String LATEST_URL = "https://github.com/mozilla/MozStumbler/releases/latest";
     private static final String APK_URL_FORMAT = "https://github.com/mozilla/MozStumbler/releases/download/v%s/MozStumbler-v%s.apk";
+    private static final long UPDATE_CHECK_FREQ = 6 * 60 * 60 * 1000; // 6 hours
+    private static long sLastUpdateCheck = 0;
 
     public Updater() {
     }
@@ -40,11 +42,21 @@ public class Updater {
         return !NetworkInfo.getInstance().isWifiAvailable() && ClientPrefs.getInstance(c).getUseWifiOnly();
     }
 
+    public boolean checkForUpdatesRateLimited(final Activity activity, String api_key) {
+        if (System.currentTimeMillis() - sLastUpdateCheck < UPDATE_CHECK_FREQ) {
+            return false;
+        }
+        return this.checkForUpdates(activity, api_key);
+    }
 
     public boolean checkForUpdates(final Activity activity, String api_key) {
 
         // No API Key means skip the update
         if (api_key == null || api_key.equals("")) {
+            return false;
+        }
+
+        if (!NetworkInfo.getInstance().isConnected()) {
             return false;
         }
 
@@ -99,6 +111,7 @@ public class Updater {
             }
         }.execute();
 
+        sLastUpdateCheck = System.currentTimeMillis();
         return true;
     }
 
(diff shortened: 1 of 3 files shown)


android/src/main/java/org/mozilla/mozstumbler/client/Updater.java
```diff
@@ -40,6 +42,12 @@ public boolean wifiExclusiveAndUnavailable(Context c) {
         return !NetworkInfo.getInstance().isWifiAvailable() && ClientPrefs.getInstance(c).getUseWifiOnly();
     }
 
+    public boolean checkForUpdatesRateLimited(final Activity activity, String api_key) {
+        if (System.currentTimeMillis() - sLastUpdateCheck < UPDATE_CHECK_FREQ) {
```
crankycoder, line 46: The dependency on System.currentTimeMillis() should be replaced with grabbing ISystemClock from the ServiceLocator.

You can then instrument the time by using the MockSystemClock.

https://github.com/mozilla/MozStumbler/blob/dev/android/src/main/java/org/mozilla/mozstumbler/svclocator/services/MockSystemClock.java

``` java
        clock = new MockSystemClock();
        ServiceLocator.getInstance().putService(ISystemClock.class, clock);
```

will replace the real clock with something you can play with under test.  The LocationChangeSensorTest is a reasonable example to look at:

https://github.com/mozilla/MozStumbler/blob/dev/android/src/test/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/LocationChangeSensorTest.java#L62

android/src/main/java/org/mozilla/mozstumbler/client/Updater.java
```diff
@@ -40,6 +42,12 @@ public boolean wifiExclusiveAndUnavailable(Context c) {
...
+            return false;
+        }
+        return this.checkForUpdates(activity, api_key);
```
crankycoder, line 49: Let's just inline the `checkForUpdates` method call or at least turn it private as there's no reason to have two publicly accessible entry points into this code.

crankycoder, line 49: In fact, we should probably just inline the code in `checkForUpdatesRateLimited` into `checkForUpdates`.  We already have a test class over here:

https://github.com/mozilla/MozStumbler/blob/dev/android/src/test/java/org/mozilla/mozstumbler/client/UpdaterTest.java#L46

You should be able to just add instrumentation of the MockSystemClock and verify that the update hasn't actually occured.

crankycoder:
@cascheberg I've added some comments on how to add some testing to this.  In short, I think you ought to be able to just add `setup` method to the UpdaterTest code to setup the ISystemClock service in the ServiceLocator.

Use the example in https://github.com/mozilla/MozStumbler/blob/dev/android/src/test/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/LocationChangeSensorTest.java#L62

After that, you should be able to manually set the system clock time in the `testUpdater` method and add appropriate assertions to verify that rapid calls to `checkForUpdates` will just short circuit and fail.

cascheberg:
@crankycoder Thanks, I'll try to have a closer look soon!

crankycoder:
@cascheberg the test infrastructure leaves a lot to be desired still - so if there are particular pain points - please **do** yell and say "hey - doing X sucks and makes it too hard to test".  

_sigh_.  java.  android.  land of ridiculousness.

crankycoder:
This has been sitting in queue for a long time. I'll add the test.

PR #1434, opened 2015-01-27
Reorganize the codebase into a reusable library.

crankycoder:
This is a first pass at extracting a reusable library from the Stumbler.

This only includes patches to push the service locator code into a library, moves some test cases into the library and pushes the logging services down into the library.

No android services have been pushed into the library yet, this patch is getting too large as it is.

The most significant change is that the ServiceLocator will now return a proxy instead of the real underlying service implementation.  This is to allow binding services in as static members of a class such as loggers.

In the event that a method is invoked on a proxy before the ServiceLocator has been configured, the proxy will execute a no-op and return either NULL, 0 or '\u0000' for methods that require return values.


android/src/main/java/org/mozilla/mozstumbler/service/utils/TelemetryWrapper.java
```diff
@@ -1,14 +1,15 @@
 package org.mozilla.mozstumbler.service.utils;
 
 import android.util.Log;
-import org.mozilla.mozstumbler.service.AppGlobals;
+
+import org.mozilla.mozstumbler.svclocator.services.log.LoggerUtil;
+
```
garvankeeley, line 6: nit on newlines

android/src/main/java/org/mozilla/mozstumbler/service/AppGlobals.java
```diff
@@ -4,12 +4,13 @@
 
 package org.mozilla.mozstumbler.service;
 
+import org.mozilla.mozstumbler.service.core.logging.ClientLog;
+
```
garvankeeley: remove this import, not used

android/src/test/java/org/mozilla/mozstumbler/client/mapview/MapFragmentTest.java
```diff
@@ -9,7 +9,7 @@
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.Mockito;
-import org.mozilla.mozstumbler.service.AppGlobals;
+import org.mozilla.mozstumbler.svclocator.services.log.LoggerUtil;
 import org.robolectric.RobolectricTestRunner;
 import org.robolectric.annotation.Config;
```
garvankeeley, line 15: On line 18, the import for Mokito.mock can be removed, not used anymore

android/src/test/java/org/mozilla/osmdroid/tileprovider/modules/TileDownloaderDelegateTest.java
```diff
@@ -40,15 +36,13 @@
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.spy;
-import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
```
garvankeeley: verify is no longer used

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/svclocator/services/log/ProductionLogger.java
```diff
@@ -0,0 +1,46 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.mozstumbler.svclocator.services.log;
+
+public class ProductionLogger implements ILogger {
+
+    public void w(String logTag, String s) {
+
+        android.util.Log.w(logTag, s);
+
```
garvankeeley: nit newlines

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/svclocator/services/log/ProductionLogger.java
```diff
@@ -0,0 +1,46 @@
...
+        }
+        android.util.Log.e(logTag, s + ":" + msg);
+        return msg;
+    }
+
+    public void i(String logTag, String s) {
+
+        android.util.Log.i(logTag, s);
+
```
garvankeeley: nit newlines

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/svclocator/services/log/ProductionLogger.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    public void d(String logTag, String s) {
+
+        return;
+
```
garvankeeley: nit newlines

libraries/stumbler/src/test/java/org/mozilla/mozstumbler/svclocator/services/DynamicProxyTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+        assertEquals("bob", t.echo("bob"));
+        assertEquals((byte) 1, t.getByte());
+        assertEquals((short)1, t.getShort());
+        assertEquals(1, t.getInt());
+        assertEquals(1, t.getLong());
+        assertTrue(0 == Float.compare((float)1, t.getFloat()));
+        assertTrue(0 == Double.compare(t.getDouble(), (double) 1));
+        assertEquals('a', t.getChar());        //assertNull(t.getServiceLocator());
+
```
garvankeeley: nit newline

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/svclocator/services/log/ProductionLogger.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
```
garvankeeley: nit newline

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/svclocator/services/log/DebugLogger.java
```diff
@@ -0,0 +1,52 @@
...
+        System.out.println(msg);
+        messageBuffer.addLast(msg);
+    }
+
+    public void d(String logTag, String s) {
+        String msg = "d: " + logTag + ", " + s;
+        System.out.println(msg);
+        messageBuffer.addLast(msg);
+    }
+
```
garvankeeley: nit newline

crankycoder:
@garvankeeley I've reformatted everything, optimized all the imports and included a new JAR file that contains the code formatting settings.  If something is off, we should update the JAR file with new settings in `android_studio/stumbler_codestyle.jar`.

android/src/main/java/org/mozilla/mozstumbler/service/utils/TelemetryWrapper.java
crankycoder, line 6: This formatting is what Android Studio does by default. I'm not sure how to change it, but I've included my code style settings as an importable JAR.

garvankeeley:
looking forward to trying to code style out!

lua86:
I don't think having an importable jar inside an open source repo can be a good idea. It is a blob and blobs are bad for maintainability. And probably this is not the most proper place to comment that though. But I needed to and I don't have much time to figure a better place.

crankycoder:
I'm not sure what the better option is.  The settings export from Android
Studio is a JAR file. Making yet-another-format would mean writing and
maintaining code to transform the JAR into a format that literally nobody
can use, then retransform it back into a format that Android Studio can use.

That seems suboptimal and not worth the effort.

lua86:
Well then I think the one to blame here for the wibbly-wobbly blob is Android Studio. So nothing to do here... (I suppose both modifying or dropping Android Studio are not options to consider so let us pretend I did not say anything about it).

PR #1437, opened 2015-01-27
Minor bugfixes

### LINKED ISSUE #1429: log lines are duplicated after putting the device to sleep and waking up it
If I have Log open, suspend the device and resume it, all the lines get reprinted. For example, there are 5 lines in here that got reprinted two times (because I suspended and resumed the device twice while looking at the log):

log-duplicated

You can easily see the duplication by the timestamps. If there are just a few lines, it's still readable, but once there's a lot of lines, it's really hard to navigate in the log with this bug occurring.

After I close the Log and open it again, it is clean again and there's no duplication:
log-clean

stumbler 1.6.1

cascheberg:
I hope these fixes are simple enough to put them into just one PR.

In `MainDrawerActivity`: set `hasBootedOnce` on first run, so that scanning does not restart unexpectedly

In `LogActivity`: fixes #1429 

In `ScanManager`: remove unused code, all those variables are always reassigned a few lines below


garvankeeley:
Looks good, thanks!

PR #1438, opened 2015-01-28
Fix and update strings

cascheberg:
- Update German strings
- Remove unused strings
- I noticed that Greek and Czech values were using the country code instead of the language code, which is not working


crankycoder:
Thanks!

PR #1443, opened 2015-01-29
1441 - NPE in Significant Motion Sensor Code

crankycoder:
This is a hot fix that may cause the signfiicant motion sensor to never register properly.

@garvankeeley - I'm writing a test case in a separate PR to make sure we can catch problems like this.


android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/MotionSensor.java
```diff
@@ -84,7 +84,11 @@ public void onTrigger(TriggerEvent event) {
                         return;
                     }
                     AppGlobals.guiLogInfo("Major motion detected.");
-                    LocalBroadcastManager.getInstance(mContext).sendBroadcastSync(new Intent(ACTION_USER_MOTION_DETECTED));
+                    LocalBroadcastManager broadcastManager = LocalBroadcastManager.getInstance(mAppContext);
+                    if (broadcastManager == null) {
```
garvankeeley: to be clear, this is the fix here, regardless of the context

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/MotionSensor.java
```diff
@@ -84,7 +84,11 @@ public void onTrigger(TriggerEvent event) {
                         return;
                     }
                     AppGlobals.guiLogInfo("Major motion detected.");
-                    LocalBroadcastManager.getInstance(mContext).sendBroadcastSync(new Intent(ACTION_USER_MOTION_DETECTED));
+                    LocalBroadcastManager broadcastManager = LocalBroadcastManager.getInstance(mAppContext);
```
garvankeeley: I want mAppContext null checked

android/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/MotionSensor.java
```diff
@@ -54,27 +54,27 @@ public static Sensor getSignificantMotionSensor(Context context) {
         return null;
     }
 
-    public MotionSensor(Context context, BroadcastReceiver callbackReceiver) {
+    public MotionSensor(Context ctx, BroadcastReceiver callbackReceiver) {
         sDebugInstance = this;
-        mContext = context;
+        mAppContext = ctx.getApplicationContext();
```
garvankeeley: null check here

garvankeeley:
I am going to be a PITA and r- this.
I'd like a fix with no other code changes, like renaming. I think we agree the crash is the async registration of the TriggerEventListener.
What is happening:
 mSensorManager.requestTriggerSensor(tr, mSignificantMotionSensor);
  app gets killed 
 tr gets called, it tries to operate on a bad context

Either we have the wrong context when we think we have an app context
or
the app is getting killed very reproducibly at that particular time, which to me is the actual bug

PR #1444, opened 2015-01-29
Czech l10n of most used strings (e.g. bug form ommited)

MikkCZ:


MikkCZ:
I've translated most of the strings in the file, bug report form and What's New strings have been ommited intentionally for now.

garvankeeley:
Thanks! This isn't automatically mergeable.
Hi can you rebase this patch on the latest code? 
Let me know if you need any help with this.

MikkCZ:
I have never rebased a pull request before, but tried to and seems to be ok now. If not, let me know and I will create a completely new request.

cascheberg:
@MikkCZ It might be better to remove the untranslated strings from the file, or to comment them out.
In that case the English strings will used anyway. That means that you do not need to update or delete those in case something changes, and it's easier to see (in Android Studio) which strings have not been translated yet.

MikkCZ:
Commented.

crankycoder:
Thanks @MikkCZ !

PR #1458, opened 2015-02-03
Cleaning up ScanManager stop/started/paused state

garvankeeley:
I find this code confusing (because the paused state was grafted on
without changing the ScanManager isScanning logic.
Now ScanManager has an enum for state (stop/started/started but paused).
I find this code more readable.

I would suggest starting by looking at the changes in ScanManager.java. Everything else was a product of this change:
https://github.com/mozilla/MozStumbler/commit/33574021b483324e17b7755e692c070a849d1e7e#diff-a0bf0529c9d81614f9223e2715512cb9R57

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/MainApp.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/MainApp.java
@@ -75,9 +75,11 @@ public class MainApp extends Application
     private ServiceBroadcastReceiver mReceiver;
     private WeakReference<IMainActivity> mMainActivity = new WeakReference<IMainActivity>(null);
     private boolean mIsScanningPausedDueToNoMotion;
-    private final BroadcastReceiver mReceivePausedState = new BroadcastReceiver() {
+    private final BroadcastReceiver mReceivePausedOrUnpausedState = new BroadcastReceiver() {
         public void onReceive(Context context, Intent intent) {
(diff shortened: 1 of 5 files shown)


android/src/main/java/org/mozilla/mozstumbler/client/MainApp.java
```diff
@@ -323,7 +326,7 @@ public boolean isScanningOrPaused() {
         if (mStumblerService == null) {
             return false;
         }
-        return mStumblerService.isScanning() || mIsScanningPausedDueToNoMotion;
+        return !mStumblerService.isStopped();
```
crankycoder, line 329: This is a lot nicer.

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/ScanManager.java
```diff
@@ -53,34 +52,41 @@ public void batteryCheckCallback(BatteryCheckReceiver receiver) {
     };
     private LocationChangeSensor mLocationChangeSensor;
     private MotionSensor mMotionSensor;
-    private boolean mIsMotionlessPausedState;
+
+    private enum ScannerState {
```
crankycoder: The Scanner has been finicky before with buggy hardware, so I'd like to be able to inspect the value of `mScannerState` when we put the ScanManager under test.

Can we make `ScannerState` and `mScannerState` to have package visibility?

crankycoder:
I have to apply this patch before we merge,  but it looks OK so far.

Issue #1457, opened 2015-02-03
Update ./test.sh to run tests in libraries


crankycoder:
There's no easy way to target individual tests in the libraries/stumbler code.

test.sh doesn't work and the ctrl-shift-t shortcut in Android Studio also don't work.

The only way to run tests in the libraries right now is to manually run:

`./gradlew test` from within libraries/stumbler.

This is dumb and will scare off contributors.

**Update:** This is actually worse than I thought.  `make test` doesn't recursively run tests on library code.  Blargh.

crankycoder:
Make test now runs tests recursively on the libraries code,  but the shell script still won't target individual tests.  

The groovy hack is still broken.  It won't run individual tests in libstumbler.

PR #1459, opened 2015-02-03
`make test` now runs tests on the libraries/stumbler package

crankycoder:
Short term fix to get all test code to run until we pull out the services package as it's own repo.

This addresses #1457


crankycoder:
@garvankeeley you'll be happy to note that I missed a robolectric annotation in libraries/stumbler for the NetworkInfoTest that this patch caught.

PR #1463, opened 2015-02-06
developer setting for unlimited map zooming

garvankeeley:


android/src/main/res/values/strings.xml
```diff
@@ -195,4 +195,5 @@
 \n<li>New testing infrastructure for developers</li>
 \n<li>Simulation mode added to the developer options.</li>
 </string>
+    <string name="limit_map_zoom">Limit map zoom</string>
```
crankycoder: Nit: "Limit zoom out" would be clearer.

crankycoder:
Other than the wording on the string, this works for me and is fine.  r+

android/src/main/res/values/strings.xml
garvankeeley: "Limit zooming out on map" ?

garvankeeley: Let's try that and see how it goes. I think this will get moved to Settings if it is liked.

PR #1464, opened 2015-02-07
Add more Updater tests

cascheberg:
Add tests to check if the Updater is handling offline and wifi-only cases correctly.

@crankycoder I had to copy `MyShadowConnectivityManager`, can/should that be improved? (Or did I miss the correct solution?)

Also removed some leftovers of the SSID blocklist removal.

### DIFF
--- a/android/src/test/java/org/mozilla/mozstumbler/client/UpdaterTest.java
+++ b/android/src/test/java/org/mozilla/mozstumbler/client/UpdaterTest.java
@@ -4,6 +4,7 @@
 package org.mozilla.mozstumbler.client;
 
 import android.content.Context;
+import android.net.ConnectivityManager;
 
 import org.junit.Before;
 import org.junit.Test;
@@ -31,10 +32,12 @@ import static org.mockito.Mockito.spy;
 @RunWith(RobolectricTestRunner.class)
 public class UpdaterTest {
 
+    private Context ctx;
     private MainDrawerActivity activity;
 
     @Before
     public void setUp() throws Exception {
+        ctx = Robolectric.application;
         activity = Robolectric.newInstanceOf(MainDrawerActivity.class);
         Updater.sLastUpdateCheck = 0;
     }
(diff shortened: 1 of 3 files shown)


android/src/test/java/org/mozilla/mozstumbler/client/UpdaterTest.java
```diff
@@ -59,16 +62,14 @@ public void testUpdater() {
         ServiceLocator.getInstance().putService(IHttpUtil.class, mockHttp);
         ServiceLocator.getInstance().putService(ISystemClock.class, clock);
 
-        // wifi is always unavailable
+        // network is always available
```
crankycoder: Re-reading this comment, I don't like either comment now.  

Let's make it really clear with something like "Skip the exclusive wifi-only check".

android/src/test/java/org/mozilla/mozstumbler/client/UpdaterTest.java
```diff
@@ -59,16 +62,14 @@ public void testUpdater() {
...
         doReturn(false).when(upd).wifiExclusiveAndUnavailable(Mockito.any(Context.class));
 
         assertFalse(upd.checkForUpdates(activity, ""));
-        now += Updater.UPDATE_CHECK_FREQ_MS;
```
crankycoder, line 66: I generally like this test, but let's split it up.   The `testUpdater()` method is meant to just test the parsing of the version numbers.  

Let's have one test for the version number parsing, and a separate one to verify that a `checkForUpdates` call that returns false does not update the sLastUpdate timestamp.

android/src/test/java/org/mozilla/mozstumbler/client/UpdaterTest.java
```diff
@@ -90,9 +92,11 @@ public void testUpdaterThrottlesRequests() {
         ServiceLocator.getInstance().putService(IHttpUtil.class, mockHttp);
         ServiceLocator.getInstance().putService(ISystemClock.class, clock);
 
-        // wifi is always unavailable
-        doReturn(false).when(upd).wifiExclusiveAndUnavailable(Mockito.any(Context.class));
+        ConnectivityManager connectivityManager = (ConnectivityManager) ctx.getSystemService(Context.CONNECTIVITY_SERVICE);
+        MyShadowConnectivityManager shadowConnManager = (MyShadowConnectivityManager) Robolectric.shadowOf(connectivityManager);
 
+        shadowConnManager.setConnectedFlag(true);
+        doReturn(false).when(upd).wifiExclusiveAndUnavailable(Mockito.any(Context.class));
```
crankycoder, line 93: It's not clear to me what these changes are adding with respect to testing.  Can we add some extra assertions that demonstrate why the extra shadow is required?

crankycoder:
This looks awesome.  Some minor comments for changes - but otherwise - this looks great.

I'm not that concerned about having multiple implementations of shadows.  I do think we need a better naming convention so that we can easily identify which testcase a particular shadow is being used.  Any suggestions there are more than welcome.

Merging all the shadows together seems like it would make things more complicated than necessary when a new contributer tries to read and understand an existing test case.

android/src/test/java/org/mozilla/mozstumbler/client/UpdaterTest.java
cascheberg, line 66: I moved the `sLastUpdateCheck` assertions to `testUpdaterThrottlesRequests`.
Also, if the last call in this test now successfully returns true, then I think one can assume that the previous calls did not have any side effects.

android/src/test/java/org/mozilla/mozstumbler/client/UpdaterTest.java
cascheberg, line 93: You are right, this does not seem to be necessary.
When I ran my modified tests this failed because there was no network connection, so I added `shadowConnManager.setConnectedFlag(true)`. But I could not reproduce this problem, so I reverted these changes now.

cascheberg:
@crankycoder PR updated, and I replied to your review comments.

crankycoder:
Thanks @cascheberg !  More tests are awesome.  Woot.

_tmp_280sw007883

PR #1472, opened 2015-02-10
Motion detection cleanup + remote unreliable linear. accel

garvankeeley:
Some minor cleanup items and remove the TYPE_LINEAR_ACCELEROMETER and use TYPE_ACCELEROMETER only. The linear. accel doesn't work reliably.


crankycoder:
Just waiting on Travis to come back with a green, but this looks good. r+

Issue #1428, opened 2015-01-22
Extract an embeddable library


crankycoder:
Since this is actually one of the 2015Q1 goals, we should probably have a proper bug filed.

Reorganize the stumbler so that we have a proper embeddable android library that can be used by other android applications.

PR #1473, opened 2015-02-11
More code for #1428 - Stumbler Service as a library

crankycoder:
This adds tests around the 3 intents inititated by Android at boot, GPS enabled change and GPS locate change as well as the start and stop intent.

Static methods have been added to PassiveServiceReceiver to make creation of the start/stop intents more explicit.


garvankeeley:
Other than my questions, this is very, very cool!

PR #1474, opened 2015-02-11
move unlimited zoom to settings screen from developer screen

garvankeeley:
I think this is a feature that regular users would like to have.


garvankeeley:
There is one bug with this, which I'll file, the zoom buttons on the map view don't change to reflect that more zooming out is available until the user zooms in (or pinch zooms in or out)

Issue #1322, opened 2014-12-03
Critical: Galaxy S5 motion detection not working, scanning never starts


garvankeeley:
Report on play store that this isn't working, after pausing scanning, the app fails to detect motion to start scanning again.

Not sure if it is a custom ROM or not. Would be good to have the log activity checked on startup to see if device has significant motion sensor or not. Also, what steps were taken to try wake the scanning?
(Chris hopefully you see this and can fill in the questions).

crankycoder:
On a related note - we should setup a google doc or an etherpad page with a list of all contributors and testers with the devices+ROM that they have on hand for testing.

garvankeeley:
We have to get to the bottom of this, I am hearing anecdotal reports from users that stumbling is "difficult" to start after pausing.
@cpeterso : did you confirm having problems on your Cyanogenmod 4.3 Nexus 4

simonpoole:
I can confirm the issue. Essentially even though no other apps have problems tracking the position at the same time (including such for which I'm the main dev) since version 1.5.2 mozstumbler seems to be essentially not working. Prior versions didn't exhibit this behaviour. 

While pausing seems to increase the issues, even freshly started the behaviour is erractic.

garvankeeley:
@simonpoole can you try 1.5.3 
https://github.com/mozilla/MozStumbler/releases/tag/v1.5.3

simonpoole:
@garvankeeley sorry, will test tomorrow,

simonpoole:
OK, the good news (1.5.3): behaviour is better in that there no unmotivated pauses and resulting hanging outside with good GPS reception, however it doesn't actually seem to pause at all even when clearly at rest. 

Bad news (repeatable): moving indoors with loss of GPS reception will, after a while lead to the app pausing recording and hanging (clear sign of this: GPS sat display in MozStumbler freezes and doesn't reflect actual available sats). 

 For completness sake, this is on a Galaxy S5 (G900F) with 4.4.2.

garvankeeley:
Any chance you can screenshot the Log screen when this happens?

My guess at the bug:
- with no GPS, then the app goes into 'paused' state.
- then if you keep moving at a constant rate, we don't get notification of a significant change in motion.

simonpoole:
![screenshot_2014-12-11-12-26-43](https://cloud.githubusercontent.com/assets/1011860/5393405/ccff1d00-8131-11e4-8d2d-bb74a7c6ad30.png)

[...]

PR #1478, opened 2015-02-11
Improvements for accelerometer motion sensing (filter false pos. and min sleep time)

garvankeeley:


crankycoder:
Let's get this merged in for now.  It seems to be ok on both my Nexus 4 and OnePlus One.  We can get the code cleaned up, tests added and documentation added in subsequent patches - but in light of #1322 - I think it's more important to get something out to users to test right away.

PR #1480, opened 2015-02-12
Fix the reset logic on the FalsePositiveFilter.

garvankeeley:
The last GPS location should stay unchanged until the user moves beyond the threshold distance or scanning is stopped+started (as opposed to when paused+started).

The FalsePositiveFilter gets initialized with a GPS location when the user is motionless, and this location should not get updated until the filter threshold is exceeded, or scanning is manually stopped+started.

### DIFF
--- a/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/MotionSensor.java
+++ b/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/MotionSensor.java
@@ -22,7 +22,8 @@ public class MotionSensor {
     public static final String ACTION_USER_MOTION_DETECTED = AppGlobals.ACTION_NAMESPACE + ".USER_MOVE";
     private static final String LOG_TAG = LoggerUtil.makeLogTag(MotionSensor.class);
     private static final ILogger Log = (ILogger) ServiceLocator.getInstance().getService(ILogger.class);
-    static final long GPS_WARM_TIME_MS = 2 * 60 * 60 * 1000; // after this, the GPS is considered cold
+    static final long TIME_UNTIL_GPS_GOES_COLD = 2 * 60 * 60 * 1000; // 2 hours
+
     /// Testing code
     private final SensorManager mSensorManager;
     private final Context mAppContext;
@@ -124,7 +125,6 @@ public class MotionSensor {
 
         mMotionSensor.stop();
 
-        final long TIME_UNTIL_GPS_GOES_COLD = 2 * 60 * 60 * 1000; // 2 hours
         if (mFalsePositiveFilter.timeSinceInitialGPS() < TIME_UNTIL_GPS_GOES_COLD) {
             mFalsePositiveFilter.start();
         }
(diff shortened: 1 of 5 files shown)


libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/MotionSensor.java
```diff
@@ -118,12 +119,18 @@ public void stop() {
     // If the GPS has gone cold (~2 hours), then stop using this filter.
     FalsePositiveFilter mFalsePositiveFilter;
     private static class FalsePositiveFilter {
-        final Context mContext;
-        final Handler handler = new Handler();
+        private final Context mContext;
+        private final Handler handler = new Handler();
```
crankycoder: If we're adding 'm' prefixes, we should change handler to mHandler as well.

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/MotionSensor.java
```diff
@@ -118,12 +119,18 @@ public void stop() {
...
         private Location mLastLocation;
-        private final long mMinMotionChangeDistanceMeters = 30;
+        private final long mMinMotionChangeDistanceMeters = 10;
         private long mTimeStartedMs;
 
+        private final Runnable mRunnable = new Runnable() {
```
garvankeeley: this should get a better name, mStopListeningAfterTimeout or something

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/MotionSensor.java
```diff
@@ -65,6 +65,7 @@ private void setTypeFromPrefs() {
     // This class can use this event to update its state
     public void scannerFullyStopped() {
         setTypeFromPrefs();
+        mFalsePositiveFilter.reset();
```
crankycoder, line 69: I'd like a test that exercises the motion detector state to clearly see what scenarios we can expect `reset()` vs `setTime()` to be invoked.

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/MotionSensor.java
```diff
@@ -167,27 +181,28 @@ void start() {
...
-                    stop();
-                }
-            }, 30 * 1000);
+            handler.postDelayed(mRunnable, 30 * 1000);
         }
 
         void stop() {
             LocationManager lm = (LocationManager)mContext.getSystemService(Context.LOCATION_SERVICE);
             lm.removeUpdates(mListener);
-            handler.removeCallbacks(null);
```
crankycoder, line 180: Oof.  I'm surprised `removeCallbacks(null)` didn't crash.

crankycoder:
Review is done, I just want a test case.  Testsssssss

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/MotionSensor.java
garvankeeley, line 180: that is legal. It removes all callbacks on the handler, but that was from WIP patch, I would avoid ever doing that in proper code

PR #1482, opened 2015-02-13
Check if offline before querying MLS

cascheberg:
Querying MLS does not check if any network is available. This can result in lots of `UnknownHostException` being displayed in the log.

STR:
- enable querying MLS
- disable wifi only
- be offline
- stumble and check GUI log


garvankeeley:
Thanks!

PR #1486, opened 2015-02-13
Features/1428 libstumbler patches

crankycoder:
This does a couple cleanups of the codebase.

`ILocationService` is now registered on application startup. So no more `new MLS()` invocations all over the place.

Logger instances are setup in more classes now.

The old mess of multiple DebugLogger source files has been fixed with reflection.

Some extra logging has been added to AsyncUploader as it was difficult to verify that uploads were happening in debug builds.


libraries/stumbler/src/main/java/org/mozilla/mozstumbler/svclocator/services/log/DebugLogger.java
```diff
@@ -14,8 +15,21 @@
  */
 public class DebugLogger implements ILogger {
 
+    private static final String LOG_TAG = LoggerUtil.makeLogTag(DebugLogger.class);
+
     static {
-        ShadowLog.stream = System.out;
+        try {
+            @SuppressWarnings("all")
```
crankycoder, line 22: This is really ugly and confusing, but it works.

crankycoder, line 22: That's a comment on ugly reflection.  Our ProductionLogger no-ops all debug level logging.

PR #1494, opened 2015-02-19
Turkish localisation of Mozilla Stumbler

asteko:
First Turkish localisation of the extension, excluding "What's New?" part.


crankycoder:
Thanks @asteko!  The strings.xml file is missing a '\' escape character that needs to be put in front of all single quote characters.

I've sent you a pull request that you can try to merge that makes the correction over here: https://github.com/asteko/MozStumbler/pull/1

PR #1498, opened 2015-02-23
Performance-safe stats reporting

garvankeeley:
The synchronous stats reporting is removed. There were 2 performance problems in the code. 

1) Problem: to show the metrics of how much data is in the queue, requires zipping the in-memory contents to get the correct size, which is very slow.
Solution: Remove the calc of the zipped size of in-memory stumbles from DataStorageManager, and put it in a new class QueuedCountsTracker, that has a rate-limited (every 2sec) max freq of updating. Alternatively, I could have had this update when new reports come in (i.e. new report arrives, recalculate the in-memory zip size) and broadcast the size to listeners. I wanted this decoupled from the report frequency, which is something that might change. Also, this would update far more frequently than necessary, as the metrics only needs to update when it is visible. (Note also, that this part of the code is client-side only, it is very specific to the Mozilla Stumbler use case, and I don't want the service to "know" about this use case).

2) Problem: the stats for previous uploaded data is stored as a small text file, and reading this from disk is slow. Solution: Change the reporting of uploaded stats from an on-demand query, to a broadcast listener, so this info is sent out only when it changes. (This code _is_ part of the service, and makes sense to have it broadcast, because saving to disk is part of the regular data storage flow, unlike case 1 which is an optional query).

### DIFF
--- a/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/datahandling/DataStorageManager.java
+++ b/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/datahandling/DataStorageManager.java
@@ -86,7 +86,7 @@ public class DataStorageManager {
         mFileList = new ReportFileList();
         mFileList.update(mReportsDir);
 
-        mPersistedOnDiskUploadStats = new PersistedStats(baseDir);
+        mPersistedOnDiskUploadStats = new PersistedStats(baseDir, c);
     }
 
     public static String getStorageDir(Context c) {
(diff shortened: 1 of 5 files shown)


garvankeeley:
To test this I kept the old code in place, added all the new code under a different package, made calls to both and assert()ed the results were equal.

crankycoder:
DataStorageManager can be removed from https://github.com/garvankeeley/MozStumbler/blob/safer-stats-reporting/android/src/main/java/org/mozilla/mozstumbler/client/navdrawer/MetricsView.java#L216 to L219 which completely removes the dependency on DataStorageManager.

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/datahandling/DataStorageManager.java
```diff
@@ -376,31 +384,15 @@ private void notifyStorageIsEmpty(boolean isEmpty) {
...
     public synchronized void incrementSyncStats(long bytesSent, long reports, long cells, long wifis) throws IOException {
         mPersistedOnDiskUploadStats.incrementSyncStats(bytesSent, reports, cells, wifis);
+
     }
 
     public interface StorageIsEmptyTracker {
         public void notifyStorageStateEmpty(boolean isEmpty);
     }
 
-    public static class QueuedCounts {
```
crankycoder, line 391: Yay. Fewer nested classes!

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/datahandling/PersistedStats.java
```diff
@@ -13,11 +19,15 @@
...
+public class PersistedStats {
     private final File mStatsFile;
+    private final Context mContext;
 
-    public PersistedStats(String baseDir) {
+    public static final String ACTION_PERSISTENT_SYNC_STATUS_UPDATED = AppGlobals.ACTION_NAMESPACE + ".PERSISTENT_SYNC_STATUS_UPDATED";
+    public static final String EXTRAS_PERSISTENT_SYNC_STATUS_UPDATED = ACTION_PERSISTENT_SYNC_STATUS_UPDATED + ".EXTRA";
+    public PersistedStats(String baseDir, Context context) {
         mStatsFile = new File(baseDir, "upload_stats.ini");
+        mContext = context;
```
crankycoder: You should probably make sure this is an instance of `ApplicationContext`.

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/datahandling/PersistedStats.java
```diff
@@ -3,6 +3,12 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 package org.mozilla.mozstumbler.service.stumblerthread.datahandling;
 
+import android.content.Context;
+import android.content.Intent;
+import android.os.Bundle;
+import android.support.v4.content.LocalBroadcastManager;
+import android.widget.Button;
```
crankycoder: `android.widget.Button` is unused

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/datahandling/PersistedStats.java
```diff
@@ -65,9 +75,15 @@ public synchronized void incrementSyncStats(long bytesSent, long reports, long c
                 observationsToday);
 
 
-        final long lastUploadMs = Long.parseLong(properties.getProperty(DataStorageContract.Stats.KEY_LAST_UPLOAD_TIME, "0"));
-        final int timeDiffSec = Long.valueOf((time - lastUploadMs) / 1000).intValue();
-        if (lastUploadMs > 0 && timeDiffSec > 0) {
+        Intent intent = new Intent();
```
crankycoder, line 80: I think we should we check that newProps has at least some valid attributes before firing it off to the broadcast service. There's no sense in sending message that can't be parsed - and will cause all receivers to probably reset their write counters to 0.

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/datahandling/ReportBatchBuilder.java
```diff
@@ -13,42 +13,39 @@
     // Once this size is reached, data is persisted to disk, mCurrentReports is cleared.
     public static final int MAX_REPORTS_IN_MEMORY = 50;
     private static final String LOG_TAG = LoggerUtil.makeLogTag(ReportBatchBuilder.class);
-    private final ArrayList<String> reports = new ArrayList<String>();
```
crankycoder, line 16: Woot.  This ought to kill all those `finalizeReports()` memory crashes we see on lower end devices like the Nexus S.

crankycoder, line 16: Oh, `ArrayList` is no longer required.

garvankeeley:
thanks updating now

Issue #1490, opened 2015-02-17
Accelerometer Motion sensor gaps in scanning


garvankeeley:
Been testing the latest accelerometer-based power savings on the S5, and results are very good overall. 
I ran into one case where the scanning stayed paused, not sure why just yet. The blue arrow is where the scanning stopped at a streetlight and did not restart. One possibility is a GPS problem.
![screenshot 2015-02-15 12 18 22](https://cloud.githubusercontent.com/assets/5495083/6230679/80aa4d88-b68f-11e4-8b70-ba16adbeed4e.png)

garvankeeley:
For comparison, this is a correct gap in the data, where I was driving and stopped for 2 mins:
![screenshot 2015-02-15 12 15 28](https://cloud.githubusercontent.com/assets/5495083/6230712/c5eed7f6-b68f-11e4-96ac-15201e283cd0.png)

crankycoder:
I've seen the same behavior on the Nexus 4 and Moto G devices while riding the bus back to Newmarket.  I'm in agreement that this is probably some problem with losing the GPS fix.

garvankeeley:
Here is a deceleration that should have triggered the sensor and did not:

![screenshot 2015-02-20 12 05 33](https://cloud.githubusercontent.com/assets/5495083/6305378/d64e3e00-b8f8-11e4-9eeb-7c800c29e7ad.png)

garvankeeley:
I am no longer getting this behaviour after #1502 landed.

PR #1502, opened 2015-02-24
Legacy sensor: use a rolling average to detect motion

garvankeeley:
Use the rolling average at t1  as a baseline, and when the average exceeds
a theshold relative to the baseline average at t1, then notify movement has
occurred, and update the baseline.
Sounds fancy but it isn't :).

### DIFF
--- a/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/LegacyMotionSensor.java
+++ b/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/LegacyMotionSensor.java
@@ -10,7 +10,6 @@ import android.hardware.SensorEvent;
 import android.hardware.SensorEventListener;
 import android.hardware.SensorManager;
 import android.support.v4.content.LocalBroadcastManager;
-import android.util.FloatMath;
 
 import org.mozilla.mozstumbler.svclocator.services.log.LoggerUtil;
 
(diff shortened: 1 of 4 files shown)


libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/LegacyMotionSensor.java
```diff
@@ -48,57 +39,65 @@ public LegacyMotionSensor(Context ctx) {
...
+        public RollingAverage(int sampleCount, float changeThreshold) {
+            mSamples = new float[sampleCount];
+            mThreshold = changeThreshold;
         }
 
-        float x = linear_acceleration[0];
-        float y = linear_acceleration[1];
-        float z = linear_acceleration[2];
+        public boolean isOverThreshold(float val) {
+            mTotal -= mSamples[mIdx];
```
crankycoder, line 62: Shouldn't there be some control to make sure each sample is taken over a regular time window?  It seems like this could be skewed if the accelerometers don't regularly send out metrics - which might actually be the case now that I'm typing this.

crankycoder, line 62: Ah, nevermind. I forgot that registering in `mSensorManager.registerListener` sets the sampling rate.

garvankeeley, line 62: We currently use SENSOR_DELAY_UI freq, which is 15 Hz. This produces 3 average values every 2 seconds. This could vary widely (say 1 average value a second) and still produce ok results, because of the sensitivity of the threshold (only a 0.5 change in average).

crankycoder:
There's no point in adding a test case here yet - we keep having weird device specific problems.  So let's just get this into people's hands to test it out and see if this really works.  We can formalize the behavior in a test case after we see that this actually does seem to work on lots of low end devices.

garvankeeley:
this fix addresses https://github.com/mozilla/MozStumbler/issues/1490

Issue #1496, opened 2015-02-20
Significant motion sensor doesn't wake up when current location dot indicates movement


crankycoder:
The significant motion sensor doesn't seem to always be hooked into the GPS location.

I think we need to do some work here to capture a delta between the last 2 known GPS locations and force a wakeup.  Here - I'm on a train that is accelerating slowly.  The phone clearly knows that I"m moving - the map centering appears to be correct, but you can see the overlay that indicates that the stumbling has been put into a pause state.

![2015-02-20 07 10 40](https://cloud.githubusercontent.com/assets/4829/6308667/2b60f73c-b914-11e4-8655-350633353a17.png)
![2015-02-20 07 13 23](https://cloud.githubusercontent.com/assets/4829/6308671/3a28d8b6-b914-11e4-9bda-8b298b333118.png)

My accelerometer indicated that I was accelerating @ 1.7m/s^2 on the train - not sure off hand if that's just below the threshold of waking up or not.

This was observed on my OnePlus One phone.   Screen was on (obviously).

garvankeeley:
I would have thought the wifi-based location changes would have triggered the sig. motion sensor. I assumed that was one of the things it was watching as a trigger.

garvankeeley:
Updated the title to reflect that the current position dot may be updating based on wifi positioning.

The MapLocationListener.java has both a GPS and a network location listener. It is likely the latter that is changing. And yes we should have that trigger that motion is occurring.

crankycoder:
dmesg shows

```
<6>[360665.764745] wlan: WCNSS software version MSM8974-PR_1_4_c2-00125
<6>[360665.764791] wlan: WCNSS hardware version WCN v2.0 RadioPhy vIris_TSMC_3.0 with 48MHz XO
```

which looks like a Motorola chip.

The accelerometers are on the QC chipset, so it seems pretty reasonable that the significant motion sensor isn't seeing the wifi location.

crankycoder:
Hrm.. the snapdragon 801 should have wifi on board.

https://www.qualcomm.com/products/snapdragon/processors/801

but who knows. Maybe the wifi location lookups aren't fed back into the chipset.

PR #1509, opened 2015-02-26
detect network location change as part of motion sensing

garvankeeley:
Addresses https://github.com/mozilla/MozStumbler/issues/1496

### DIFF
--- a/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/MotionSensor.java
+++ b/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/MotionSensor.java
@@ -23,7 +23,7 @@ public class MotionSensor {
     private static final String LOG_TAG = LoggerUtil.makeLogTag(MotionSensor.class);
     private static final ILogger Log = (ILogger) ServiceLocator.getInstance().getService(ILogger.class);
     static final long TIME_UNTIL_GPS_GOES_COLD = 2 * 60 * 60 * 1000; // 2 hours
-
+    private final NetworkLocationChangeDetector mNetworkLocationChangeDetector = new NetworkLocationChangeDetector();
     /// Testing code
     private final SensorManager mSensorManager;
     private final Context mAppContext;
@@ -113,6 +113,8 @@ public class MotionSensor {
         mHandler.postDelayed(mStartMotionSensorRunnable, delayMs);
 
         mFalsePositiveFilter.stop();
+
+        mNetworkLocationChangeDetector.start(mAppContext);
     }
 
     public void stop() {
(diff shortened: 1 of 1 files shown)


libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/MotionSensor.java
```diff
@@ -128,8 +130,54 @@ public void stop() {
...
+        public void start(Context c) {
+            mContext = c;
+            LocationManager lm = (LocationManager) mContext.getSystemService(Context.LOCATION_SERVICE);
+
+            if (mLastLocation == null) {
+                mLastLocation = lm.getLastKnownLocation(LocationManager.GPS_PROVIDER);
+            }
+
+            if (mLastLocation != null) {
+                lm.requestLocationUpdates(LocationManager.NETWORK_PROVIDER, 0, 0, mNetworkLocationListener);
```
crankycoder: Why is this null checking before requesting locations?  I would think it's ok to request new locations even if we didn't have a last known location in the GPS provider.

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/MotionSensor.java
```diff
@@ -128,8 +130,54 @@ public void stop() {
...
+
+    static class NetworkLocationChangeDetector {
+        Location mLastLocation;
+        Context mContext;
+        private float DIST_THRESHOLD_M = 30.0f;
+
+        private LocationListener mNetworkLocationListener = new LocationListener() {
+            public void onLocationChanged(Location location) {
+                if (mLastLocation == null) {
+                    return;
```
crankycoder: Shouldn't this do `mLastLocation = location` before returning?

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/motiondetection/MotionSensor.java
```diff
@@ -128,8 +130,54 @@ public void stop() {
...
+            public void onProviderEnabled(String provider) {}
+            public void onProviderDisabled(String provider) {}
+        };
+
+        public void start(Context c) {
+            mContext = c;
+            LocationManager lm = (LocationManager) mContext.getSystemService(Context.LOCATION_SERVICE);
+
+            if (mLastLocation == null) {
+                mLastLocation = lm.getLastKnownLocation(LocationManager.GPS_PROVIDER);
```
crankycoder, line 166: Maybe change this to get whatever the highest accuracy location is by doing something like:

``` java
matchingProviders = locationManager.getAllProviders();
for (String provider: matchingProviders) {
   location = locationManager.getLastKnownLocation(provider);

   // check that location is not null and grab the location with the best accuracy 
   // here, then bind it to mLastLocation
}
```

crankycoder:
I've completed the review.  Mostly - I think we should try to just grab any location at first.

garvankeeley:
Note: I did not change the false positive filter to allow for late initialization of its reference location (mLastLocation), that is by-design required to have an initial GPS location and then wait to see if later locations are the same location. Without the initial GPS location, the class won't attempt to filter a false positive, which is correct behaviour.

Issue #1520, opened 2015-02-27
ConcurrentModification error in v1.6.1


crankycoder:
```
Caused by: java.util.ConcurrentModificationException
at java.util.LinkedList$LinkIterator.next(LinkedList.java)
at org.mozilla.mozstumbler.client.serialize.ObservationPointSerializer.writeOut(ObservationPointSerializer.java:130)
at org.mozilla.mozstumbler.client.serialize.ObservationPointSerializer.doInBackground(ObservationPointSerializer.java:309)
at org.mozilla.mozstumbler.client.serialize.ObservationPointSerializer.doInBackground(ObservationPointSerializer.java:43)
at android.os.AsyncTask$2.call(AsyncTask.java)
```

crankycoder:
Fixed in #1522

PR #1522, opened 2015-02-27
Fix concurrency for #1520

crankycoder:


garvankeeley:
The problem starts here:
https://github.com/mozilla/MozStumbler/blob/dev/android/src/main/java/org/mozilla/mozstumbler/client/serialize/KMLFragment.java#L55

The point list is passed in, a reference to this list is kept. It is then passed in to the ObservationPointSerializer
https://github.com/mozilla/MozStumbler/blob/dev/android/src/main/java/org/mozilla/mozstumbler/client/serialize/KMLFragment.java#L161

Again this list is not copied. If a copy of the list was made, either around line 55 or inside the constructor called at 161, this code would be safe.

crankycoder:
I've changed the accessor at `ObservedLocationsReceiver.getInstance().getObservationPoints()` to provide a copy all the time. Synchronization is also tweaked a bit so that the list is always safely copied.

Issue #1524, opened 2015-03-01
Metrics not updated on app start


cascheberg:
The metrics in the navdrawer are not set when the app starts (i.e. last upload time: never). The stats are updated for the first time when a report is finished. This is a recent regression.

PR #1528, opened 2015-03-02
(1524) Force sync stats broadcast on startup so the metrics view has initial values

garvankeeley:
When the ClientStumblerService is initialized, force a single read of the sync stats, and broadcast this for the metrics view to update on startup.
Note that in the app the UI components are initialized before the service.

https://github.com/mozilla/MozStumbler/issues/1524


crankycoder:
Minor nit I just noticed that `PersistedStats::incrementSyncStats` has a call to System.currentTimeMillis, that should be changed to ISystemClock.

crankycoder:
review is done.

crankycoder:
Hmm. Hold up actually.  I just whipped up a small testcase and exposed a minor bug.

crankycoder:
@garvankeeley I sent a PR to your branch.

garvankeeley:
looking at it now

garvankeeley:
Thanks for the test case @crankycoder! Much gooder now.

PR #1535, opened 2015-03-03
changing known model list to include A0001 for oneplus

garvankeeley:


garvankeeley:
Fixing my previous commit. I thought @crankycoder had an Android One, he has a one plus, which has a working sig. motion sensor.

PR #1541, opened 2015-03-07
Fix notification updates

### LINKED ISSUE #1484: status bar is only updated on intervals -> wrong numbers on start
if there hasn't happened an upload the last time the app was open then the numbers in the status bar are wrong for some time
In this screenshot there aren't too many, but the status bar should use the actual numbers on app start and then start updating them periodically, I think (faster wasn't possible I think due to interface limitations)

cascheberg:
Fixes #1484 
Now sets the last upload time in the notification on start.
The metrics are always zero on app start, and (of course) it still takes one minute until they are updated (on devices < Android 5).


garvankeeley:
Looks good, thanks!

crankycoder:
Please add a test case as I'm refactoring all of this right now.  I'll guarantee a regression otherwise.

Djfe:
why is lollipop working better than kitkat and lower?

cascheberg:
@Djfe This is just because on older Android versions notification updates are animated (notifications are removed from the list and added back, or become minimized). This makes it impossible to read the notification if updates are happening too frequently.

Djfe:
ok makes sense
good that Google changed it then :)

Issue #1371, opened 2014-12-28
MapLocationListener is too slow


cascheberg:
If you are traveling fast, the location updates of the `MapLocationListener` are slower than those of the `GPSScanner`. As a result, new reports are drawn on the map without the blue user position dot being updated. So it is lagging behind which looks odd.
I think an easy fix would be to make the `MapLocationListener` listen to `GPSScanner.ACTION_GPS_UPDATED` intents.

![stumbler](https://cloud.githubusercontent.com/assets/3845150/5578099/8d21f11e-9020-11e4-92ed-f0abbe02ff1d.png)

Djfe:
saw this happening as well still

why are the updates slower btw.? are there some calculations being done at the same time or is the interval hardcoded?

PR #1542, opened 2015-03-07
NetworkLocationListener PR that is now superseded by 1588 and 1589

cascheberg:
Right now the `MapLocationListener` is activating the network location listener when scanning is paused. Does that make any sense? These listeners are only for updating the position marker on the map, not for motion detection, right?
And as far as I know, the network location provider will scan for cells and wifis and use the internet, which also consumes energy.

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapLocationListener.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapLocationListener.java
@@ -25,7 +25,7 @@ class MapLocationListener {
     private static final String LOG_TAG = LoggerUtil.makeLogTag(MapLocationListener.class);
     private final WeakReference<MapFragment> mMapActivity;
     private final LocationUpdateListener mNetworkLocationListener =
-            new LocationUpdateListener(LocationManager.NETWORK_PROVIDER, 1000, null);
+            new LocationUpdateListener(LocationManager.NETWORK_PROVIDER, 5000, null);
     private final ReceivedLocationCallback mReceivedGpsLocation = new ReceivedLocationCallback() {
         @Override
         public void receivedLocation() {
@@ -35,6 +35,8 @@ class MapLocationListener {
     };
     private final LocationUpdateListener mGpsLocationListener =
             new LocationUpdateListener(LocationManager.GPS_PROVIDER, 5000, mReceivedGpsLocation);
+    private final LocationUpdateListener mPassiveLocationListener =
+            new LocationUpdateListener(LocationManager.PASSIVE_PROVIDER, 1000, null);
     private LocationManager mLocationManager;
     private final GpsStatus.Listener mSatelliteListener = new GpsStatus.Listener() {
         public void onGpsStatusChanged(int event) {
@@ -70,6 +72,7 @@ class MapLocationListener {
             return;
         }
 
+        // the GPS status listener is invoked if any GPS location listener is active
         mLocationManager.addGpsStatusListener(mSatelliteListener);
 
         Location lastLoc = mLocationManager.getLastKnownLocation(LocationManager.GPS_PROVIDER);
@@ -85,12 +88,8 @@ class MapLocationListener {
         if (app == null) {
             return;
         }
-        if (app.isIsScanningPausedDueToNoMotion()) {
-            pauseGpsUpdates(true);
-        } else {
-            enableLocationListener(true, mNetworkLocationListener);
-            enableLocationListener(true, mGpsLocationListener);
-        }
+        enableActiveLocationUpdates(!app.isIsScanningPausedDueToNoMotion());
+        enableLocationListener(true, mPassiveLocationListener);
     }
 
     private void enableLocationListener(boolean isEnabled, LocationUpdateListener listener) {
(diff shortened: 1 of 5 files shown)


garvankeeley:
I don't have time to look at the patch yet, but to answer your questions:

"Right now the MapLocationListener is activating the network location listener when scanning is paused. Does that make any sense?"
Is is not enabled when paused or stopped? Both of those are no-GPS states. The intention is to continue to show location updates when the GPS is not active.
Wouldn't the map have to be showing for this to be updating?
Do you have any info on the power consumption of the network provider?

cascheberg:
Currently, if the scanning is stopped, then GPS is used. I think this is actually good, since I often have the mobile internet and usually also the network provider both disabled. And it is not a big problem, since (as you said) location requests are active only while the map is shown.

But only if scanning is paused, then only the network provider is used. I'd say if we are sure that the device is not moving, then we do not need to update the location on the map at all. So I removed this in my patch.

Some excerpt from the [API reference](http://developer.android.com/reference/android/location/LocationManager.html#requestLocationUpdates%28long, float, android.location.Criteria, android.app.PendingIntent%29) of `requestLocationUpdates` about power consumption:

> [...] **Choosing a sensible value for minTime is important to conserve battery life. Each location update requires power from GPS, WIFI, Cell and other radios.** Select a minTime value as high as possible while still providing a reasonable user experience.
> [...] The minDistance parameter can also be used to control the frequency of location updates. If it is greater than 0 then the location provider will only send your application an update when the location has changed by at least minDistance meters, AND at least minTime milliseconds have passed. However **it is more difficult for location providers to save power using the minDistance parameter, so minTime should be the primary tool to conserving battery life**.

And [here](https://android.googlesource.com/device/lge/hammerhead/+/android-4.4.4_r2.0.1/overlay/frameworks/base/core/res/res/xml/power_profile.xml) is a document that is describing the actual power consumption of hardware of one device in some way.

garvankeeley:
What is the bug this patch is fixing? Could you also rename the title more specifically? The word 'improve' is vague to me.

> But only if scanning is paused, then only the network provider is used. I'd say if we are sure that the device is not moving, then we do not need to update the location on the map at all. So I removed this in my patch.

Your comment makes sense, is this the bug that is fixed?

cascheberg:
Yes, the only change is to not enable NetworkLocationListener if ScanningPausedDueToNoMotion.
The NetworkLocationListener is now only enabled until there is a GPS fix.
That means that all location updates are disabled if the device is motionless.

crankycoder:
We should be formalizing the motion detector code.  Wrapping up the location change sensors and when specific sensors go into an active or inactive listening state needs to be done. 

We can't keep slapping code on top and expect no regressions. Get some test cases written or explicitly say "I can't bother to write a test because I'm lazy".

garvankeeley:
Yeah, this would be my original code that has no test case for this stuff. I prefer the answer: the test framework makes it more difficult to write the test than to write the code.  
With regards to this patch, it is UI map view only, which we tend not to have test cases for.

crankycoder:
We can start adding tests in. No time like the present to start.

android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapLocationListener.java
```diff
@@ -24,7 +24,7 @@
     private static final String LOG_TAG = LoggerUtil.makeLogTag(MapLocationListener.class);
     private final WeakReference<MapFragment> mMapActivity;
     private final LocationUpdateListener mNetworkLocationListener =
-            new LocationUpdateListener(LocationManager.NETWORK_PROVIDER, 1000, null);
+            new LocationUpdateListener(LocationManager.NETWORK_PROVIDER, 5000, null);
```
garvankeeley, line 28: Why change this? Won't it make the UI less responsive? I argue there is no evidence to power consumption decrease without significant changes to this parameters on the order of minutes. We have see how the wifi scanner is independent of request frequency (on small time differences) in the wifi scanner code in the app.

Aside:
- the second parameter is particularly ineffective for this purpose, because to detect distance changes requires the location sensor to be active. Now for GPS providers, both of these parameters have different power implications, but the type of GPS chip becomes a factor. For instance some GPS chips can use use a lower-power coarse location state to determine roughly when distance has changed, which makes the second parameter potentially useful.

cascheberg:
I have a better idea on how to improve the `MapLocationListener`. I will check if it works and update this PR then.

cascheberg:
PR updated:
- rebased
- added a passive provider which fixes #1371
- start/stop active location updates if scanning is started/stopped (removing the listener when scanning is stopped was a bug)

crankycoder:
regarding https://github.com/mozilla/MozStumbler/pull/1542#issuecomment-77936239 - I actually _like_ having the map try to update with the network location listener if possible.  In the case the map is not visible, Android can optionally discard that activity to conserve resources.

In the case where the map **is** visible, it is useful to be able to visually verify that the device is registering movement.  Concretely - this is useful when debugging the motion sensor code.

For example #1496 was filed because the network location sensor code updated the position of myself on the map while the service based sensor code never woke up.  I'm deeply skeptical that disabling the network location listener will have any measurable impact on power consumption.  Android is scanning for background networks all the time anyway - unless the user has explicitly gone into the advanced wifi settings and disabled that, the OS has already done all the power up of the antennas.  

Can we get some metrics to verify that this has any measurable impact on power?

Otherwise, I'd like to cut this patch back to just the fix for #1371

cascheberg:
@crankycoder The map's passive listener is still receiving location updates from the network location listener that is enabled in stumbler's motion detection code. Is that enough for debugging?

So, since the listener is enabled there, that means that there is not much power to save here anyway. But I think the code is a bit simpler now that `showPausedDueToNoMotionMessage` / `start` / `stop` all use `enableActiveLocationUpdates`.

cascheberg:
I split this PR up into #1588 and #1589, but without the changes to the paused state.

crankycoder:
I'm closing this as the PR has been recreated in #1588 and #1589.

PR #1543, opened 2015-03-07
Improve description of MLS location queries

### LINKED ISSUE #1427: improve description of "enable MLS location queries"
In settings, I can enable "Enable MLS location queries". But the dialog that pops up doesn't really explain what an MLS location is. I can then display it on the map (I see no red dot if I enable it), and I'm still clueless about what it is.

Can you please improve the description and describe what it is exactly?
Thanks!

stumbler 1.6.1

cascheberg:
Fixes #1427
Further improvements are still welcome of course.

### DIFF
--- a/android/src/main/res/values-de/strings.xml
+++ b/android/src/main/res/values-de/strings.xml
@@ -120,8 +120,8 @@
     <string name="action_developer">Entwickler</string>
     <string name="developer_title">Entwickler</string>
     <string name="enable_option_show_mls_on_map">MLS Ortungsabfragen aktivieren</string>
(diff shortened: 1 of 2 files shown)


cascheberg:
@garvankeeley PR updated. Removing the indentations seemed to be possible only by putting everything in one line. I think the part about 'energy' was added because the feature is fetching data from the internet every few seconds if activated.

garvankeeley:
I would need more convincing that energy is worth mentioning. I don't think it is a measurable effect, being that the cell and wifi are likely active as part of the scanning that is happening when these queries are taking place. And, if the effect of the GPS being active is considered, the effect here is minuscule compared to that.

cascheberg:
OK. Removed 'energy' and updated German translation as well.

PR #1544, opened 2015-03-07
Check zoom buttons when setting min/max zoom level

### LINKED ISSUE #1476: Limited zoom setting: when disabled, zoom buttons on map don't update immediately
The zoom buttons on the map view don't change to reflect that more zooming out is available until the user zooms in (or pinch zooms in or out).

Related to #1474

cascheberg:
Fixes #1476


garvankeeley:
LGTM!

PR #1545, opened 2015-03-08
Draw newer report dots on top

### LINKED ISSUE #1087: the big blue gps dot doesn't get redrawn completely
as you can see the big blue gps dot doesn't get redrawn correctly

the issue vanishes if I move around in the map because that enforces a complete redraw

cascheberg:
Fixes #1087 (see latest comments there)


garvankeeley:
r- because of 
https://github.com/mozilla/MozStumbler/blob/509f0975edc2b1d16f0b6df03911f43ab52b1315/android/src/main/java/org/mozilla/mozstumbler/client/mapview/ObservationPointsOverlay.java#L193

which stops rendering points if the device can't keep up.

cascheberg:
Sorry, I missed that. So assuming that there is no simple fix then, can you close #1087 as mostly FIXED and partly WONTFIX?

garvankeeley:
I'll close this then and I'll explain more in #1087 about how/why the code is doing that

PR #1546, opened 2015-03-08
Fixes to passive scanning mode

cascheberg:
I found some issues in passive scanning mode:
- number of passive scans per location was always one too many because of post increment
- the scanners always needed to be stopped even if a new position was reported instantly
- scanner data was flushed after 10 seconds while wifi scanning continued (3 \* 5 sec = 15 sec). So I decreased `MAX_SCANS_PER_GPS` and `WIFI_MIN_UPDATE_TIME` to fit into one interval, which should also increase data accuracy

I learned that changes to the library code are especially critical, so I'm awaiting comments.

### DIFF
--- a/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/AppGlobals.java
+++ b/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/AppGlobals.java
@@ -26,7 +26,7 @@ public class AppGlobals {
     /* Location constructor requires a named origin, these are created in the app. */
     public static final String LOCATION_ORIGIN_INTERNAL = "internal";
     /* In passive mode, only scan this many times for each gps. */
-    public static final int PASSIVE_MODE_MAX_SCANS_PER_GPS = 3;
+    public static final int PASSIVE_MODE_MAX_SCANS_PER_GPS = 2;
     public static final String NO_TRUNCATE_FLAG = "~";
     public static final String ACTION_TEST_SETTING_ENABLED = "stumbler-test-setting-enabled";
     public static final String ACTION_TEST_SETTING_DISABLED = "stumbler-test-setting-disabled";
--- a/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/ScanManager.java
+++ b/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/ScanManager.java
@@ -38,6 +38,13 @@ public class ScanManager {
     private static Context mAppContext;
     private Timer mPassiveModeFlushTimer;
 
+    // how often to flush a leftover bundle to the reports table
+    // If there is a bundle, and nothing happens for 10sec, then flush it
+    // MAX_SCANS_PER_GPS    is 2
+    // CELL_MIN_UPDATE_TIME is 1sec
+    // WIFI_MIN_UPDATE_TIME is 4sec
+    private static final int FLUSH_RATE_MS = 10000; // 10 sec
+
     private GPSScanner mGPSScanner;
     private WifiScanner mWifiScanner;
     private CellScanner mCellScanner;
@@ -126,16 +133,12 @@ public class ScanManager {
         mWifiScanner.start(ActiveOrPassiveStumbling.PASSIVE_STUMBLING);
         mCellScanner.start(ActiveOrPassiveStumbling.PASSIVE_STUMBLING);
 
-        // how often to flush a leftover bundle to the reports table
-        // If there is a bundle, and nothing happens for 10sec, then flush it
-        final int flushRate_ms = 10000;
-
         if (mPassiveModeFlushTimer != null) {
             mPassiveModeFlushTimer.cancel();
         }
 
         Date when = new Date();
-        when.setTime(when.getTime() + flushRate_ms);
+        when.setTime(when.getTime() + FLUSH_RATE_MS);
         mPassiveModeFlushTimer = new Timer();
         mPassiveModeFlushTimer.schedule(new TimerTask() {
             @Override
--- a/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiScanner.java
+++ b/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiScanner.java
@@ -158,7 +158,7 @@ public class WifiScanner {
             @Override
             public void run() {
                 if (stumblingMode == ActiveOrPassiveStumbling.PASSIVE_STUMBLING &&
-                        mPassiveScanCount++ > AppGlobals.PASSIVE_MODE_MAX_SCANS_PER_GPS) {
+                        ++mPassiveScanCount > AppGlobals.PASSIVE_MODE_MAX_SCANS_PER_GPS) {
                     mPassiveScanCount = 0;
                     stop(); // set mWifiScanTimer to null
                     return;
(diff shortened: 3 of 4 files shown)


libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/AppGlobals.java
```diff
@@ -26,7 +26,7 @@
     /* Location constructor requires a named origin, these are created in the app. */
     public static final String LOCATION_ORIGIN_INTERNAL = "internal";
     /* In passive mode, only scan this many times for each gps. */
-    public static final int PASSIVE_MODE_MAX_SCANS_PER_GPS = 3;
+    public static final int PASSIVE_MODE_MAX_SCANS_PER_GPS = 2;
```
garvankeeley, line 29: Agreed

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/ScanManager.java
```diff
@@ -38,6 +38,13 @@
     private static Context mAppContext;
     private Timer mPassiveModeFlushTimer;
 
+    // how often to flush a leftover bundle to the reports table
+    // If there is a bundle, and nothing happens for 10sec, then flush it
+    // MAX_SCANS_PER_GPS    is 2
+    // CELL_MIN_UPDATE_TIME is 1sec
+    // WIFI_MIN_UPDATE_TIME is 4sec
+    private static final int FLUSH_RATE_MS = 10000; // 10 sec
```
garvankeeley, line 46: This is better.

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiScanner.java
```diff
@@ -35,10 +35,11 @@
     public static final int STATUS_WIFI_DISABLED = -1;
 
     private static final String LOG_TAG = LoggerUtil.makeLogTag(WifiScanner.class);
-    private static final long WIFI_MIN_UPDATE_TIME = 5000; // milliseconds
+    private static final long WIFI_MIN_UPDATE_TIME = 4000; // milliseconds
```
garvankeeley, line 38: Agreed

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiScanner.java
```diff
@@ -153,13 +156,11 @@ synchronized void activatePeriodicScan(final ActiveOrPassiveStumbling stumblingM
         // Ensure that we are constantly scanning for new access points.
         mWifiScanTimer = new Timer();
         mWifiScanTimer.schedule(new TimerTask() {
-            int mPassiveScanCount;
 
             @Override
             public void run() {
                 if (stumblingMode == ActiveOrPassiveStumbling.PASSIVE_STUMBLING &&
-                        mPassiveScanCount++ > AppGlobals.PASSIVE_MODE_MAX_SCANS_PER_GPS) {
```
garvankeeley, line 161: ++mPassiveScanCount instead?

cascheberg, line 161: @garvankeeley The way it was before, the scanner will always do `PASSIVE_MODE_MAX_SCANS_PER_GPS` scans and then always `stop()`.
But it can happen that `ScanManager` is calling `start()` on the scanners faster than the scanner can to its scans, so I changed the code to reset the scan count in `start()` instead, to avoid unnecessary stops. And I think that you need `AtomicInteger` for that to be safe. That's why it is now `mPassiveScanCount.incrementAndGet()`. All related changes are  bundled in my second commit.

garvankeeley, line 161: Agreed, is safer, thanks for the explanation.

garvankeeley, line 161: Shouldn't this be 
AtomicInteger mPassiveScanCount = new AtomicInteger();

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiScanner.java
```diff
@@ -153,13 +156,11 @@ synchronized void activatePeriodicScan(final ActiveOrPassiveStumbling stumblingM
...
-                    mPassiveScanCount = 0;
```
garvankeeley, line 162: Aside: This was unnecessary in the original code. A new instance of this TimerTask would have to be created after a stop() (this TimerTask instance is permanently dead and is not reused).

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/cellscanner/CellScanner.java
```diff
@@ -54,6 +56,8 @@ public void start(final ActiveOrPassiveStumbling stumblingMode) {
             return;
         }
 
+        mPassiveScanCount.set(0);
```
garvankeeley, line 60: Not needed if this becomes a member of TimerTask.

cascheberg, line 60: The idea is that the variable is member of the scanner so that it can be reset from outside the TimerTask.

garvankeeley, line 60: See my comment above on line 161 in WifiScanner.java

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiScanner.java
cascheberg, line 161: Sorry, I do not understand this comment either.
In a1cbaf5392fb824c9bb8f6ff4d7de3c08c5d8752 I made the counter a member of the scanner so that the variable can be reset in `WifiScanner.start()` to keep the TimerTask alive. That way it will scan data for more than one location if there is a new one already.
This is not possible if the variable is a member of the TimerTask (as far as I can see). Otherwise it would be the same as before, or not?

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiScanner.java
```diff
@@ -73,6 +74,8 @@ private boolean isScanEnabled() {
     }
 
     public synchronized void start(final ActiveOrPassiveStumbling stumblingMode) {
+        mPassiveScanCount.set(0);
```
garvankeeley, line 78: Add a comment:
// If the scan timer is active, this will reset the number of times it has run

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/WifiScanner.java
garvankeeley, line 161: I get it now, if you add the comment I mentioned above, this is ready to land.

cascheberg:
Thanks, done in followup commit.
I really should keep in mind to be more verbose about changes...

garvankeeley:
Thanks!

cascheberg:
@crankycoder These changes have not been ported to Fennec. I'd be happy to submit a patch to Bugzilla, should I?

crankycoder:
I'd rather not at this point to be honest.  I'd really like to rework the libstumbler stuff into a proper Android Service and bind that into Fennec.   Any interest in starting to look at that?

Issue #1479, opened 2015-02-11
Service startup consolidation: the app uses binding, passive mode uses startService


garvankeeley:
Also, the passive mode startService starts the StumblerService on its own thread. 
The Mozilla Stumbler app service binding, runs the StumblerService in the main thread.

crankycoder:
I need to read up on the AOSP code to see if there's a real difference between bindService and startService from a thread lifecycle perspective.

crankycoder:
I dug through the AOSP code and ended up tracing the bindService and startService calls into android.app.ActivityManagerNative.  I'm pretty unconvinced there's any good reason to have this split.

I'm refactoring the MainApp and ClientStumblerService so that we run the active stumbler in it's own thread just like the passive mode.

crankycoder:
`bindService` is required to keep the service alive.  It's bizarre and undocumented Android behavior but it's true.

See #1571 for details and evidence of the service lifecycle being different with a bound service vs an unbound service.

PR #1548, opened 2015-03-08
Set map gps/cell/wifi info to zero if not scanning actively

### LINKED ISSUE #1436: paused mode should set gps/wifi/cell values to zero
this is an issue that was brought up in #1322

since the device isn't scanning anything in paused mode those values probably should be set to zero to avoid confusion

cascheberg:
Fixes #1436

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/MainApp.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/MainApp.java
@@ -312,6 +312,10 @@ public class MainApp extends Application
                 .penaltyLog().build());
     }
 
+    public boolean isScanning() {
+        return scannerState == ScannerState.STARTED;
+    }
+
     public boolean isScanningOrPaused() {
          return scannerState == ScannerState.STARTED_BUT_PAUSED_MOTIONLESS ||
                  scannerState == ScannerState.STARTED;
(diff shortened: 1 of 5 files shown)


crankycoder:
Use the intent. I'm removing the service pointer.

cascheberg:
@crankycoder Which intent? Can you already tell when your changes will be ready to be merged?

crankycoder:
The persistedstats class emits an intent for metrics that gets picked up by
the metrics view.

Can you do something similar so that we aren't reliant on a pointer to the
service? We're moving to using an unbound service for both active and
passive stumbling.

crankycoder:
You know what, I'm going to change tact for #1479 so that I don't have to hold patches like this up. I'll send you a small patch to this PR so that we can get this fix in shortly.

cascheberg:
PR updated

crankycoder:
Were you able to reliably reproduce a bug?

cascheberg:
@crankycoder What exactly are you asking? The PR fixes #1436, which is a minor cosmetic issue. The updated PR is rebased and I changed the way the displayed GPS satellites are set to zero.

Aside from that, I think the UI update methods are a bit messy. Could we remove `updateUiOnMainThread` and put more dedicated intent receivers directly into `MainDrawerActivity`?

crankycoder:
Thanks!

crankycoder:
This fixes the lag in #1567

PR #1549, opened 2015-03-09
Revert "fix notification updates"

crankycoder:
This reverts commit df68ed93d32472a5072cd4c84ca678a5f9a184b0.

I'm reverting this because we can do this in a more simplified way that doesn't impose a dependency on the service pointer.

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/navdrawer/MainDrawerActivity.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/navdrawer/MainDrawerActivity.java
@@ -36,7 +36,6 @@ import org.mozilla.mozstumbler.client.Updater;
 import org.mozilla.mozstumbler.client.mapview.MapFragment;
 import org.mozilla.mozstumbler.client.subactivities.FirstRunFragment;
 import org.mozilla.mozstumbler.client.subactivities.LeaderboardActivity;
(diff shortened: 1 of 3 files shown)


cascheberg:
@crankycoder Can you be more specified on how this improves dependencies? The references to `service.getUniqueCellCount()` and `service.getUniqueAPCount()` have been there before and are still there.

crankycoder:
Sorry, I wasn't clear on this.  

This is all related to #1479 where we are removing the bindService call and using a startService to start up both active and passive stumbling modes.

The end goal is that we want to be able to expose not just passive, but active stumbling to libstumbler users.  

Getting rid of the bindService call means that a pointer to `service` will no longer be available.  All communication between the application and the stumbling service will need to happen over Intent messaging.

I've already removed both of the method calls you're talking about in my own branch.  I'm simply reluctant to have any inbound patches that do not have test cases that cover a feature where there is a very good chance that I will cause a regression due to my service refactoring. 

I'm hitting a lot of side effects where UI code expects a service pointer to be available - without having some test to exercise a feature, I can't safely refactor.  We really need to get better at adding tests to cover features so that more aggressive refactorings can occur safely.

cascheberg:
Thanks! Looking forward to see your changes merged ;)

PR #1550, opened 2015-03-09
Fix for 1484, on startup update the last upload in the notification

garvankeeley:
See #1484


crankycoder:
reviewed this in IRC

garvankeeley:
@crankycoder r? 
This should be a minimal way of fixing this without affecting your refactor.

cascheberg:
Now the notification is always shown, and it is not going away if the stumbler is stopped or even killed. This is probably because it is calling `update` once after the stumbler has stopped.
Do I need to reboot my device now to get rid of it? I think a quick fix would be great.

garvankeeley:
Thanks @cascheberg 
This should fix it https://github.com/mozilla/MozStumbler/pull/1553

PR #1551, opened 2015-03-09
removing whats new dialog

garvankeeley:
I left the pref in, and the block of code where we would put this again in future, in hopes it get replaced with a whats new list pulled from the web


cascheberg:
@garvankeeley You removed some actual German translations instead of the `whats_new` string.

garvankeeley:
Oh no, i did. Have to fix

PR #1559, opened 2015-03-11
Always use passive scanning in cell and wifi scanners

### LINKED ISSUE #1055: Investigate network scanner scanning frequency
Currently the frequencies of GPS position updates, cell scanning, and wifi scanning are configured independently from each other. The `WiFiScanner` has `WIFI_MIN_UPDATE_TIME` = 5000ms and the `CellScanner` has `CELL_MIN_UPDATE_TIME` = 1000ms, while the `GPSScanner` requests location updates after 10m movement occured and at least 1s passed (#1039 -> 30m, 2s).

I would assume that based on the range of different network types, wifis need to be scanned much more frequently than cells, because a wifi might not be visible anymore after moving >10m, while cells can potentially span several kilometers. This makes sense even when only having observation points every 30m.

If you want to scan for wifis every ~15m at 50km/h, you would need to scan every 1 second.
I wonder if it would be best to hook up the network scanners to the GPS too, to start each scan based on passed time and movement as well.

### LINKED ISSUE #1301: only scan for GPS before first GPS fix
this is probably interesting if there will be some automation in the future (NFC, Tasker, boot, etc.):
MozStumbler should only search for GPS to save battery (on application start)
it doesn't make sense to fire wifi/cell timers if the data doesn't get saved into reports

### LINKED ISSUE #1503: Unique number of available cells keeps increasing without a gps fix (on map)
I think this bug is only optical because afterwards (acoording to the log) only 3 cells were flushed to the disk (report).

If a device doesn't find a GPS fix it keeps on scanning (without the motion sensor), the number of wifi aps stays normal but the number of cells (on top of the map) is steadily increasing (cells that aren't available anymore aren't removed until the device gets a gps fix and reports are flushed to disk

I managed to get the number up to 95 even (and it is probably possible to get it to 1000 if you wait long enough)

how to reproduce this:
sit in train activate your gps and start scanning
if your device hasn't been connected to GPS before you went into the train it is likely that your device has trouble getting a gps fix in the train -> you will keep on scanning for a longer time but don't get any fix

^ in the first screenshot you can see it went up to 15, the second screenshot shows the log, when I went out of the train and the first report was flushed to disk (2 cells) -> the number of visible cells immediatly went down to 2/3 cells on the map (this issue only seems to be optical)

I can provide a kml file if needed

cascheberg:
I think it might be a good idea to always run the cell and wifi scans in passive mode.
As a result, only two scans will be done per location, and only when a new location is reported by the active GPS. After that the report will be flushed.
That also means that scanning is not active if the device is not reporting new locations because it is either not moving (which works independently from motion sensing settings) or has no GPS fix.
And the number of scans is reduced if you are moving slowly. The current min distance for GPS to report a new location is 30 meters. At a regular walking speed of 5-6km/h this takes ~20 seconds. In that time, 20 cell and 5 wifi scans are made. With this patch, it's always only 2 scans each.
With the current settings, wifi scanning will become permanently active at ~13km/h, cell scanning at ~50km/h.
And Stumbler users will be running and actively testing the same code that is run in passive mode used in Android Firefox.
Also fixes #1503, fixes #1301 and basically fixes #1055.

### DIFF
--- a/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/cellscanner/CellScanner.java
+++ b/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/cellscanner/CellScanner.java
@@ -39,7 +39,7 @@ public class CellScanner {
     private final ISimpleCellScanner mSimpleCellScanner;
     private Timer mCellScanTimer;
     private Handler mBroadcastScannedHandler;
-    private AtomicInteger mPassiveScanCount = new AtomicInteger();
+    private AtomicInteger mScanCount = new AtomicInteger();
 
     public CellScanner(Context appCtx) {
         mAppContext = appCtx;
@@ -56,7 +56,7 @@ public class CellScanner {
         }
 
         // If the scan timer is active, this will reset the number of times it has run
-        mPassiveScanCount.set(0);
+        mScanCount.set(0);
 
         if (mCellScanTimer != null) {
             return;
@@ -86,7 +86,7 @@ public class CellScanner {
                     return;
                 }
 
-                if (mPassiveScanCount.incrementAndGet() > AppGlobals.PASSIVE_MODE_MAX_SCANS_PER_GPS) {
+                if (mScanCount.incrementAndGet() > AppGlobals.MAX_SCANS_PER_GPS) {
                     stop();
                     return;
                 }
--- a/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/cellscanner/CellScanner.java
+++ b/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/cellscanner/CellScanner.java
@@ -13,7 +13,6 @@ import android.os.Message;
 import android.support.v4.content.LocalBroadcastManager;
 
 import org.mozilla.mozstumbler.service.AppGlobals;
-import org.mozilla.mozstumbler.service.AppGlobals.ActiveOrPassiveStumbling;
 import org.mozilla.mozstumbler.service.Prefs;
 import org.mozilla.mozstumbler.service.stumblerthread.Reporter;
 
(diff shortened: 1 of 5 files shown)


garvankeeley:
This is a great idea. Historically, there was a design intention behind keeping the cell and wifi scanners active before the GPS came in, but that predates my involvement. I think this is left over from some approaches at maximizing the correlation of the Wi-Fi/cell reading and the GPS, but we no longer have an interest this area of optimization. There is an inherent inaccuracy in the devices for wifi scanning, but also, errors are smoothed out on the server side once repeated scans are received.

lua86:
Which remainds me, where should I report a hundreds of kilometers error in the location?

crankycoder:
@lua86 please just file a separate bug instead of commenting on an existing unrelated bug in the future.  Keeping the message threads clean helps us triage bugs quickly.  Your issue is probably something we addressed in the 1.7.7 release I just cut.  See #1558

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/cellscanner/CellScanner.java
```diff
@@ -128,8 +129,6 @@ private synchronized void addToCells(String cell) {
     }
 
     public synchronized void stop() {
-        mReportWasFlushed.set(false);
-        clearCells();
```
garvankeeley, line 132: isn't clearCells() still needed? If the scanning starts again, the old cells will still be in the list (I think)

garvankeeley:
r+
Other than my one comment about clearCells(), this is ready to land.

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/cellscanner/CellScanner.java
cascheberg, line 132: I set `mReportWasFlushed` in `start()` to clear the list

PR #1563, opened 2015-03-11
Unbind the ClientStumblerService

crankycoder:
This is a WIP cherry picking from lines and reworking them from : https://github.com/crankycoder/MozStumbler/tree/features/1479-remove-bindservice

This unbinds dependencies to the ClientStumblerService pointer that is exposed with `bindService` from the client and spins the stumbler service off into a completely separate thread.

Please see each commit message for details.

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/ClientStumblerService.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/ClientStumblerService.java
@@ -4,19 +4,25 @@
 
 package org.mozilla.mozstumbler.client;
 
+import android.app.Notification;
+import android.content.BroadcastReceiver;
+import android.content.Context;
 import android.content.Intent;
+import android.content.IntentFilter;
 import android.os.Binder;
 import android.os.IBinder;
 import android.os.Looper;
 import android.support.v4.content.LocalBroadcastManager;
-import android.util.Log;
 
+import org.mozilla.mozstumbler.R;
+import org.mozilla.mozstumbler.client.util.NotificationUtil;
 import org.mozilla.mozstumbler.service.AppGlobals;
 import org.mozilla.mozstumbler.service.stumblerthread.StumblerService;
 import org.mozilla.mozstumbler.service.stumblerthread.datahandling.ClientDataStorageManager;
-import org.mozilla.mozstumbler.service.stumblerthread.datahandling.DataStorageManager;
 import org.mozilla.mozstumbler.service.utils.BatteryCheckReceiver;
 import org.mozilla.mozstumbler.service.utils.BatteryCheckReceiver.BatteryCheckCallback;
+import org.mozilla.mozstumbler.svclocator.ServiceLocator;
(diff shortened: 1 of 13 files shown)


cascheberg:
Cool! Should be easy for me to adjust #1548 once this lands.

Maybe you could generate the intent action strings from the enum and vice versa, like this:

```
public static enum State {
    STARTED, STOPPED;

    private final static String prefix = "NAMESPACE.PREFIX_" ;

    @Override
    public String toString() {
        return prefix + this.name();
    }

    public static State fromString(String name) throws IllegalArgumentException {
        return State.valueOf(name.substring(State.prefix.length()));
    }
}
```

crankycoder:
@cascheberg Sweet! I totally forgot you could do that with enum types.  That's a great idea.

android/src/main/java/org/mozilla/mozstumbler/client/MainApp.java
```diff
@@ -65,7 +67,9 @@
 public class MainApp extends Application
         implements AsyncUploader.AsyncUploaderListener {
     public static final AtomicBoolean isUploading = new AtomicBoolean();
```
garvankeeley: no longer used

android/src/main/java/org/mozilla/mozstumbler/client/ClientStumblerService.java
```diff
@@ -25,6 +31,30 @@
...
+    private static final ILogger Log = (ILogger) ServiceLocator.getInstance().getService(ILogger.class);
+
+    public static enum RequestChangeScannerState {
+        START, STOP;
+
+        public static final String NAMESPACE = "org.mozilla.mozstumbler.clientstumblerservice.state";
+
+        @Override
+        public String toString() {
+            return NAMESPACE+ this.name();
```
garvankeeley: nit: space before +

android/src/main/java/org/mozilla/mozstumbler/client/MainApp.java
```diff
@@ -77,10 +81,38 @@
     private ServiceBroadcastReceiver mReceiver;
     private WeakReference<IMainActivity> mMainActivity = new WeakReference<IMainActivity>(null);
     private boolean mIsScanningPausedDueToNoMotion;
+
+    // These track the state of the currently running service
+    private ScanManager.ScannerState scannerState = ScannerState.STOPPED;
+
+
+    boolean isStumblerStopped() {
```
garvankeeley: currently unused, but lets leave this in.

android/src/main/java/org/mozilla/mozstumbler/client/MainApp.java
```diff
@@ -390,6 +412,10 @@ public boolean isIsScanningPausedDueToNoMotion() {
         return mIsScanningPausedDueToNoMotion;
     }
 
+    public boolean isStopped() {
```
garvankeeley, line 352: oops, we have 2 'is stopped' functions in this file.

garvankeeley:
r+
Other than my few comments, this is ready to land

PR #1566, opened 2015-03-12
WIP ->Show cells on map

garvankeeley:
This is a WIP patch. It is largely functional, highly graphics performant, but needs more work and more documentation.

I would like feedback on slower devices, the one I have to test with isn't working.

The rate limiting step is the server, which is slow to respond.
The CellDataRequester currently has to turn map tile ids of the string form "tilex/tiley" into lat long bounding boxes for server queries. This is temporary. The server will be changed to handle the "tilex/tiley" queries.

Future items:
- when map zoomed in, draw mini towers instead of black dots.
- when drawn as towers (above), make them tappable, and show the tower info
- manage the storage of cells on disk better. 
  --- They are cached, need to expire this cache
  --- Need to limit size on disk


garvankeeley:
This is more for review than for landing. Since it isn't behind a developer setting. If someone wants to try the patch they are welcome to do so. Not sure when I'll have time to finish it. Hopefully soon.

garvankeeley:
Another TODO item will be in CellPointsOverlay, when the server side gets change to tile handling, will be to change the JSON parsing to match and changes in the JSON structure.

PR #1577, opened 2015-03-17
ignore GPS records with a time < the unix buildtime of the app

### LINKED ISSUE #1526: check phone clock to avoid GPS issues (unix time, see mailing list)
We should warn the user if the phone clock is off (for example: Unix time: 1970, or if the phone clock timestamp is in general lower than the build timestamp) to avoid collecting wrong data

this is not only about wrong timestamps but also about the GPS
if the phone clock is off that far, then the GPS position is likely wrong as well, so we shouldn't collect data then and avoid uploading it.

Maybe we could even implement a time check (compare current phone to server time before uploading or something like that)

^this issue is a result of the following discussion on the mailing list:
https://groups.google.com/forum/#!topic/mozilla.dev.geolocation/biwiie__M10

crankycoder:
This fixes #1526 by comparing the time collected from the GPS location record against the bulid time of the application.  The Mozilla build servers should be ok, but I'm uncertain about the F-Droid server. That _should_ be ok, but the only way to know for sure is to just make a release.

### DIFF
--- a/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/StumblerFilter.java
+++ b/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/StumblerFilter.java
@@ -5,6 +5,7 @@
 package org.mozilla.mozstumbler.service.stumblerthread.scanners;
 
 import android.location.Location;
+import android.os.Build;
 import android.util.Log;
 
 import org.mozilla.mozstumbler.svclocator.services.log.LoggerUtil;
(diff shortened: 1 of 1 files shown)


libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/scanners/StumblerFilter.java
```diff
@@ -21,7 +22,7 @@
     private static final double MIN_ALTITUDE = -418;      // Dead Sea's altitude in meters
     private static final float MAX_SPEED = 340.29f;   // Mach 1 in meters/second
     private static final float MIN_ACCURACY = 500;       // meter radius
-    private static final long MIN_TIMESTAMP = 946684801; // 2000-01-01 00:00:01
+    private static final long MIN_TIMESTAMP = Build.TIME; // The build time of the application
```
crankycoder, line 25: This is wrong.  Build.TIME is the buildtime of the operating system, not the application. We don't seem to record the application build time anywhere and Android doesn't do it out of the box for us.

PR #1580, opened 2015-03-17
disbles requirement for accelerometers

### LINKED ISSUE #1309: Set required flag on accelerometer to false
Android devices should have accelerometers, but don't have to.

http://static.googleusercontent.com/media/source.android.com/en//compatibility/android-cdd.pdf

crankycoder:
This fixes #1309


crankycoder:
Just a note, I have no idea how we'd actually test this as we have no hardware to test this on.

PR #1584, opened 2015-03-20
Minor fixes

### LINKED ISSUE #1436: paused mode should set gps/wifi/cell values to zero

cascheberg:
first commit: immediately update number display
STR:
- do some scanning, then stop
- rotate device etc -> report count is zero

second commit: properly fix #1436

third commit: for debug builds, use the app debug name in the notification as well

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/navdrawer/MainDrawerActivity.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/navdrawer/MainDrawerActivity.java
@@ -250,7 +250,7 @@ public class MainDrawerActivity
     @Override
     public void onPostResume() {
         super.onPostResume();
-        mMetricsView.update();
+        updateNumberDisplay(true);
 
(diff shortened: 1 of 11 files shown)


android/src/main/java/org/mozilla/mozstumbler/client/navdrawer/MainDrawerActivity.java
```diff
@@ -250,7 +250,7 @@ protected void onPostCreate(Bundle savedInstanceState) {
     @Override
     public void onPostResume() {
         super.onPostResume();
-        mMetricsView.update();
+        updateNumberDisplay(true);
```
crankycoder, line 253: Yay. Much nicer.

crankycoder:
Thanks @cascheberg !

PR #1585, opened 2015-03-20
Remove conditional location of stumbler files

crankycoder:
I've removed a block of code which put the stumbler reports into a special directory for debug builds.

This was just confusing when diagnosing a problem with the DemoStumbler.  In a nutshell - AppGlobals.isDebug is not final and it had a value which I wasn't expecting, so files were not being written or read from an expected directory.

I don't think there's really any benefit to multiple directories anyway. If we really want to have an alternate directory, I would opt for an explicit path that is set into the DataStorageManager.


crankycoder:
I've updated ths to use a commented out block if you really really need to debug this.

PR #1587, opened 2015-03-23
Fix persisted stats metrics

### LINKED ISSUE #1568: Intermittent failure to have a persistent "Last upload date".  Occasionally resets to 'never'
MozStumbler: 1.7.7
CyanogenMod CM11 

Following an update to MozStumbler 1.7.7, the application is no longer reporting results in the `Already Sent` section.

`Last uploaded` is always listed as `never` despite successful uploads according to the logs. Previously this would show the last uploaded date
`Roports` is left blank, previously this would show the number of records uploaded since install

cascheberg:
Making `mPersistedStats` -> `static sPersistedStats` fixes #1568


garvankeeley:
Looks ok to me!

PR #1588, opened 2015-03-23
Add passive location listener to MapLocationListener

### LINKED ISSUE #1371: MapLocationListener is too slow

cascheberg:
Fixes #1371

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapLocationListener.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapLocationListener.java
@@ -35,6 +35,8 @@ class MapLocationListener {
     };
     private final LocationUpdateListener mGpsLocationListener =
             new LocationUpdateListener(LocationManager.GPS_PROVIDER, 5000, mReceivedGpsLocation);
+    private final LocationUpdateListener mPassiveLocationListener =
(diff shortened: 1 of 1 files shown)


garvankeeley:
Except now there would be a GPS listener, a network listener, AND a passive listener. I don't think all 3 are needed. The passive listener could get GPS locations, so can the GPS listener be removed?

cascheberg:
I think you could stop this GPS listener while the scanning is active, but that will make the code more complicated without much benefit.
Completely removing the GPS listener here will result in not getting a location if scanning is stopped and Android's network provider is disabled / offline.

garvankeeley:
Agreed, I think 3 are still needed.
- Enable passive listener at all times
- when scanning stopped, enable GPS+network listener

Maybe pauseGpsUpdates() can go away?

cascheberg:
@garvankeeley Your suggestions are basically dealing with code changes in #1589, so I commented there. Maybe you can merge this PR and then continue the discussion there about what "properly configure" actually should be like?

PR #1589, opened 2015-03-23
Properly configure map location and GPS listeners on start/stop

cascheberg:
This fixes the bug that the GPS status listener is removed on stop, and that map location listeners are not re-enabled on start.

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/IMainActivity.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/IMainActivity.java
@@ -18,6 +18,9 @@ public interface IMainActivity {
     public void isPausedDueToNoMotion(boolean isPaused);
 
     // Call from main thread only
+    public void start();
+
+    // Call from main thread only
     public void stop();
 }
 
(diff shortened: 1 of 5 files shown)


cascheberg:
This is referring to @garvankeeley's comment in #1588:

> when scanning stopped, enable GPS+network listener
> Maybe pauseGpsUpdates() can go away

If scanning is started but there is no GPS fix, then only enabling the passive listener will not update the location (while enabling the network provider could do that). I am not sure if this will always be handled by the `mSatelliteListener` checking `MIN_SAT_USED_IN_FIX`?

About enabling GPS when scanning is stopped: if the user clicks the button to stop scanning, then Android's GPS indicator icon would not go away as long as the map is shown, which might be confusing.
Currently the GPS is then only reactivated in this state when the map is resumed. But on the other hand, depending on the use case, that behavior could be called a bug. 

About the paused state: the device should not be moving, so location updates are unnecessary, but @crankycoder mentioned enabling active location updates for debugging purposes

I hope my thoughts are comprehensible. The easiest solution would be to always enable all listeners while the map is shown.

garvankeeley:
I am going to land this and the other patch, and follow up with a minor refactor of this class which is getting a bit dense with stateful logic.

Issue #1579, opened 2015-03-17
Power consumption is up on Nexus 5


crankycoder:
User bug reported: "The last 2 or 3 updates seem to have regressed motion detection on my nexus 5. Even overnight where the phone is completely motionless, seems to continually scan and not pause causing significant battery usage."

sshvetsov:
I've got Nex5, so I can try verifying this. Any suggestions on the test procedure?

crankycoder:
I think you just leave the phone idle overnight while you sleep.  If significant power drain happens,  then we have a bug.

sshvetsov:
Did one test last night. So far doesn't seem out of the ordinary, just 1% consumption by MozStumbler when stationary over period of 7 hours, compared to 7%  over period of 2 hours while moving. Will do another test tonight.

Test 1:
Phone laying on the desk over night:
![idle1](https://cloud.githubusercontent.com/assets/2501176/6746120/88cfbb96-cef5-11e4-90ea-c5e9423f37d7.png)![idle2](https://cloud.githubusercontent.com/assets/2501176/6746121/88d1a30c-cef5-11e4-8681-0e30f970bdbd.png)
Trip to work:
![active1](https://cloud.githubusercontent.com/assets/2501176/6746122/88d47028-cef5-11e4-8db7-7f2128732416.png)

sshvetsov:
Second test produed similar results. Here are the screen shots from my tests:

Test 2:
Phone laying on the desk over night:
![idle1](https://cloud.githubusercontent.com/assets/2501176/6746148/f9bcfbde-cef5-11e4-9935-0fba3072e60a.png)![idle2](https://cloud.githubusercontent.com/assets/2501176/6746149/f9bd8162-cef5-11e4-9242-bd357ed44333.png)![idle3](https://cloud.githubusercontent.com/assets/2501176/6746150/f9c01af8-cef5-11e4-954e-1750a26d7a47.png)
Trip to work:
![active1](https://cloud.githubusercontent.com/assets/2501176/6746171/5ec4d628-cef6-11e4-8b66-8cc88f73a8ed.png)![active2](https://cloud.githubusercontent.com/assets/2501176/6746170/5ec4c34a-cef6-11e4-924d-e8414d350d1c.png)

dhawkshaw:
Hi, I'm the original user reporting this issue.

[...]

PR #1590, opened 2015-03-24
Use requestLocationUpdates options min time and distance to sense motion

garvankeeley:
Issue #1579 

I think the bug is that the NetworkLocationChangeDetector sets the
initial location with a GPS location, then compares it to the Network
location.
The last time I tried using these options was a year ago, and I was not
getting onLocationChanged when I expected it to (sporadically missing
sending this event). I have avoided using them as a result, but in this case, 
they may be the solution.


garvankeeley:
Alternatively, we could scan WiFis ourselves, and if a 50% of them change, assume a motion has taken place.

crankycoder:
Looks good. Just testing this now.

PR #1592, opened 2015-03-24
Only reschedule alarm when the alarm delay, or isRepeating, are changed

garvankeeley:


garvankeeley:
n.b. change DemoStumbler to match this change

crankycoder:
I sent you a PR with a test fix and an extra test case.

PR #1597, opened 2015-03-27
MapLocationListener refactor for dedicated classes for the 2 listening states.

garvankeeley:
This should make the logic clearer for future, and modularizes the location listeners to their purpose.
This is a follow up to #1588 and #1589.


garvankeeley:
@cascheberg do you have time to review this? You know this area of the code very well.

cascheberg:
Do you have any specifications about what you expect to happen depending on the state? (See https://github.com/mozilla/MozStumbler/pull/1589#issuecomment-85737974)

I did not run your updated code yet, so I hope I got this right:
- scanning: use passive provider only
  - downside: no position update as long as there is no GPS fix
  - current: enable both
- paused: use GPS and network provider, but keeps using 'passive provider only' as long as `MapFragment` is not paused and resumed
  - downside: this should not be necessary, but is this (or just enabling network provider) wanted for debugging?
  - current: enable network provider only
- stopped: use GPS and network provider
  - this is OK, but it will keep the GPS active if the user stops scanning from the map view
  - current: disable both, re-enable both when `MapFragment` is paused and resumed

Side note about the passive provider:
- scanning: GPS location updates every 30m / 2s (if there is a GPS fix)
- paused:  `MotionSensor` enables a network provider which updates every 30s

garvankeeley:
Good questions!

> I did not run your updated code yet, so I hope I got this right:
> - scanning: use passive provider only
>   - downside: no position update as long as there is no GPS fix
>   - current: enable both

Hmm, I focused too much on cleanup and simplification, and forgot about this downside. (*)

> - paused: use GPS and network provider, but keeps using 'passive provider only' as long as `MapFragment` is not paused and resumed

LocationListenerWhileScanningOn only has a passive listener, and that class encompasses the 2 substates of "scanning on" which are paused and not-paused.

> - downside: this should not be necessary, but is this (or just enabling network provider) wanted for debugging?

I don't think so.

> - stopped: use GPS and network provider
>   - this is OK, but it will keep the GPS active if the user stops scanning from the map view

As long as the map view is visible. This would be the same in virtually any app with a map view, that as the user is viewing it, GPS scanning is active.

> - current: disable both, re-enable both when `MapFragment` is paused and resumed

Yikes, I hope I didn't unhook this, will check after this. (*)

> Side note about the passive provider:
> - scanning: GPS location updates every 30m / 2s (if there is a GPS fix)

Good point, I was testing in a car after I changed this. I suspect if I walk I won't like like this behaviour, as the map view will seem strangely unresponsive, and jump to new positions. (*)

> - paused:  `MotionSensor` enables a network provider which updates every 30s

Which provides only a single update.

I marked the items with a (*) that I am following up on.

cascheberg:
Hm based on the things you want to follow up on, it seems that `LocationListenerWhileScanningOn` will end up using GPS and network providers as well. Which makes me think that cleaning up the existing code just a bit more might be just as good.

android/src/main/java/org/mozilla/mozstumbler/client/mapview/maplocation/MapLocationListener.java
```diff
@@ -119,59 +119,40 @@ public void removeListener() {
...
 
-    private class LocationUpdateListener implements LocationListener {
         final ReceivedLocationCallback mCallback;
         final long mFreqMs;
         final String mType;
         boolean mIsActive;
+        private final WeakReference<MapFragment> mMap;
 
-        public LocationUpdateListener(String type, long freq, ReceivedLocationCallback callback) {
+        public MapPositionUpdater(MapFragment map, String type, long freq, ReceivedLocationCallback callback) {
```
crankycoder: javadoc here please.

android/src/main/java/org/mozilla/mozstumbler/client/mapview/maplocation/LocationListenerWhileScanningOff.java
```diff
@@ -0,0 +1,69 @@
...
+
+    public void start() {
+        setActiveLocationUpdatesEnabled(true);
+    }
+
+    public void stop() {
+        setActiveLocationUpdatesEnabled(false);
+    }
+
+    private  void setActiveLocationUpdatesEnabled(boolean enable) {
```
crankycoder: Can I get a docstring here for clarity?

android/src/main/java/org/mozilla/mozstumbler/client/mapview/maplocation/LocationListenerWhileScanningOff.java
```diff
@@ -0,0 +1,69 @@
...
+
+    private final MapPositionUpdater mNetworkLocationListener;
+    private final MapPositionUpdater mGpsLocationListener;
+
+    private final WeakReference<MapFragment> mMapFragment;
+
+    LocationListenerWhileScanningOff(LocationManager manager, MapFragment mapFragment) {
+        mLocationManager = manager;
+        mMapFragment = new WeakReference<MapFragment>(mapFragment);
+        mNetworkLocationListener = new MapPositionUpdater(mapFragment, LocationManager.NETWORK_PROVIDER, 1000, null);
```
crankycoder: Nit pick.  I'd rather push the location provider argument `LocationManager.NETWORK_PROVIDER` and `LocationManager.GPS_PROVIDER` down into the class definition itself.   The location provider implies a type of the class anyway.

android/src/main/java/org/mozilla/mozstumbler/client/mapview/maplocation/LocationListenerWhileScanningOff.java
```diff
@@ -0,0 +1,69 @@
...
+    public void start() {
+        setActiveLocationUpdatesEnabled(true);
+    }
+
+    public void stop() {
+        setActiveLocationUpdatesEnabled(false);
+    }
+
+    private  void setActiveLocationUpdatesEnabled(boolean enable) {
+        if (mLocationManager == null) {
```
crankycoder: Should we move - or at least have a check for the location manager being null in `LocationListenerWhileScanningOff` constructor?

android/src/main/java/org/mozilla/mozstumbler/client/mapview/maplocation/LocationListenerWhileScanningOn.java
```diff
@@ -0,0 +1,28 @@
...
+import org.mozilla.mozstumbler.client.mapview.maplocation.MapLocationListener.MapPositionUpdater;
+
+// The "scan on" state encompasses the paused state
+public class LocationListenerWhileScanningOn {
+    private final LocationManager mManager;
+    private final MapPositionUpdater mPassiveLocationListener;
+
+    LocationListenerWhileScanningOn(LocationManager manager, MapFragment mapFragment) {
+        mManager = manager;
+        mPassiveLocationListener = new MapPositionUpdater(mapFragment, LocationManager.PASSIVE_PROVIDER, 1000, null);
```
crankycoder: Same nitpick as before - I'd rather see a subclass like `PassiveMapPositionUpdater`. So something like `<LocationProviderType>MapPositionUpdater` as subclass names.

android/src/main/java/org/mozilla/mozstumbler/client/mapview/maplocation/MapLocationListener.java
```diff
@@ -51,21 +42,24 @@ public void onGpsStatusChanged(int event) {
...
                 }
-                if (mMapActivity.get() != null) {
-                    mMapActivity.get().updateGPSInfo(satellites, fixes);
+                if (mMapFragment.get() != null) {
+                    mMapFragment.get().updateGPSInfo(satellites, fixes);
                 }
+
                 if (fixes < GPSScanner.MIN_SAT_USED_IN_FIX) {
-                    enableLocationListener(true, mNetworkLocationListener);
+                    mLocationListenerScanningOff.notEnoughSatellites();
```
crankycoder: This is a lot nicer!

android/src/main/java/org/mozilla/mozstumbler/client/mapview/maplocation/MapLocationListener.java
```diff
@@ -119,59 +119,40 @@ public void removeListener() {
         }
 
         mLocationManager.removeGpsStatusListener(mSatelliteListener);
-        setActiveLocationUpdatesEnabled(false);
-        enableLocationListener(false, mPassiveLocationListener);
+        mLocationListenerScanningOff.stop();
```
crankycoder: +1 much clearer.

crankycoder:
A class level docstring to explain the state transititions will make it easier for someone new to the code to follow through the state changes and help people learn how to think about this chunk of code.

garvankeeley:
@cascheberg I agree now that there isn't 2 states here, trying to impose that just causes regressions/bugs. Let me try again to clean this up and update the PR...

garvankeeley:
I am going to rebase this to remove the first patch, as the diff view is confusing. I only want to see this diff against the original code

garvankeeley:
Squashed and forced the push.

garvankeeley:
MapLocationListener -> is now UserPositionUpdateManager, which handles when the map is shown and hidden, and start/stop of the MultiSourceLocationListener.

The new class MultiSourceLocationListener has 2 location listeners (network and gps types) in the form of MapUpdatingLocationListeners

android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java
```diff
@@ -676,23 +675,20 @@ public void showMapNotAvailableMessage(NoMapAvailableMessage noMapAvailableMessa
...
-            mMapLocationListener.pauseGpsUpdates(show);
-        }
         updateGPSInfo(0, 0);
         dimToolbar();
     }
 
     public void start() {
         if (mMapLocationListener != null) {
-            mMapLocationListener.setActiveLocationUpdatesEnabled(true);
+            mMapLocationListener.setMapVisible(true);
```
cascheberg: The map becomes visible in `MapFragment.doOnResume` and invisible in `MapFragment.onPause`. These methods indicate a scanner state change while the map is visible.

garvankeeley:
@cascheberg addressed your review comment in 3b3c351

android/src/main/java/org/mozilla/mozstumbler/client/mapview/maplocation/UserPositionUpdateManager.java
```diff
@@ -0,0 +1,108 @@
...
+                    satellites++;
+                    if (sat.usedInFix()) {
+                        fixes++;
+                    }
+                }
+                if (mMapFragment.get() != null) {
+                    mMapFragment.get().updateGPSInfo(satellites, fixes);
+                }
+
+                if (fixes < GPSScanner.MIN_SAT_USED_IN_FIX) {
```
crankycoder: Can we push `mSatelliteListener`, the registration of `addGpsStatusListener` and this `if` block into the multisource location listener - possibly down into the GPS update listener?  It doesn't make sense that we have GPS code up in the UserPositionUpdateManager when we already have an explcit GPS listener class that is assisting here.

crankycoder:
I'm going to let this run for a day to make sure we didn't miss anything, but otherwise let's get this merged in tomorrow (2015/04/02)

android/src/main/java/org/mozilla/mozstumbler/client/mapview/maplocation/MultiSourceLocationListener.java
```diff
@@ -0,0 +1,94 @@
...
+
+        if (enable) {
+            // enable NetworkLocationListener first because GpsLocationListener will disable it
+            UserPositionUpdateManager.enableLocationListener(mLocationManager, true, mNetworkLocationListener);
+            UserPositionUpdateManager.enableLocationListener(mLocationManager, true, mGpsLocationListener);
+        } else {
+            // disable GpsLocationListener first because it might enable NetworkLocationListener
+            UserPositionUpdateManager.enableLocationListener(mLocationManager, false, mGpsLocationListener);
+            UserPositionUpdateManager.enableLocationListener(mLocationManager, false, mNetworkLocationListener);
+        }
```
cascheberg, line 83: You could now remove this method and merge it into `start` and `stop`, because it's only called from these methods, once with parameter `true`, and from the other with parameter `false`.

cascheberg:
I did not run the code, but it looks nice. So basically the map will now keep both location providers enabled while it's visible?

garvankeeley:
@cascheberg yes

Issue #1599, opened 2015-03-30
MLS queries are slow and can cause a concurrent modification exception


crankycoder:
In v1.7.10, ConcurrentModifcationException is thrown sometimes.

``` java
java.util.ConcurrentModificationException
at java.util.LinkedList$LinkIterator.next(LinkedList.java:124)
at org.mozilla.mozstumbler.client.ObservedLocationsReceiver$1.run(ObservedLocationsReceiver.java:101)
at android.os.Handler.handleCallback(Handler.java:739)
at android.os.Handler.dispatchMessage(Handler.java:95)
at android.os.Looper.loop(Looper.java:135)
at android.app.ActivityThread.main(ActivityThread.java:5312)
at java.lang.reflect.Method.invoke(Native Method)
at java.lang.reflect.Method.invoke(Method.java:372)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:901)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:696)
```

Reading the code, the ObservedLocationsReceiver has a method `getMapActivity` which iterates over the queuedForMLS observation points and iteratively queries MLS one at a time using AsyncTasks.

There should be a lock put around the use of the list so that calls to `addFirst` and the iterator don't cause this exception.

Ideally, I'd like to see the the mQueuedForMLS list wrapped with a proper class so that we can just push observation points around without worrying about locking semantics.

crankycoder:
Merged in #1603

PR #1603, opened 2015-03-30
1599 synchronization mls locs crash

garvankeeley:
#1599

Synchronize access to the mQueuedForMLS list to avoid concurrentmodificationexception


crankycoder:
r+

PR #1604, opened 2015-03-30
Remove observation point list copy during draw operations

garvankeeley:
If we require the calling thread to lock the observation point list, then the list can be shared between threads. The simplest way to do this is to make the accessor function name clear that this is required, hence getObservationPoints_callerMustLock().

@crankycoder this is possibly a WIP, lets look at this and see if there is a preferred approach we should be using


android/src/main/java/org/mozilla/mozstumbler/client/ObservedLocationsReceiver.java
```diff
@@ -38,7 +40,7 @@
     private static final int MAX_QUEUED_MLS_POINTS_TO_FETCH = 10;
     private static final long FREQ_FETCH_MLS_MS = 5 * 1000;
     private static ObservedLocationsReceiver sInstance;
-    private final LinkedList<ObservationPoint> mCollectionPoints = new LinkedList<ObservationPoint>();
+    private final List<ObservationPoint> mCollectionPoints = Collections.synchronizedList(new LinkedList<ObservationPoint>());
```
crankycoder, line 43: Yay collections.

crankycoder:
I'm not really clear on the benefit of this added concurrency.  That list looks like it's only used in two cases:
- fetching a new MLS data point
- drawing a MLS data point

I think it's better to use a ConcurrentLinkedQueue in this case. We can just push points into that queue and pop them out in FIFO order.  If a MLS fetch is required, we can process the MLS fetch for the point and push the point back into the queue for drawing at a later point in time.

If no MLS fetch is required, we can just process the draw operation and discard the dequeued observation.  This would eliminate all the locking we need.

garvankeeley:
The problem: Every time the observation points are redrawn by a zoom or pan, the list of observation points was being copied for the drawing thread to iterated them. This copy has both performance and memory implications when the app has been collecting for a week and thus has thousands of points.

I can avoid the copy as shown by this PR. I wish there was a nice RAII-way of locking the observation point list when accessed externally ('nice' == not write a bunch of code for little added benefit).

crankycoder:
Ah. Ok.  So assuming that this is fixing some kind of jankiness - let's try it and see if it fixes our immediate problem.

crankycoder:
Can you rebase this?  I can't get ObservedLocationsReceiver to merge in manually.

garvankeeley:
I _could_ profile this and see the effect. 
It does seem like a low hanging fruit, I mean going from copying a list, to not copying a list seems like it can't be any worse :).

crankycoder:
Note to self after talking with @garvankeeley - he hates list copies because of performance. I love list copies because no side effects.  Let's do a performance profile to see who gets to paint the bike shed.

garvankeeley:
Oops, this is actually a necessary fix for ConcurrentModificationException:
http://mozstumbler-crashes.dev.mozaws.net/acralyzer/_design/acralyzer/index.html#/report-details/stumbler/210b8a71-eaeb-4b70-a0dd-4824cb8ea884

garvankeeley:
I need to rebase this, will do that now

android/src/main/java/org/mozilla/mozstumbler/client/ObservedLocationsReceiver.java
```diff
@@ -164,13 +159,13 @@ public void onReceive(Context context, Intent intent) {
 
         synchronized(mCollectionPoints) {
             while (mCollectionPoints.size() > MAX_SIZE_OF_POINT_LISTS) {
-                mCollectionPoints.removeFirst();
+                mCollectionPoints.remove(0);
```
garvankeeley, line 162: I didn't see a way to do LinkedList foo = Collections.synchronized... so I had to change those convenience functions to generic List equivalents.
Let me look again, I am sure there is some way to use the synchronized list as a LinkedList...
Nevermind, all the methods I see require casting, which is sketchy, I'll leave the code as-is.

PR #1606, opened 2015-04-01
Improve map centering button

cascheberg:
The map was always re-centered when the device was rotated, even when the user was panning previously
- make `mFirstLocationFix` and `mUserPanning` static to keep the information when the device is rotated

The centerme button is always shown when the user starts panning, even if there has not been a first location fix; the button initially has the wrong icon
- fix both in `updateCenterButtonIcon`, which replaces `setCenterButtonToNotCenteredIcon`

Rename `sFirstLocationFix` to `sHadFirstLocationFix`. This inverts the meaning, but I think it makes the name more clear to understand.

The follow mode stops when zooming, and when sliding in the sidebar
- change `onTouchEvent` to `onScroll` as this seems more reliable

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java
@@ -10,7 +10,6 @@ import android.content.res.Configuration;
 import android.graphics.Canvas;
 import android.graphics.Paint;
 import android.location.Location;
-import android.location.LocationManager;
 import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
 import android.os.Build;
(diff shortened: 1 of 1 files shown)


android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java
```diff
@@ -83,8 +81,8 @@
     private final Timer mGetUrl = new Timer();
     private MapView mMap;
     private AccuracyCircleOverlay mAccuracyOverlay;
-    private boolean mFirstLocationFix;
-    private boolean mUserPanning = false;
```
crankycoder, line 87: Note to self: sUserPanning should be renamed to "sFollowModeNotEnabled"

leodoni33: Ha6

leodoni33: Hay

PR #1607, opened 2015-04-01
Fix Android Studio MapView rendering problem

cascheberg:
Extend `isInEditMode` checks to other code parts to fix Android Studio's rendering problems regarding the `MapView`.

preview


crankycoder:
Doing a `git diff -w` makes it really obvious that this patch only has an impact when `isInEditMode()` is true.  Much easier than reading the raw diff.

PR #1612, opened 2015-04-02
Fix wifi only observation point

### LINKED ISSUE #1611: wifi-only dots aren't drawn green
This is recent regression, I don't know since when though
I noticed it first in 1.7.11, but it could have been earlier as well:

Wifi-Only reports get green dots instead of purple ones on the map

how I know that this is the case:
the log only shows wifi data but no cell data at all for these reports

cascheberg:
Fixes #1611 
The problem is that an exception is thrown when the cell count is set, so the wifi count is never set. Regression since 1170ed029e91fe6b45c4c30d41e23ae17a828303

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/mapview/ObservationPoint.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/mapview/ObservationPoint.java
@@ -8,6 +8,7 @@ import android.location.Location;
 
 import com.ekito.simpleKML.model.Coordinate;
 
+import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 import org.mozilla.mozstumbler.client.ClientPrefs;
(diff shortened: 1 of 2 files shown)


crankycoder:
Test case please

Djfe:
thx for the fix! ;)
is an NPE the only bad thing that could happen, or was there another reason for the try-catch block?

and wouldn't it be better to fix ichnaeaQueryObj.getJSONArray, so that it doesn't return null in case that there are no cells/wifis available but an empty array?

cascheberg:
@crankycoder Added tests for setting counts in `ObservationPoint`.

@Djfe I did not want to modify the `toMLSGeosubmit` method to always add an empty array because I wasn't sure about negative impacts regarding client-server communication.

Djfe:
@cascheberg Is `ichnaeaQueryObj.getJSONArray` used in `toMLSGeosubmit` as well or why?

android/src/main/java/org/mozilla/mozstumbler/client/mapview/ObservationPoint.java
```diff
@@ -58,12 +59,11 @@ public void setMLSQuery(StumblerBundle stumbleBundle) {
     }
 
     public void setCounts(JSONObject ichnaeaQueryObj) {
-        try {
-            mCellCount = ichnaeaQueryObj.getJSONArray(DataStorageContract.ReportsColumns.CELL).length();
-            mWifiCount = ichnaeaQueryObj.getJSONArray(DataStorageContract.ReportsColumns.WIFI).length();
-        } catch (JSONException ex) {
-            Log.w(LOG_TAG, "Unable to set count length: " + ex.toString());
-        }
+        JSONArray cells = ichnaeaQueryObj.optJSONArray(DataStorageContract.ReportsColumns.CELL);
```
crankycoder, line 62: Nice. I had no idea that `getJSONArray` and `optJSONArray` both exist.

cascheberg:
@Djfe [JSONObject](http://developer.android.com/reference/org/json/JSONObject.html) is provided by the Android API, so you cannot change it's behavior.
But `toMLSGeosubmit` is assembling the structure of the `ichnaeaQueryObj`, so you could modify it to always put an empty `JSONArray` at that index, which I did not want to do.

PR #1616, opened 2015-04-06
Got a non-reproducible NPE in a caller of getTileData(), this guards against it

garvankeeley:
The callers of getTileData() check its length() for validity, so always return a non-null.


crankycoder:
tData can be null if the file is deleted between  or is corrupt.

File missing case :

https://github.com/mozilla/MozStumbler/blob/dev/android/src/main/java/org/mozilla/osmdroid/tileprovider/modules/SerializableTile.java#L46

Corruption case :
https://github.com/mozilla/MozStumbler/blob/dev/android/src/main/java/org/mozilla/osmdroid/tileprovider/modules/SerializableTile.java#L265

PR #1624, opened 2015-04-15
bn-IN values add at res

shiladittyapaul:


garvankeeley:
Thanks for the translation!

PR #1625, opened 2015-04-16
The context-less access to NetworkInfo singleton is error-prone, and not needed

garvankeeley:
Motivated by fixes for this NPE crash:
https://bugzilla.mozilla.org/show_bug.cgi?id=1155237


garvankeeley:
I broke tests. Will get them in-line with the changes

android/src/main/java/org/mozilla/mozstumbler/client/ObservedLocationsReceiver.java
```diff
@@ -100,8 +103,9 @@ public void run() {
                 Iterator<ObservationPoint> li = queued.iterator();
                 while (li.hasNext() && count < MAX_QUEUED_MLS_POINTS_TO_FETCH) {
                     ObservationPoint obs = li.next();
-                    if (obs.needsToFetchMLS()) {
-                        obs.fetchMLS();
+                    if (obs.needsToFetchMLS() && mContext != null) {
```
crankycoder, line 106: I don't think this can actually be null because of the guard at L87. That said, I trust android not at all, so an extra guard isn't going to hurt.

PR #1635, opened 2015-05-03
Fix typo

stripTM:


lua86:
Yep. It is true that is the correction For such typo. Hard not to make typos on this smartphone.

garvankeeley:
Thanks for the fix!

PR #1642, opened 2015-05-20
Features/1640 fdroid is crazy

crankycoder:
I can't reproduce this bug at all.  I suspect the error on ichnaea is someone mangling our MLS code and re-using the fdroid supplied API key.

In anycase, this PR adds a check for filepaths and just short circuits. I do not unpack the gzip data because that's pretty CPU intensive and I worry about battery life.


crankycoder:
Ugh. I screwed this PR up

Issue #1640, opened 2015-05-20
F-droid build uploading broken body


hannosch:
I've gotten a couple error reports on the service side for our f-droid build uploading a broken body. Instead of a proper gzip encoded body or a JSON payload, it sends the literal value `'/data/data/org.fdroid.fdroid/files/fdroid/repo/../../index.html'` as the body.

These are requests against https://location.services.mozilla.com/v1/geosubmit?key=... with our f-droid api key.

PR #1643, opened 2015-05-20
Fix for #1640

crankycoder:
This is fix for #1640. I can't reproduce this bug at all, so this is just a guard and a testcase exercising the guard.

Note that I do not unpack the gzip data as that would be very CPU intensive.


garvankeeley:
r+

PR #1648, opened 2015-05-26
Fix: passive mode battery checker never starts

cascheberg:
The `mPassiveModeBatteryChecker` is never started. I assume that is a bug and regression from 6b8dc2d19b129d4dbdbca18b2d20075c7aeefd9f.


garvankeeley:
Thanks again @cascheberg!

PR #1655, opened 2015-05-29
Fix significant motion sensor initial state

cascheberg:
The initial state of the significant motion sensor is active. I fixed that and inverted the semantic of the variable to hopefully make it more comprehensible.
This is really a small bug, I only noticed it because the false positive motion filter was triggered although I had motion sensors disabled.


cascheberg:
Update: added some tests. The first added `assert` will fail without this fix.

crankycoder:
Sorry @cascheberg - we were away at a conference the last little while.  Thanks for the patch!

PR #1663, opened 2015-06-09
Add position following in passive mode

cascheberg:
Add position following for passive mode which can be enabled for testing in developer options


garvankeeley:
Sorry for the wait. Looks good, thanks!

PR #1670, opened 2015-06-29
Fix unlimited zoom

### LINKED ISSUE #1481: Limited zoom feature problems
@garvankeeley there are some more problems with #1463 / #1474 
- zooming out on low-res map becomes blank
- both res types do not show the MLS coverage anymore when zooming out too far
- I think the description text is not quite accurate. I suggest something like: zooming out can increase data usage (instead of conveying that the feature generally reduces data usage)

### LINKED ISSUE #1506: unlimited map zoom bug (tiles not visible)
we don't get MLS tiles above the limited zoom level (pls fix this) (I just see the map without the coverage tiles

-on the low resolution mode you don't see any map at all at a certain zoom level (just the circles on the map)

cascheberg:
First commit:
- cleanup some code
- make the full unlimited zoom mode apply only in high res mode

Second commit:
- add an additional coverage overlay that is used if zoom is lower than the current minimum coverage zoom level

Fixes #1506, fixes #1481

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java
@@ -72,7 +72,7 @@ public class MapFragment extends android.support.v4.app.Fragment
 
     private static final String COVERAGE_REDIRECT_URL = "https://location.services.mozilla.com/map.json";
     private static final String ZOOM_KEY = "zoom";
-    private static final int LOWEST_UNLIMITED_ZOOM = 3;
+    public static final int LOWEST_UNLIMITED_ZOOM = 3;
     private static final int DEFAULT_ZOOM = 13;
     private static final int DEFAULT_ZOOM_AFTER_FIX = 16;
     private static final String LAT_KEY = "latitude";
@@ -90,6 +90,7 @@ public class MapFragment extends android.support.v4.app.Fragment
     private LowResMapOverlay mLowResMapOverlayLowZoom;
     private Overlay mCoverageTilesOverlayLowZoom;
     private Overlay mCoverageTilesOverlayHighZoom;
+    private Overlay mCoverageTilesOverlayOriginalZoom; // used in unlimited zoom mode for levels < LOW_ZOOM_LEVEL
     private ITileSource mHighResMapSource;
     private View mRootView;
     private TextView mTextViewMapResolutionInfo;
@@ -263,9 +264,11 @@ public class MapFragment extends android.support.v4.app.Fragment
 
     private void initCoverageTiles(String coverageUrl) {
         ClientLog.i(LOG_TAG, "initCoverageTiles: " + coverageUrl);
-        mCoverageTilesOverlayLowZoom = new CoverageOverlay(CoverageOverlay.LowResType.LOWER_ZOOM,
+        mCoverageTilesOverlayLowZoom = new CoverageOverlay(AbstractMapOverlay.TileResType.LOWER_ZOOM,
                 mRootView.getContext(), coverageUrl, mMap);
-        mCoverageTilesOverlayHighZoom = new CoverageOverlay(CoverageOverlay.LowResType.HIGHER_ZOOM,
+        mCoverageTilesOverlayHighZoom = new CoverageOverlay(AbstractMapOverlay.TileResType.HIGHER_ZOOM,
+                mRootView.getContext(), coverageUrl, mMap);
+        mCoverageTilesOverlayOriginalZoom = new CoverageOverlay(AbstractMapOverlay.TileResType.ORIGINAL_ZOOM,
                 mRootView.getContext(), coverageUrl, mMap);
     }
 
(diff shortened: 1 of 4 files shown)


android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java
```diff
@@ -90,6 +90,7 @@
     private LowResMapOverlay mLowResMapOverlayLowZoom;
     private Overlay mCoverageTilesOverlayLowZoom;
     private Overlay mCoverageTilesOverlayHighZoom;
+    private Overlay mCoverageTilesOverlayDefaultZoom;
```
garvankeeley: Hmm, not sure what to call this, but saying 'default' here is confusing. Unfortunately, "mCoverageTilesOverlayUnlimitedModeAtLowZoom" is the only thing I can think of :) that makes the purpose of this totally clear.

android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java
```diff
@@ -267,6 +268,7 @@ private void initCoverageTiles(String coverageUrl) {
                 mRootView.getContext(), coverageUrl, mMap);
         mCoverageTilesOverlayHighZoom = new CoverageOverlay(CoverageOverlay.LowResType.HIGHER_ZOOM,
                 mRootView.getContext(), coverageUrl, mMap);
+        mCoverageTilesOverlayDefaultZoom = new CoverageOverlay(mRootView.getContext(), coverageUrl, mMap);
```
garvankeeley: Instead of another constructor, CoverageOverlay.TileResType.UNMODIFIED_ZOOM or ORIGINAL_ZOOM

android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java
```diff
@@ -287,35 +289,37 @@ private void updateOverlayBaseLayer(int zoomLevel) {
...
     private void updateOverlayCoverageLayer(int zoomLevel) {
-        if (mCoverageTilesOverlayLowZoom == null || mCoverageTilesOverlayHighZoom == null) {
-            return;
-        }
-
         final List<Overlay> overlays = mMap.getOverlays();
-        int idx = 0;
-        if (overlays.indexOf(mLowResMapOverlayHighZoom) > -1 || overlays.indexOf(mLowResMapOverlayLowZoom) > -1) {
-            idx = 1;
+        if (mCoverageTilesOverlayLowZoom == null || mCoverageTilesOverlayHighZoom == null
```
garvankeeley, line 304: The precondition of this function should be the first line:
if (!isCoverageTilesInitialized()) {
  return;
}

android/src/main/java/org/mozilla/mozstumbler/client/mapview/tiles/CoverageOverlay.java
```diff
@@ -16,18 +16,27 @@
  * This class provides the Mozilla Coverage overlay
  */
 public class CoverageOverlay extends AbstractMapOverlay {
+    public static final String MLS_MAP_TILE_COVERAGE_NAME = "Mozilla Location Service Coverage Map";
     // Use a lower zoom than the LowResMapOverlay, the coverage can be very low detail and still look ok
-    private static final int LOW_ZOOM_LEVEL = 10;
+    public static final int LOW_ZOOM_LEVEL = 10;
+
+    public CoverageOverlay(final Context aContext, final String coverageUrl, MapView mapView) {
```
garvankeeley: See comment about changing the type to avoid another constructor.

android/src/main/java/org/mozilla/mozstumbler/client/mapview/tiles/CoverageOverlay.java
```diff
@@ -16,18 +16,27 @@
...
+        super(aContext);
+        setup(1, AbstractMapOverlay.LARGE_SCREEN_MIN_ZOOM, coverageUrl, mapView);
```
garvankeeley: Is the '1' correct?

android/src/main/java/org/mozilla/mozstumbler/client/mapview/tiles/CoverageOverlay.java
```diff
@@ -16,18 +16,27 @@
...
     public CoverageOverlay(LowResType type, final Context aContext, final String coverageUrl, MapView mapView) {
         super(aContext);
 
         final int zoomLevel = (type == LowResType.HIGHER_ZOOM) ?
                 AbstractMapOverlay.getDisplaySizeBasedMinZoomLevel() : LOW_ZOOM_LEVEL;
+        setup(zoomLevel, zoomLevel, coverageUrl, mapView);
+    }
 
-        final ITileSource coverageTileSource = new XYTileSource("Mozilla Location Service Coverage Map",
+    private void setup(int aZoomMinLevel, int aZoomMaxLevel, final String coverageUrl, MapView mapView) {
```
garvankeeley: no 'a' prefix (that is a Gecko C++ standard that we sometimes add accidentally AFAICT)

garvankeeley:
r+ after nits addressed, thanks for fixing this!

android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java
cascheberg: How about "OriginalZoom" and a comment?

garvankeeley: Sounds good.

android/src/main/java/org/mozilla/mozstumbler/client/mapview/tiles/CoverageOverlay.java
cascheberg: I just noticed that [zoom level 0](https://d17pt8qph6ncyq.cloudfront.net/tiles/0/0/0.png) is used as an image on the MLS website frontpage. But it is not available currently, and it wasn't when I checked before. Just a temporary server side issue?

garvankeeley: I don't know, but I can't see needing it, isn't the unlimited zoom min level 3?

^^^ hold on, I am misreading the arguments

^^^^ no, I am not, I was expecting the setup values to be (3, 13), although using constants for these of course.

(I should re-read the code, I thought unlimited zoom was still limited to greater than 1 or 2)

cascheberg: It's true that zoom levels lower than 3 will not be used currently. I was thinking that these parameters should just reflect the actual server side capability anyway (similar to `mHighResMapSource`).

garvankeeley: Fair enough; It seems self-documenting as to how it will be used if those values are tied to zoom constants in the codebase, which I like as a reader. 
Aside: just noticed, isn't AbstractMapOverlay.LARGE_SCREEN_MIN_ZOOM too high for how this is used?

cascheberg: Yes, the relevant range is only from `LOWEST_UNLIMITED_ZOOM` to `LOW_ZOOM_LEVEL - 1` (3 to 9). But as I said above, restricting the range at this part of the code seemed unnecessary to me. But I can change that if it improves code readability.

garvankeeley: I vote yes, since when I look at in future, I'll know exactly for what zoom levels this is active from those values. Arguably, that is also in MapFragment, but that info is hard to see at a glance there.

cascheberg:
@garvankeeley I updated the second commit to address your review comments.

PR #1676, opened 2015-07-08
added guards around the leaderboard

crankycoder:
I added a couple guards to prevent the leaderboard code from storing or uploading records if FxA isn't logged in.  This was causing me confusion while testing the TLS code - mostly because the  leaderboard server isn't running and stumbler was spewing HTTP POST errors.


garvankeeley:
this commit is missing files

garvankeeley:
not collecting this info when not logged in seems a little weird, because the login is only required at submit time. However we don't have any other clear,simple way of not collecting leaderboard info if the user has no interest in it. So lgtm.

Issue #490, opened 2014-03-12
Make it possible to export the gps trace only to Open Street Map ( OSM )


lhirlimann:
I would love to have a button that would let me submit my gps traces to OSM.

Djfe:
Part of #3.
Make your suggestions there ;)

crankycoder:
OpenStreetMaps has a very minimal requirement - we just need to be able to export GPX tracks:

http://wiki.openstreetmap.org/wiki/GPX

rugk:
Yes I also like this idea.
It would be nice if you could export the GPX files not only for OSM, but also for other uses.
E.g. you can later analyse where you walked or drove with any website or app which can read GPX files. This way you don't need another app running if you just want to view your route later.

cascheberg:
Currently only locations that contain at least one wifi or cell observation are recorded. For the GPX export it would be much better to be able to export all scanned locations. Would it be ok to change that, and filter empty reports at a different place within the client?

Djfe:
it would also make sense to save gps coordinates, cell scans and wifi scans differently (in general).
accuracy could be increased drastically.
right now the next report gets the wifi/cell reports even though they might be more fitting for the last report or something like that
All three scans would have a timestamp and the cell positions and wifi positions would be interpolated over the gps positions. (roughly one wifi scan would be interpolated over 5 gps positions)
this especially makes sense since the scans aren't synchronized afaik
not sure if everything below here is possible, but I would think that it is and it doesn't look accurate to me :/

```
0 secs gps scan
0,5 secs cell scan
1,5 secs cell scan
2,5 secs cell scan
3,5 secs gps scan (moved 30meters limitation)
4 secs wifi scan (started at -1 secs)
4,5 secs cell and gps scan
5,5 secs cell scan
6 secs gps scan ( driving 30 meters took longer than 1 sec again)
...
8 secs gps scan
8,5 secs cell scan
9 secs wifi scan
...
11 secs next gps scan (took longer again)
```

For the implementation accurate timestamps would be needed again (only on client side, one month accuracy isn't enough obviously) to be able to get a rough postion of the client while the wifi/cell scan has been done
(maybe it's helpful to get the speed the gps chip reports  for this in addition to the average speed we get from dividing the distance through the time)

by doing the interpolation we would get the reports and can obfuscate the timestamps there

(for the osm export I would say daily timestamps would make more sense than monthly so we would need more accurate timestamps anyways)

then we have the reports and can save them in the storage

[...]

PR #1689, opened 2015-07-19
Improve KML export

cascheberg:
This fixes a bug in the KML export and hopefully cleans up some code.
1. The heading for an observation was computed from the current position to the previous one, which is in the wrong direction. The heading icon specified in the KML file was showing downwards at 0 heading, which visually compensated this bug.
2. I switched to use the `Location` class instead of `GeoPoint` in the `ObservationPoint` to store more of the original information (i.e. accuracy, speed, `hasAltitude`). This can later be used for GPX export (#490).

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/ObservedLocationsReceiver.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/ObservedLocationsReceiver.java
@@ -24,7 +24,6 @@ import org.mozilla.mozstumbler.service.utils.NetworkInfo;
 import org.mozilla.mozstumbler.svclocator.ServiceLocator;
 import org.mozilla.mozstumbler.svclocator.services.log.ILogger;
 import org.mozilla.mozstumbler.svclocator.services.log.LoggerUtil;
-import org.mozilla.osmdroid.util.GeoPoint;
 
 import java.lang.ref.WeakReference;
 import java.util.Collections;
@@ -150,7 +149,7 @@ public class ObservedLocationsReceiver extends BroadcastReceiver {
         if (position == null) {
             return;
         }
-        ObservationPoint observation = new ObservationPoint(new GeoPoint(position));
+        ObservationPoint observation = new ObservationPoint(position);
 
         try {
             observation.setCounts(bundle.toMLSGeosubmit());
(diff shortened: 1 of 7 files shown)


crankycoder:
Thanks @cascheberg!

Can you add a specification for the expected KML output from Stumbler?  I'll admit that I rarely use the KML output myself, but @garvankeeley uses it pretty frequently.  Once there's a spec, I can whip up a small testcase for verification of correctness.

cascheberg:
Hm I did not modify the KML output format, and I barely know the format myself. Are you thinking of a scheme file? Or can you just use some manually generated data to create an output file once, and then use that as a fixed reference for comparison during tests?

crankycoder:
Ah - sorry. I thought you added some extra fields to the KML output.

I'll do a quick stumble tonight to get some KML data that we can use in a testcase.  I'll do a quick review of this tonight.

android/src/main/java/org/mozilla/mozstumbler/client/serialize/ObservationPointSerializer.java
```diff
@@ -48,7 +49,7 @@
     private static final String MLS_NAME = "MLS";
     private static final String ICON_RED_CIRCLE = "http://maps.google.com/mapfiles/kml/shapes/placemark_circle_highlight.png";
     private static final String STYLE_NAME_RED_CIRCLE = "redcircle";
-    private static final String ICON_ARROW = "http://maps.google.com/mapfiles/kml/shapes/arrow.png";
+    private static final String ICON_ARROW = "http://earth.google.com/images/kml-icons/track-directional/track-0.png";
```
crankycoder, line 52: Hrm.  These links are probably not ok.  We probably don't want to be loading images from the google domain at all.

cascheberg, line 52: I don't know if we should use those. MozStumbler does not load these icons. The only other app I used was Google Earth, which does change the icon based on these links, but it already has them available locally. It could even do some magic:

> Note: The track icons are treated special in Google Earth. If you specify to use the gx:Track extension as used for GPS-tracks and specify a Style using the http://earth.google.com/images/kml-icons/track-directional/track-0.png icon (shown above) then the icon is dynamically changed to appropriate icon using the computed compass heading from the previous position.

I found the information and icon URL here: http://kml4earth.appspot.com/icons.html

crankycoder, line 52: Cool. Thanks for double checking this!  :)

android/src/main/java/org/mozilla/mozstumbler/client/ObservedLocationsReceiver.java
```diff
@@ -172,8 +171,9 @@ public void onReceive(Context context, Intent intent) {
                 mCollectionPoints.remove(0);
             }
 
-            if (mCollectionPoints.size() > 0) {
-                observation.mHeading = observation.pointGPS.bearingTo(mCollectionPoints.get(mCollectionPoints.size() - 1).pointGPS);
+            if (mCollectionPoints.size() > 0 && !observation.pointGPS.hasBearing()) {
+                final Location previous = mCollectionPoints.get(mCollectionPoints.size() - 1).pointGPS;
+                observation.pointGPS.setBearing(previous.bearingTo(observation.pointGPS));
```
crankycoder, line 176: This is much smarter. Nice work.

crankycoder:
This works great for me.  Thanks!

PR #1695, opened 2015-07-26
Export as GPX file

### LINKED ISSUE #490: Make it possible to export the gps trace only to Open Street Map ( OSM )

cascheberg:
Fixes #490. The code is based on the KML export code, but without the ability to load a file.


garvankeeley:
Looks great, thanks @cascheberg

PR #1704, opened 2015-08-08
Allow exporting points that have no radio data

cascheberg:
GPX tracks (#490) are incomplete when the path contains observations without radio data.
The second commit modifies the `Reporter` to also send intents for observations without radio data, but without counting or adding them to the `DataStorageManager`. They are also filtered in the client so that they are not displayed on the map.

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/serialize/ObservationPointSerializer.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/serialize/ObservationPointSerializer.java
@@ -52,9 +52,10 @@ public class ObservationPointSerializer extends AsyncTask<Void, Void, Boolean> {
     private static final String STYLE_NAME_RED_CIRCLE = "redcircle";
     private static final String ICON_ARROW = "http://earth.google.com/images/kml-icons/track-directional/track-0.png";
     private static final String STYLE_NAME_ARROW = "arrow";
-    private static final String COLOR_HAS_WIFI = "ffff0000"; //green in AABBGGRR
-    private static final String COLOR_HAS_CELLS = "ff00ffff"; // yellow
-    private static final String COLOR_HAS_BOTH = "ff00ff00"; // blue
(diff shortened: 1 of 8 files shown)


garvankeeley:
vng, I am reviewing this (so we don't cross streams)

garvankeeley:
This is a very clean approach, I like this. There is some additional in-memory data storage pressure as ObservedLocationsReceiver.mCollectionPoints will get these data points, but I think it is ok.

android/src/main/java/org/mozilla/mozstumbler/client/serialize/ObservationPointSerializer.java
```diff
@@ -52,9 +52,10 @@
     private static final String STYLE_NAME_RED_CIRCLE = "redcircle";
     private static final String ICON_ARROW = "http://earth.google.com/images/kml-icons/track-directional/track-0.png";
     private static final String STYLE_NAME_ARROW = "arrow";
-    private static final String COLOR_HAS_WIFI = "ffff0000"; //green in AABBGGRR
-    private static final String COLOR_HAS_CELLS = "ff00ffff"; // yellow
-    private static final String COLOR_HAS_BOTH = "ff00ff00"; // blue
+    private static final String COLOR_HAS_WIFI = "ff0000ff"; // red in AABBGGRR
+    private static final String COLOR_HAS_CELLS = "ffff0000"; // blue
+    private static final String COLOR_HAS_BOTH = "ff00ff00"; // green
+    private static final String COLOR_HAS_NONE = "aa000000"; // transparent black
```
hannosch, line 58: This changes the colors of the on-map points, right? Do we have a legend for these that would need to get updated? Or should we make a note to mention this in the news section for the next play store release?

garvankeeley:
This is ok, this is the color of the exported points

hannosch:
Ah ok, good :)

PR #1706, opened 2015-08-17
Simple high-power mode

### LINKED ISSUE #1386: add explicit high-power mode
There are a lot of [efforts to limit power consumption](https://github.com/mozilla/MozStumbler/labels/Power%20Consumption) to "reasonable" values.
However, there are situations where I just don't care and I'd rather find more networks than saving power: Driving fast with a 5s scan interval means that you miss most WiFi networks.

My wish for an option:
-  [x] smart tradeoff: power vs. thoroughness
-  [ ] high-power mode if charging
-  [ ] high-power mode

We are currently locked in the first option. It would probably make sense to default to the second option or at least offer it.

High-power mode could dispense with motion detection and increase the scanning frequency. I find both mostly annoying, because motion detection only works to detect the _end_ of motion; I then have to start scanning manually. The low scanning frequency does only make sense if you are walking. Even cycling is too fast: 5s · 20 km/h = 28m

cascheberg:
Basically fixes #1386. It should also increase the accuracy of exported GPS tracks (#490) when enabled.

Effects:
- Allow scans every 15m (instead of 30m), but still only every 2s (~27km/h)
- Scan for WiFi every 2s (instead of 4s)
- Allow max of 3 WiFi scans per GPS position (instead of 2)

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/subactivities/DeveloperActivity.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/subactivities/DeveloperActivity.java
@@ -38,7 +38,6 @@ import org.mozilla.mozstumbler.service.stumblerthread.motiondetection.MotionSens
 import org.mozilla.mozstumbler.service.utils.BatteryCheckReceiver;
 import org.mozilla.mozstumbler.svclocator.ServiceLocator;
 import org.mozilla.mozstumbler.svclocator.services.log.LoggerUtil;
-import org.mozilla.osmdroid.tileprovider.modules.SerializableTile;
 
 import java.io.File;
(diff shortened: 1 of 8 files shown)


garvankeeley:
Thanks! I think this is useful as a developer option.

There is some info here on Wifi scan freq: http://stackoverflow.com/questions/7969174/reducing-delay-between-two-wifi-scans
One of the comments states that some devices do allow scanning every 1-2 seconds. I mention that because I have not seen wifi APs changed that frequently on my Nexus 5, so it is useful to note that it does seem possible to get more frequent AP scans.

I'll be interested in the change in power consumption with this feature turned on.

PR #1709, opened 2015-08-23
Move unlimited zoom pref to settings

cascheberg:
See #1474 / #1491

Should we note that zoom is still limited in low-res map mode?


garvankeeley:
Perhaps it could be:
<code>
[ ] Unlimited map zoom in high detail mode

```
Low detail is always limited to reduce network usage.
```

cascheberg:
@garvankeeley Your suggested title is a bit too long. But I like the idea of inverting the semantics.

My suggestion in the followup commit:

> [ X ] Unlimited map zoom
> Limiting zoom can help to reduce the network data used by the map view. Unlimited zoom only applies to high detail map mode.

garvankeeley:
Oops forgot to land this, thanks!

PR #1710, opened 2015-08-26
getNeighboringCellInfo() is deprecated on Android M.

garvankeeley:
On Android M, don't call this function, and on older OS's keep using it.

https://bugzilla.mozilla.org/show_bug.cgi?id=1195854


crankycoder:
Ugh. Travis has it's own SDK thing.  I'll switch this over to use a 'real' Android SDK instead of whatever travis uses.  This should at least get rid of a mismatch between build enviroments with real developers vs the CI system.

crankycoder:
This patch was modified in fennec https://hg.mozilla.org/mozilla-central/rev/0c92aae1b559 

I'm recreating a patch based on what Fennec actually accepted.

PR #1728, opened 2015-10-10
Fix zoom

cascheberg:
In some cases, especially when using pinch to zoom, the coverage overlay mode is not updated.

There is a `MapListener` in [MapFragment.java#L171](https://github.com/mozilla/MozStumbler/blob/dev/android/src/main/java/org/mozilla/mozstumbler/client/mapview/MapFragment.java#L171) to trigger `updateOverlayCoverageLayer`. 
But the `DelayedMapListener` will sometimes cancel a zoom event when dispatching a scroll event. So as a result the zoom listener will not run.

Using separate `Runnables` in `DelayedMapListener` fixes the problem.

Not using `DelayedMapListener` also fixes the problem, but I do not know if that is somehow otherwise problematic. The code was introduced in d8e1faae9f88c269d41fd759d561b42789fa192f.

### DIFF
--- a/android/src/main/java/org/mozilla/osmdroid/events/DelayedMapListener.java
+++ b/android/src/main/java/org/mozilla/osmdroid/events/DelayedMapListener.java
@@ -25,7 +25,8 @@ public class DelayedMapListener implements MapListener {
      */
     protected long delay;
     protected Handler handler;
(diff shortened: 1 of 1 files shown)


crankycoder:
Nice catch @cascheberg !

PR #1754, opened 2016-04-07
Adds millisec timestamp from epoch for GPS and Wifi

crankycoder:
- add timestamps for wifi on geosubmit
- dropped unused getLocation methods
- added a proper timestamp fixture into the geosubmit test

Both timestamps are millisecond accuracy from the epoch.  The top level timestamp for the geosubmit blob is extracted from the GPS scan.

### DIFF
--- a/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/datahandling/StumblerBundle.java
+++ b/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/datahandling/StumblerBundle.java
@@ -105,7 +105,6 @@ public final class StumblerBundle implements Parcelable {
 
     @TargetApi(Build.VERSION_CODES.JELLY_BEAN_MR1)
     public MLSJSONObject toMLSGeosubmit() throws JSONException {
-        ISystemClock clock = (ISystemClock) ServiceLocator.getInstance().getService(ISystemClock.class);
         MLSJSONObject headerFields = new MLSJSONObject();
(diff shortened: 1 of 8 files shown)


libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/datahandling/StumblerBundle.java
```diff
@@ -98,7 +103,9 @@ public boolean hasRadioData() {
         return Collections.unmodifiableMap(mCellData);
     }
 
+    @TargetApi(Build.VERSION_CODES.JELLY_BEAN_MR1)
     public MLSJSONObject toMLSGeosubmit() throws JSONException {
+        ISystemClock clock = (ISystemClock) ServiceLocator.getInstance().getService(ISystemClock.class);
```
hannosch: This looks like it isn't used in the modified code?

libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/datahandling/StumblerBundle.java
```diff
@@ -127,15 +134,20 @@ public MLSJSONObject toMLSGeosubmit() throws JSONException {
...
+                if (scanResult.frequency != 0) {
+                    wifiEntry.put("frequency", scanResult.frequency);
                 }
-                if (s.level != 0) {
-                    wifiEntry.put("signalStrength", s.level);
+                if (scanResult.level != 0) {
+                    wifiEntry.put("signalStrength", scanResult.level);
                 }
+
+                long gpsTimeSinceBootInMS = (mGpsPosition.getElapsedRealtimeNanos()/1000000);
```
hannosch: I think you could move this outside the for loop.

PR #1755, opened 2016-04-08
Features/gps speed heading

crankycoder:
This adds float fields for GPS Speed, GPS Heading and pressure sensors to the geosubmit upload.  

A separate commit will handle the whitespace and import optimizations as there are lots of orphaned imports across the codebase.


hannosch:
r+, looks good to me.

PR #1756, opened 2016-04-09
Update LB_SUBMIT_URL

hannosch:
For a release build, this should probably point to the real thing? Or should this config go into the product flavors lbdev/stage/prod sections?


crankycoder:
Crap.  I need to drop that LB_SUBMIT_URL entirely. We don't use it anymore.

PR #1768, opened 2016-04-14
added code to disable IP based fallbacks when doing geolocate

### LINKED ISSUE #1751: Disable geoip fallback in map view queries
We got a couple of reports about confusing results of the "Show MLS position" option on the map.

One thing we could do, is to explicitly disable the GeoIP fallback in the MLS queries we use to get those results. Instead of getting a lot of lines that all go to the same far away place, we'd get no lines at all for these. At least that should add less visual clutter on the map.

I think all we'd have to do is extend `StumblerBundle.toMLSGeolocate` (https://github.com/mozilla/MozStumbler/blob/dev/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/stumblerthread/datahandling/StumblerBundle.java#L154) to do this.

The fallback options are another top-level JSONObject with values of:

```
    "fallbacks": {
        "lacf": false,
        "ipf": false
    }
```

The full example at https://mozilla.github.io/ichnaea/api/geolocate.html shows this.

The `ipf` disables GeoIP (IP-based fallback). I'd also disable `lacf` (cell location area based fallback), as area fallback also means we don't have information about the concrete cell observed at the given point.

In addition we probably need to change `AsyncGeolocate.doInBackground` (https://github.com/mozilla/MozStumbler/blob/dev/android/src/main/java/org/mozilla/mozstumbler/client/mapview/AsyncGeolocate.java#L75) and instead of:

```
if (resp.isErrorCode400BadRequest()) {
```

do

```
if (resp.isErrorCode4xx()) {
```

Otherwise not found responses (404) would contain a valid JSON body, but not one that has any coordinate fields, and thus would break. By treating 'not found' as a kind of error, the logic should simply stop trying to do a MLS query for the particular observation point.

@crankycoder what do you think?

crankycoder:
This fixes #1751

### DIFF
--- a/android/src/main/java/org/mozilla/mozstumbler/client/mapview/AsyncGeolocate.java
+++ b/android/src/main/java/org/mozilla/mozstumbler/client/mapview/AsyncGeolocate.java
@@ -72,8 +72,7 @@ public class AsyncGeolocate extends AsyncTask<String, Void, Location> {
             return null;
         }
 
-        if (resp.isErrorCode400BadRequest()) {
-            //TODO detect malformed request, and clear out mlsrequest on observation point
+        if (resp.isErrorCode4xx()) {
             mIsBadRequest = true;
             return null;
         }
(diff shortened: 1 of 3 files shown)


hannosch:
Looks like a real test failure, since the expected output changed in https://github.com/mozilla/MozStumbler/blob/dev/libraries/stumbler/src/test/java/org/mozilla/mozstumbler/service/stumblerthread/datahandling/StumblerBundleTest.java#L69

hannosch:
r+, minus the test failure

PR #1769, opened 2016-04-14
dropped the clickbox from the metrics view to enable MLS queries

crankycoder:
This fixes a long standing UI nit that's been bothering me.  

When you enable MLS queries in the settings screen, you also have to find the little checkbox at the bottom of the MetricsView; the pop out drawer; and turn that on as well.

I think we had this split initially when we were just testing the MLS query ability before we had a way to render the locations.

This no longer makes any sense - if users turn on the MLS query setting in the preferences - we should just render the layer.


hannosch:
looks ok to me, though I can't really tell, since this is all weird Android UI code :)

PR #1771, opened 2016-04-15
Fix some issues

cascheberg:
- `invalidateOptionsMenu` requires higher API level
- fix string resources


android/src/main/java/org/mozilla/mozstumbler/client/navdrawer/MainDrawerActivity.java
```diff
@@ -333,7 +333,7 @@ public void updateUiOnMainThread(final boolean updateMetrics) {
             public void run() {
 
                 // Force the leaderboard menu item to update it's enabled status
-                invalidateOptionsMenu();
+                supportInvalidateOptionsMenu();
```
crankycoder, line 336: Nice!  Thanks for catching this one.

Issue #1779, opened 2016-04-20
appcompat-v4:23 and appcompat-v7:23 GUI glitches


crankycoder:
The change to use appcompat at SDK level 23 has broken some of the UI.

Things I've noticed which are not 'broken' but just different:
- Initial welcome screen 
- Triple button over the map view now renders the "View Leaderboard" inline rather than below
- PrefencesScreen is now full screen and rendering has been modified slightly.  All UI controls have been updated.

Broken things:
- Power saving button is now whited out in the popup drawer

We need to update the button to not be whited out and verify this on at least a couple versions of Android.  I'm running Android 5.1.1 and have a Nexus 5 device running Android 6 that I can test on.

cascheberg:
There are more things:
- button for showing the popup drawer is now a back button, stumbler icon is missing there
- on/off button is different, not labeled

What was the reason for the update? Btw, the versions still don't match in `android/build.gradle`.

crankycoder:
Urgh. Thanks @cascheberg .  I've got a branch over here which at least partially fixes the 'Change' button for power management.

I've sync'd up the `android/build.gradle` file to use appcompat SDK 23 versions now. 

The driver for these changes is F-Droid.  The builds were failing over there and the only way I could find to make everything happy again was to update appcompat to SDK 23.

I'm going to start working on the popup drawer button that has turned into a back button.  I'm not terribly concerned about the on/off button as the styling is just inherited from the main appcompat theme for Lollipop.

crankycoder:
Oops!  Forgot to add the link to the branch I'm working in: https://github.com/crankycoder/MozStumbler/tree/features/1779-appcompat-ui-fix

crankycoder:
This is 'fixed' in #1782 - but a lot of the UI bits need to be reworked for Material Design.  I'll leave the Material Design changes up to a separate bug report, but for now - we haven't regressed beyond some sizing changes to the actionbar and the power management button.

PR #1785, opened 2016-04-28
Fix power settings button alignment

cascheberg:
@crankycoder The alignment of the power settings button is not as nice as it was before (#1779).
This PR fixes it for me, see screenshots.

**Old:**
stumbler_1

**Current:**
stumbler_2

**Fixed:**
stumbler_3


crankycoder:
Thanks!   I couldn't get the XML just right for the layout before.

Issue #1614, opened 2015-04-06
Use system date format rather than locale date format


ghost:
Android allows you to specify the date format in the settings, but Mozilla Stumbler seems to just use the locale-specified date format instead.

You can get the system date format this way ([source](http://stackoverflow.com/a/6982154/492203)):

``` java
final String format = Settings.System.getString(getContentResolver(), Settings.System.DATE_FORMAT);

if (TextUtils.isEmpty(format)) {
    dateFormat = android.text.format.DateFormat.getMediumDateFormat(getApplicationContext());
} else {
    dateFormat = new SimpleDateFormat(format);
}
```

gcsdev:
I tried solve it in https://github.com/mozilla/MozStumbler/pull/1809

PR #1808, opened 2017-01-06
Merge pull request #1614 from gcsdev-patch

gcsdev:
Using system date format rather than locale date format.


gcsdev:
The first commit presented one error and the second commit is a fix it.

PR #1812, opened 2017-03-06
Added Hungarian translation

meskobalazs:
I made the Hungarian translation file for the project.


kolesar-andras:
What a great surprise! I have joined today from Hungary. I'm glad you have translated the application.

PR #1818, opened 2017-05-12
Fix #1800: Open-source builds use non-existent mapquest tiles, use MapBox instead

### LINKED ISSUE #1800: MapQuest does not provide the OpenStreetMap-Tiles anylonger
MapQuest discontinued their free/anonymous access to their OpenStreetMap-tiles. This means that Mozilla Stumbler will always show a grey grid rather than the OpenStreetMap.

The only free plan by MapQuest only includes 15,000 requests per month. Since this is not enough for Mozilla Stumbler, a new (maybe Mozilla-hosted?) tiles provider is required.

As @lua86 mentioned: It is also possible not to use tiles at all and render the raw map data live.

See also:
http://devblog.mapquest.com/2016/06/15/modernization-of-mapquest-results-in-changes-to-open-tile-access/

garvankeeley:
screenshot:
screen shot 2017-05-12 at 4 01 53 pm

### DIFF
--- a/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/AppGlobals.java
+++ b/libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/AppGlobals.java
@@ -10,6 +10,15 @@ import java.util.concurrent.ConcurrentLinkedQueue;
 
 public class AppGlobals {
 
+    public static final String MAPBOX_ACCESS_CODE = "pk.eyJ1IjoibW96aWxsYS13ZWJwcm9kIiwiYSI6ImNqMml4MXNhaTAwMjgzM28yYXRqbjlmZmkifQ.OfL6NY7noQth9mbbN_tDbA";
+    public static final String MAPBOX_LAYER_TYPE = "mapbox.streets";
+    public static final String[] MAPBOX_TILE_URLS = new String[]{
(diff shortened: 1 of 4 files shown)


libraries/stumbler/src/main/java/org/mozilla/mozstumbler/service/AppGlobals.java
```diff
@@ -10,6 +10,15 @@
 
 public class AppGlobals {
 
+    public static final String MAPBOX_ACCESS_CODE = "pk.eyJ1IjoibW96aWxsYS13ZWJwcm9kIiwiYSI6ImNqMml4MXNhaTAwMjgzM28yYXRqbjlmZmkifQ.OfL6NY7noQth9mbbN_tDbA";
```
crankycoder, line 13: Is there any licensing we have to worry about for this access code?

garvankeeley: I defer to @hannosch

hannosch: There's nothing to worry about. I've created that access code specifically for all MozStumbler usage, independent of the app store / way of downloading / building it.

hannosch (approved):

PR #1848, opened 2019-02-07
dropped the leaderboard and FxA which is only used for the leaderboard

crankycoder:
WIP: 

This removes the leaderboard and the Firefox Accounts code that supported the leaderboard. 

For UX - that means all leaderboard buttons and activities have gone away.

User nickname, Firefox accounts login screens and libfxa dependencies have been dropped.


crankycoder:
Travis's build environment for Android JDK work has changed. blargh.

JonnyTech:
@crankycoder: Is there a problem with the leaderboard?

PR #1862, opened 2019-07-27
Add support for notification channels for Android Oreo and higher

theopensourceguy:
Added basic support for notification channels, so the notification for the foreground service actually shows up on devices running Oreo or newer. Channels are only utilized on those devices using runtime version code checks.

The string resources still need translations and probably better wording. I'm not good at that.

+ Added method for registering notification channels to NotificationUtil
+ Registering channels in ClientStumblerService


theopensourceguy:
Not entirely sure why the CI build is failing, but it looks like a problem with the JDK version. If it's a f**k up on my end, let me know. But I might just be talking to myself here :/

cpeterso:
Thanks for the fix, @theopensourceguy! [Mozilla has retired MozStumbler](https://discourse.mozilla.org/t/retiring-mozilla-stumbler/75206), but I'm merging your PR in case anyone wants to fork MozStumbler.

